PR #1, opened 2017-01-15
Call athenz instead of athens

okuryu:
Athenz UI seems not to be yet, but it's should be called athenz on docs probably.


yahoocla:
CLA is valid!

PR #8, opened 2017-01-20
use idiomatic full import paths, goimports ordering

dmitris:
I would suggest to use the absolute import paths - I think it would be more Go-idiomatic and would allow to do `go get github.com/yahoo/athenz/...` at once without having to resort to running `make`.  It would also allow to browse the API docs in godoc.org and avoid the current error on https://godoc.org/github.com/yahoo/athenz/utils/zms_cli:
```
The go get command cannot install this package because of the following issues:

Unrecognized import path "../../clients/go/zms" (zms-cli.go:22:2)
Unrecognized import path "../../libs/go/zmscli" (zms-cli.go:23:2)
```

I also propose to switch to the goimports [1] convention of ordering the imports (in blocks, first stdlib, then third-party ones) to avoid spurious diffs when someone who uses goimports (which is quite popular) would save the source code.

[1] https://godoc.org/golang.org/x/tools/cmd/goimports

### DIFF
--- a/clients/go/zms/Makefile
+++ b/clients/go/zms/Makefile
@@ -12,7 +12,6 @@ RDL_LIB=github.com/ardielle/ardielle-go/rdl
 
 # check to see if go utility is installed
 GO := $(shell command -v go 2> /dev/null)
-export GOPATH=$(PWD)
 
 ifdef GO
 
--- a/clients/go/zms/auth.go
+++ b/clients/go/zms/auth.go
@@ -5,9 +5,10 @@ package zms
 
 import (
 	"fmt"
-	"github.com/ardielle/ardielle-go/rdl"
 	"log"
 	"strings"
+
+	"github.com/ardielle/ardielle-go/rdl"
 )
 
 //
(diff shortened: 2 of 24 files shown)


dmitris:
hmm... having the code in $GOPATH/src/github.com/yahoo/athenz, I can successfully run `mvn install -DskipTests=true -Dmaven.javadoc.skip=true -B -V`

havetisyan:
Hi Dimitry. This is on my list of things to be changed. The current setup allows a change in of the shared code and then re-using that same code right away to build the tools. So that's why we're using relative paths for now as oppose to copying source code around as a short term solution. This is also why we don't reference go get support now and instead the zms-cli is available through the bin package. I'm planning an end-to-end example using go with Athenz - should be published in about 2 weeks and as part of that change the plan is to remove the the relative paths and use more standard absolute paths with other checks that we currently do internally. Thanks.

havetisyan:
PR #13 addresses the issues raised in this PR. Thanks.

PR #41, opened 2017-04-04
updating data model diagram

mujibur:


mujibur:
Thanks for catching this, Lee. Agree that it changes the meaning of the diagram. I will get that replaced and update this PR.

boynton:
👍 Thanks!

PR #43, opened 2017-04-07
Patches for generalization

tatyano:


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/UserAuthority.java
```diff
@@ -36,6 +36,7 @@
     private static final Logger LOG = LoggerFactory.getLogger(UserAuthority.class);
     static final String ATHENZ_PROP_PAM_SERVICE_NAME = "athenz.auth.user.pam_service_name";
 
+    final String userDomain = System.getProperty("athenz.user_domain", "user");
```
havetisyan: Not sure what is the reason for this? Is there is a reason why you want to configure the domain name? Most likely you need to define your own authority as oppose to changing the default user authority.

servers/zms/schema/zms_server.sql
```diff
@@ -37,7 +37,8 @@ CREATE TABLE IF NOT EXISTS `zms_server`.`domain` (
   INDEX `idx_modified` (`modified` ASC),
   INDEX `idx_account` (`account` ASC),
   INDEX `idx_ypmid` (`ypm_id` ASC))
-ENGINE = InnoDB;
+ENGINE=InnoDB DEFAULT CHARSET=utf8 ROW_FORMAT=DYNAMIC;
+-- Prevent ERROR 1709 (HY000)
```
havetisyan: Can you please provide additional details as far as what is causing this error. 

btw, this is auto generated file so it cannot be modified. if necessary changes should be done in the workbench file and generate a new version of the sql script.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -740,10 +740,6 @@ void initObjectStore() {
             adminUsers.add(adminUser);
         }
         
-        if (!ZMSConsts.USER_DOMAIN.equals(userDomain)) {
```
havetisyan: I think the bug here is not to remove this block. Instead we want to create both domains - userDomain and USER_DOMAIN.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
```diff
@@ -122,7 +122,8 @@ public Domain postSubDomain(@PathParam("parent") String parent, @HeaderParam("Y-
     public Domain postUserDomain(@PathParam("name") String name, @HeaderParam("Y-Audit-Ref") String auditRef, UserDomain detail) {
         try {
             ResourceContext context = this.delegate.newResourceContext(this.request, this.response);
-            context.authorize("create", "user." + name + ":domain", null);
+            String userDomainPrefix = System.getProperty(ZMSConsts.ZMS_PROP_USER_DOMAIN, ZMSConsts.USER_DOMAIN) + ".";
```
havetisyan: this is auto generated file so no changes are allowed here

tatyano:
Henry,

Thank you for the review.
I'd like to talk about some changes but for now I'll just separate them and send another PR for ui first.

Thank you,
Chris

PR #92, opened 2017-05-15
Modularized login logic

tatyano:


tatyano:
Modularized login logic.
Eventually, TODO comments will be replaced with auth_core.

PR #97, opened 2017-05-16
Remove complete flag component from role token

havetisyan:
We're not using the complete role set component yet, so for now we're going to remove it from the token list and we'll consider re-adding it later when it becomes necessary.


gjoranv (approved):
Clean upstream sync, all files identical to v1.12.41. Copilot comments are about pre-existing upstream patterns, not our code.

PR #108, opened 2017-05-25
Use latest AWS java sdk and simply get object class

havetisyan:
Update to the latest AWS java sdk 1.11.135

When getting an object from S3, use the guava ByteStreams class to get the data from S3ObjectInputStream instance. Also use try-with-resources so the S3ObjectInputStream is automatically closed when the data is retrieved.


bluelamar:
+1

PR #110, opened 2017-05-25
GetUserList and DeleteUser api

havetisyan:
2 new api endpoints:

GET /user - returns all user principals defined in the system

DELETE /user/{name} - delete the user.{name} domain and all of it's subdomains. Remove user.{name} and user.{name}.* principals from all roles in the system.

A typical use case would be when a user leaves the company, we want to remove any personal domains the user has created and remove users from all roles that he/she was part of.


bluelamar:
+1

PR #111, opened 2017-06-01
Support disable flag on user domains

havetisyan:
Providing capability to system administrators to disable users (e.g. when they leave the company). The functionality (will be exposed in zms-cli) will allow system administrator to disable a user thus disabling all user personal domains. Once disabled, the services in those domains will no longer be authenticated and ZTS will not process these domains.  This would allow the system administrator to disable the domain for some short period (e.g. 1-2 weeks) to make sure that some services are not incorrectly using user domains for their functionality instead of proper domains. Then the user can be deleted from the system that would automatically delete all the user domains (including subdomains) and remove the user from any role in the system.


bluelamar:
+1

PR #114, opened 2017-06-02
Add support for user (disable/enable/list) support in zms-cli

havetisyan:
zms-cli has been updated to support 4 new user commands:

list-user -> list all registered user principals in Athens system
delete-user -> delete all personal domains and subdomains for this user (e.g. user.joe, user.joe.qa, etc) and remove user from all roles in the system
disable-user -> mark all the personal domains and subdomains for this user (e.g. user.joe, user.joe.qa, etc) as disabled
enable-user -> mark all the personal domains and subdomains for this user (e.g. user.joe, user.joe.qa, etc) as enabled


bluelamar:
+1

PR #116, opened 2017-06-05
Additional constructors for zms/zts clients

havetisyan:
additional constructors for ZMSClient and ZTSClient objects in order to avoid asking the developers to pass nulls when using athenz configuration file to extract the server urls


upendram:
👍

PR #117, opened 2017-06-07
make signer/verifier + ybase64 support public

havetisyan:
allows other utilities to take advantage of these interfaces instead of implementing their own. The version go version of zpu will use these methods.


psasidhar:
👍

Issue #126, opened 2017-06-12
zms client handling of add credentials from authority with cookie headers


havetisyan:
If the authority requires credentials to be passed with cookie headers, then the header is defined as "Cookie.<cookie-name>". However, zms java client does not have support for this format. It needs to check if the header starts with "Cookie." and handle accordingly.

PR #130, opened 2017-06-12
[126] zms client handling of add credentials from authority with cookie headers

havetisyan:
Addresses #126 

If the authority header is Cookie.NT, then the header is called Cookie and the cookie is sent as NT=<credentials>


hawk259:
👍

PR #140, opened 2017-06-15
Bug fix for loading the configuration file.

tatyano:
https://github.com/yahoo/athenz/compare/master...tatyano:auth_core_nodejs_bugfix?expand=1#diff-9e39287a2ed9f4daeea12dd2c36f7e05L37
The line above contains a critical bug only when using inside other libraries.


havetisyan:
👍

Issue #170, opened 2017-07-14
Adding implementation of AWS Credentials Provider 


gurleen-gks:
The AWSTemporaryCredentials is not an implementation of AWSSessionCredentials. So client libraries need to wrap it in their own providers. 
Adding implementation of AWSSessionCredentialsProvider as part of ZTS library

PR #171, opened 2017-07-14
#170 adding implementation of aws credentials provider

gurleen-gks:


yahoocla:

clients/java/zts/pom.xml
```diff
@@ -28,6 +28,21 @@
 
   <dependencies>
     <dependency>
+      <groupId>com.amazonaws</groupId>
```
havetisyan: this adds the full aws jdk as a dependency where we don't need most of it. check out how zts server handles this with the aws-java-sdk-bom and we should use the same version as zts server. 1.0.67 is quite old.

clients/java/zts/pom.xml
```diff
@@ -28,6 +28,21 @@
...
+      <artifactId>aws-java-sdk</artifactId>
+      <version>1.10.67</version>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
```
havetisyan: this is  not needed because the parent pom already has it.

clients/java/zts/pom.xml
```diff
@@ -28,6 +28,21 @@
...
+      <artifactId>slf4j-api</artifactId>
+      <version>1.7.5</version>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
```
havetisyan: this is  not needed because the parent pom already has it.

clients/java/zts/src/main/java/com/yahoo/athenz/zts/AWSCredentialsProviderImpl.java
```diff
@@ -0,0 +1,51 @@
...
+    @Override
+    public AWSCredentials getCredentials() {
+        this.refresh();
+        return this.credentials;
+    }
+
+    @Override
+    public void refresh() {
+        try {
+            long start = System.currentTimeMillis();
```
havetisyan: time stamp support was only for debugging and must not be in the code. let's remove any references.

clients/java/zts/src/main/java/com/yahoo/athenz/zts/AWSCredentialsProviderImpl.java
```diff
@@ -0,0 +1,51 @@
...
+            AWSTemporaryCredentials creds = ztsClt.getAWSTemporaryCredentials(athensSvcDomain, athensDomRole);
+            LOG.debug("AWSCredentialsProviderImpl:refresh: Credentials with id: \"" + creds.accessKeyId + "\" were fetched");
+            long end = System.currentTimeMillis();
+            long credentialFetchingTime = (end - start);
+            LOG.debug("AWSCredentialsProviderImpl:refresh: The Zts client took " + credentialFetchingTime + " milliseconds to fetch the credentials");
+            this.credentials = new BasicSessionCredentials(
+                    creds.getAccessKeyId(),
+                    creds.getSecretAccessKey(),
+                    creds.getSessionToken());
+        } catch (Exception exp) {
```
havetisyan: We should catch ZTSClientException first and do logging based on that exception. ZTSClientException returns the http status code so we should log that since that's useful for debugging rather than relying on the message only.

clients/java/zts/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
```diff
@@ -1278,6 +1278,18 @@ public Identity postInstanceRefreshRequest(String domain, String service, Instan
...
+
+    /**
+     * For a given domain and role return AWS credential provider
+     *
+     * @param domainName name of the domain
+     * @param roleName   is the name of the role
+     * @return AWSCredentialsProviderImpl AWS credential provider
+     */
+    public AWSCredentialsProviderImpl getAWSCredentialProvider(String domainName, String roleName) {
+        return new AWSCredentialsProviderImpl(this, domainName, roleName);
```
havetisyan: We need to improve the documentation for the API to clarify to the user that the ZTS client must not be closed while the credentials provider is being used or an alternative implementation must be considered where we make a copy of the zts client object. Otherwise when the caller closes the zts client object, our provider will not longer function properly.

PR #181, opened 2017-07-18
Use Throwable instead of Exception for updater thread run

havetisyan:
This would allow us to log any major issues like out of heap, etc without having the data updater thread exit with no error messages.  Would be useful for debugging if there are any issues.


hawk259:
👍

Issue #186, opened 2017-07-26
Add RoleToken support in Go libs


psasidhar:
Add a ztsroletoken library

PR #187, opened 2017-07-26
#186 Adding ztsroletoken library

psasidhar:


libs/go/ztsroletoken/Makefile
```diff
@@ -0,0 +1,84 @@
+#
+# Makefile to build ZTS Role Token client utility
+# Prerequisite: Go development environment
+#
+# Copyright 2016 Yahoo Inc.
```
havetisyan: should be 2017 Yahoo Holdings, Inc

please make the change in all other files as well.

libs/go/ztsroletoken/role-token.go
```diff
@@ -0,0 +1,177 @@
+package roletoken
+
+import (
+	"sync"
```
havetisyan: aren't we running the import check on these files? they seem to be out of order

libs/go/ztsroletoken/role-token.go
```diff
@@ -0,0 +1,177 @@
+package roletoken
```
havetisyan: add license lines:

// Copyright 2017 Yahoo Holdings, Inc.
// Licensed under the terms of the Apache version 2.0 license. See LICENSE file for terms.

libs/go/ztsroletoken/pom.xml
```diff
@@ -0,0 +1,72 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright 2016 Yahoo Inc.
```
havetisyan: fix copyright line

libs/go/ztsroletoken/role-token.go
```diff
@@ -0,0 +1,177 @@
...
+
+	"crypto/x509"
+	"fmt"
+
+	"github.com/yahoo/athenz/clients/go/zts"
+	"github.com/yahoo/athenz/libs/go/zmssvctoken"
+)
+
+const (
+	defaultPrincipalAuthHeader = "Athenz-Principal-Auth"
```
havetisyan: Having this as a const, will it prohibit the use in a different environment with a different header name?

libs/go/ztsroletoken/role-token_test.go
```diff
@@ -0,0 +1,202 @@
+package roletoken
```
havetisyan: please add license info

psasidhar: done

libs/go/ztsroletoken/role-token.go
psasidhar: User can pass the custom header through RoleTokenOptions.AuthHeader member

libs/go/ztsroletoken/role-token.go
psasidhar: fixed the formatting

PR #212, opened 2017-08-16
Enable client x.509 cert generation for lambda

boynton:
The general idea is to specify no instance identity document for
lambda, which results in a client-only x.509 cert, and no SSH certs at
all. This is a short term fix on the legacy API only. The new
InstanceRegisterInformation API will need an attestationData definition
that accommodates this feature, when it gets implemented.


havetisyan:
👍

PR #223, opened 2017-08-25
Okta

charlesk40:
Support for adding application id to domain.
Check for an application id when requesting zts role tokens

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
@@ -199,7 +199,7 @@ public class DBService {
     }
     
     Domain makeDomain(ResourceContext ctx, String domainName, String description, String org,
-            Boolean auditEnabled, List<String> adminUsers, String account, int productId,
+            Boolean auditEnabled, List<String> adminUsers, String account, int productId, String applicationId,
             List<String> solutionTemplates, String auditRef) {
         
         final String caller = "makedomain";
@@ -212,7 +212,8 @@ public class DBService {
                 .setId(UUID.fromCurrentTime())
                 .setAccount(account)
                 .setYpmId(productId)
-                .setModified(Timestamp.fromCurrentTime());
+                .setModified(Timestamp.fromCurrentTime())
+                .setApplicationId(applicationId);
         
         // get our connection object
         
@@ -1996,6 +1997,13 @@ public class DBService {
                 } else {
                     updatedDomain.setYpmId(meta.getYpmId());
                     updatedDomain.setAccount(meta.getAccount());
+                }
+                
+                // if meta application ID is null, update to existing application ID
+                if (meta.getApplicationId() == null) {
+                    updatedDomain.setApplicationId(domain.getApplicationId());
+                } else {
+                    updatedDomain.setApplicationId(meta.getApplicationId());
                 }
                 
                 con.updateDomain(updatedDomain);
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -741,17 +741,18 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
             }
             adminUsers.add(adminUser);
         }
-        
+
         if (!ZMSConsts.USER_DOMAIN.equals(userDomain)) {
             createTopLevelDomain(null, ZMSConsts.USER_DOMAIN, "The reserved domain for user authentication",
-                    null, null, adminUsers, null, 0, null, null);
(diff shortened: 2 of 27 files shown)


hawk259:
👍

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -200,7 +200,7 @@ void auditLogRequest(ResourceContext ctx, String domainName, String auditRef,
     
     Domain makeDomain(ResourceContext ctx, String domainName, String description, String org,
             Boolean auditEnabled, List<String> adminUsers, String account, int productId,
-            List<String> solutionTemplates, String auditRef) {
+            List<String> solutionTemplates, String auditRef, String applicationId) {
```
havetisyan: Can we please move applicationId argument to be after productId. It'd like to be consistent and have auditRef/caller as last arguments in our APIs.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5889,15 +5889,15 @@ void validate(Object val, String type, String caller) {
     
     Domain createTopLevelDomain(ResourceContext ctx, String domainName, String description,
             String org, Boolean auditEnabled, List<String> adminUsers, String account,
-            int productId, List<String> solutionTemplates, String auditRef) {
+            int productId, List<String> solutionTemplates, String auditRef, String applicationId) {
```
havetisyan: Can we please move applicationId argument to be after productId. It'd like to be consistent and have auditRef/caller as last arguments in our APIs.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5889,15 +5889,15 @@ void validate(Object val, String type, String caller) {
...
         List<String> users = validatedAdminUsers(adminUsers);
         return dbService.makeDomain(ctx, domainName, description, org, auditEnabled, 
-                users, account, productId, solutionTemplates, auditRef);
+                users, account, productId, solutionTemplates, auditRef, applicationId);
     }
     
     Domain createSubDomain(ResourceContext ctx, String parentName, String name, String description,
             String org, Boolean auditEnabled, List<String> adminUsers, String account,
-            int productId, List<String> solutionTemplates, String auditRef, String caller) {
+            int productId, List<String> solutionTemplates, String auditRef, String caller, String applicationId) {
```
havetisyan: Can we please move applicationId argument to be after productId. It'd like to be consistent and have auditRef/caller as last arguments in our APIs.

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
```diff
@@ -723,16 +723,32 @@ void processTrustMembership(DataCache data, String identity, String rolePrefix,
...
+    public void getAccessibleRoles(DataCache data, String domainName, String identity,
+            String roleName, Set<String> accessibleRoles, boolean keepFullName, String applicationId) {
 
         /* if the domain hasn't been processed then we don't have anything to do */
         
         if (data == null) {
             return;
         }
+        
+        /* check if application id matches the principal's application id */
```
havetisyan: Any reason why we can't do the check one layer above in the ZMSImpl?

charlesk40: We can do that in ZMSImpl. I think the reason for this to be in getAccessibleRoles is that I noticed that this api is called through out other apis in ZMSImpl which might need this check as well. If not, I can directly add it to getRoleTokens api.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
charlesk40: done

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
charlesk40: done

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
charlesk40: done

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
havetisyan: Actually in all those other cases we don't want to add the restriction. It must only be checked when the actual getRoleToken api is called since that's the only time that we sign and return a role token.

charlesk40: done.

PR #224, opened 2017-08-31
command line zms utility to add/delete public key

charlesk40:


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImplMain.java
```diff
@@ -0,0 +1,99 @@
...
+
+import java.io.BufferedReader;
+import java.io.FileReader;
+import java.io.IOException;
+import java.util.Arrays;
+
+import com.yahoo.athenz.auth.util.Crypto;
+
+public class ZMSImplMain {
+    private static final String AUDITREF = "zms_standalone";
```
havetisyan: We should set the audit reference to something like - "Server bootstrap: instance private key registration"

charlesk40:
@havetisyan please review

servers/zms/scripts/zms_standalone.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
+#############
+# Usage: See display_usage
+# Requires zms.war and various system property files and must run as root or sudo user
+# Assumes that this run on AWS.
+###############
+
+if [[ $EUID -ne 0 ]]; then
```
havetisyan: @charlesk40 we want this check within the zmsimplmain otherwise I can skip the script and call the zms directly.

charlesk40: done

PR #232, opened 2017-09-07
Add nodejs client

tatyano:


Syzygy24:
@tatyano Have you published these nodejs clients to npm already and would you be able to point me to the module? If not, are you thinking about doing so?

havetisyan:
@Syzygy24 we haven't yet published these clients to npm but they should be available within a week or so.

PR #241, opened 2017-09-14
AWS implementation of PrivateKeyStore using S3 bucket and KMS

charlesk40:


libs/java/auth_core/pom.xml
```diff
@@ -37,6 +37,11 @@
       <artifactId>libpam4j</artifactId>
       <version>1.8</version>
     </dependency>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-s3</artifactId>
+      <version>1.11.194</version>
```
havetisyan: We should use ${aws.version} property from parent pom to have consistent aws library usage in Athenz.

charlesk40: fixed

PR #250, opened 2017-09-22
Improve npm package names

okuryu:
I feel Athenz npm package's names aren't better for practices of Node.js.

1. Avoid redundant naming: "nodejs" in package name is redundant. npm packages are always provided for Node.js runtime.
2. Use hyphen instead of underscore: npm package name usually uses hyphen. e.g. `eslint-`, `babel-`.
3. Push all packages into one namespace: `auth-core` doesn't associate Athenz with the package name. Use scoped package name instead.

This changes is my proposal these issues. Please let me know if you have more idea on this.

### DIFF
--- a/assembly/ui/ui.xml
+++ b/assembly/ui/ui.xml
@@ -46,12 +46,12 @@
       <directory>${basedir}/../../ui/node_modules</directory>
       <outputDirectory>node_modules</outputDirectory>
       <excludes>
-        <exclude>auth_core/</exclude>
+        <exclude>@athenz/auth-core/</exclude>
       </excludes>
     </fileSet>
     <fileSet>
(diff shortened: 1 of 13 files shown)


havetisyan:
@okuryu Thanks for the PR but have you had a chance to look at the Travis CI build failure and provide an update to fix the build issue. Thanks.

okuryu:
The build failure due to the `@athenz/auth-core` package isn't available on the npm registry yet. I just reverted the path to `@athenz/auth-core` to a relative path to resolve dependencies of ui in 586b77e.

The building passed now, but I think we should revert 586b77e again after the `@athenz/auth-core` package is published on the npm registry.

PR #261, opened 2017-09-28
UI depends on @athenz/auth-core instead

okuryu:
`@athenz/auth-core` package is available on npm registry now. The UI should
depends on `@athenz/auth-core` package instead of the relative path. This is
one more change as a part of #250.

See also: https://github.com/yahoo/athenz/pull/250#issuecomment-332199611


havetisyan:
👍

PR #265, opened 2017-10-01
optional cloud info and allow dash in role names

charlesk40:


havetisyan:
@charlesk40 not sure about this dash support. that should not be necessary. I need to review the full method to see how it is being used.

charlesk40:
@havetisyan  There is currently check for role name that would look for period.
Oct  1 21:54:14 ip-10-0-1-78 start.sh[9842]: [ZTS-SERVER] 21:54:14.081 [main] ERROR c.yahoo.athenz.zts.store.CloudStore - CloudStore: malformed service name: athenz-zts-service

havetisyan:
https://github.com/yahoo/athenz/pull/268

this should address this issue by removing all the extra bits that we no longer need.

PR #288, opened 2017-10-16
ZPE support for certificates

charlesk40:
AllowAccess check for X509Certificate
Upgrade to Mockito current version 2

### DIFF
--- a/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
+++ b/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
@@ -15,19 +15,26 @@
  */
 package com.yahoo.athenz.zpe;
 
+import java.security.PublicKey;
+import java.security.cert.X509Certificate;
+import java.util.ArrayList;
+import java.util.HashSet;
+import java.util.List;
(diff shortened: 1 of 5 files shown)


hawk259:
👍

clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
```diff
@@ -165,6 +210,68 @@ public static PublicKey getZmsPublicKey(String keyId) {
...
+        }
+        int idx = subject.indexOf(":");
+        if (idx == -1) {
+            if (LOG.isDebugEnabled()) {
+                LOG.debug("AUTHZPECLT:allowAccess: missing domain");
+            }
+            return AccessCheckStatus.DENY_CERT_MISSING_DOMAIN;
+        }
+        String domainName = subject.substring(0, idx);
+        int roleIndex = subject.indexOf(".");
```
havetisyan: This check is not correct since the domain name can contain a "."
e.g. sys.auth:role.readers
we should search for ":role." for the domain name/role name index. Anything before that is the domain name and everything after is the role name

havetisyan:
👍

PR #296, opened 2017-10-23
#295 Extend openhouse template to create default service 'openhouse'

pratikgote:


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2349,6 +2391,54 @@ Policy updateTemplatePolicy(Policy policy, String domainName, String policyName,
         return templatePolicy;
     }
     
+    ServiceIdentity updateTemplateServiceIdentity(ServiceIdentity serviceIdentity, String domainName, String serviceIdentityName, List<TemplateParam> params) {
+        ServiceIdentity templateServcieIdentity = new ServiceIdentity().setName(ZMSUtils.serviceResourceName(domainName, serviceIdentityName));
+        templateServcieIdentity.setDescription("");
```
havetisyan: no need to default to empty string.  we should be able just to call:

templateServiceIdentity.setDescription(serviceIdentity.getDescription());

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2349,6 +2391,54 @@ Policy updateTemplatePolicy(Policy policy, String domainName, String policyName,
         return templatePolicy;
     }
     
+    ServiceIdentity updateTemplateServiceIdentity(ServiceIdentity serviceIdentity, String domainName, String serviceIdentityName, List<TemplateParam> params) {
+        ServiceIdentity templateServcieIdentity = new ServiceIdentity().setName(ZMSUtils.serviceResourceName(domainName, serviceIdentityName));
```
havetisyan: templateServcieIdentity service is misspelled

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2349,6 +2391,54 @@ Policy updateTemplatePolicy(Policy policy, String domainName, String policyName,
...
+                PublicKeyEntry newPublicKeyEntry = new PublicKeyEntry();
+                newPublicKeyEntry.setId(publicKeyEntry.getId());
+                newPublicKeyEntry.setKey(publicKeyEntry.getKey());
+                newPublicKeyEntries.add(newPublicKeyEntry);
+            }
+        }
+        templateServcieIdentity.setPublicKeys(newPublicKeyEntries);
+        
+        List<String> hosts = serviceIdentity.getHosts();
+        List<String> newHosts = new ArrayList<String>();
```
havetisyan: Same as above. No need to set hosts if the current host is null. 

Also don't need to iterate and add one string at a time. We can just call
templateServiceIdentity.setHosts(new ArrayList<String>(hosts));
this helps with simpler tests and better code coverage.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2349,6 +2391,54 @@ Policy updateTemplatePolicy(Policy policy, String domainName, String policyName,
...
+        
+        if (hosts != null && !hosts.isEmpty()) {
+            for (String host: hosts) {
+              newHosts.add(host);
+            }
+        }
+        
+        templateServcieIdentity.setHosts(newHosts);
+        
+        if (serviceIdentity.getExecutable() != null) {
```
havetisyan: For the next 4 if blocks, we don't need the if blocks. We should just go head and set the values even if they're null directly:
templateServcieIdentity.setExecutable(serviceIdentity.getExecutable());
...

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
havetisyan: This is actually not correct. The name in the object must match to how we're returning json objects to clients - which is always in <domain>.<service> format.

servers/zms/src/test/resources/solution_templates.json
```diff
@@ -236,6 +236,97 @@
...
+                        "resource": "_domain_:vip*",
+                        "role": "_domain_:role.vip_admin",
+                        "action": "*"
+                    }
+                ]
+            }
+        ],
+        "services": [
+            {
+                "name": "testService",
```
havetisyan: The correct value here must be: "name": "_domain_:testservice". Please update all other test cases and handling of the service name accordingly.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2349,6 +2392,42 @@ Policy updateTemplatePolicy(Policy policy, String domainName, String policyName,
         return templatePolicy;
     }
     
+    ServiceIdentity updateTemplateServiceIdentity(ServiceIdentity serviceIdentity, String domainName, String serviceIdentityName, List<TemplateParam> params) {
+        
+        String templateServiceName = serviceIdentityName;
+        if (params != null) {
+            for (TemplateParam param : params) {
+                final String paramKey = "_" + param.getName() + "_";
+                System.out.println("param key &&&&&&&&&& " + paramKey + " &&&&&&");
```
havetisyan: this line should be removed.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2349,6 +2392,42 @@ Policy updateTemplatePolicy(Policy policy, String domainName, String policyName,
...
+                templateServiceName = templateServiceName.replace(paramKey, param.getValue());
+            }
+        }
+        
+        ServiceIdentity templateServiceIdentity = new ServiceIdentity().setName(ZMSUtils.serviceResourceName(domainName, serviceIdentityName));
```
havetisyan: Here we should be using updated service name: templateServiceName and not serviceIdentityName

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2349,6 +2392,42 @@ Policy updateTemplatePolicy(Policy policy, String domainName, String policyName,
...
+            for (PublicKeyEntry publicKeyEntry : publicKeyEntries) {
+                PublicKeyEntry newPublicKeyEntry = new PublicKeyEntry();
+                newPublicKeyEntry.setId(publicKeyEntry.getId());
+                newPublicKeyEntry.setKey(publicKeyEntry.getKey());
+                newPublicKeyEntries.add(newPublicKeyEntry);
+            }
+        }
+        templateServiceIdentity.setPublicKeys(newPublicKeyEntries);
+        
+        templateServiceIdentity.setHosts(serviceIdentity.getHosts());
```
havetisyan: we shouldn't set a reference to the host list since it could be modified. Instead we should do:
if (hosts != null) {
templateServiceIdentity.setHosts(new ArrayList(hosts));
}

servers/zms/src/main/java/com/yahoo/athenz/zms/utils/ZMSUtils.java
```diff
@@ -114,6 +114,14 @@ public static String removeDomainPrefix(String objectName, String domainName, St
         return objectName;
     }
     
+    public static String removeDomainPrefixForService(String serviceName, String domainName) {
+        String valPrefix = domainName + ".";
```
havetisyan: please make this final

PR #307, opened 2017-10-31
Client tls

charlesk40:
TLS client support in zms and zts.

ClientSSLContextBuilder is provided to create SSLContext to zms and zts clients.

Example:
        ClientSSLContextBuilder builder = new SSLUtils.ClientSSLContextBuilder(sslProtocol);
        SSLContext sslContext = builder.build();

If there are multiple certs in keystore, client can send particular cert by setting the certAlias method.

Example:
        ClientSSLContextBuilder builder = new SSLUtils.ClientSSLContextBuilder(sslProtocol);
        builder.certAlias("myalias");
        SSLContext sslContext = builder.build();


bluelamar:
+1

libs/java/client_common/src/test/resources/certs/client/.srl
```diff
@@ -0,0 +1 @@
+D9AF854508733FBB
```
havetisyan: What is this file for?

libs/java/client_common/src/test/resources/certs/server/.srl
```diff
@@ -0,0 +1 @@
+CE33F458FC553293
```
havetisyan: Same here - what is this for?

libs/java/client_common/src/main/java/com/yahoo/athenz/common/utils/SSLUtils.java
```diff
@@ -0,0 +1,280 @@
...
+            KeyManager[] keyManagers = null;
+            TrustManager[] trustManagers = null;
+            
+            if (keyStorePath == null && trustStorePath == null) {
+                return null;
+            }
+            
+            try {
+                if (keyStorePath != null) {
+                    LOGGER.info("createSSLContextObject: using SSL KeyStore path: " + keyStorePath);
```
havetisyan: We should use {} instead of + so that we don't create the string if the info level is not enabled.

libs/java/client_common/src/main/java/com/yahoo/athenz/common/utils/SSLUtils.java
```diff
@@ -0,0 +1,280 @@
...
+                    kmf = KeyManagerFactory.getInstance(KeyManagerFactory.getDefaultAlgorithm());
+                    if (keyManagerPassword == null) {
+                        throw new IllegalArgumentException("Missing key manager password for the key store: " + keyStorePath);
+                    }
+                    keyManagerPassword = getPassword(keyManagerPassword, privateKeyStore, keyManagerPasswordAppName);
+                    kmf.init(keyStore, keyStorePassword);
+                    keyManagers = getAliasedKeyManagers(kmf.getKeyManagers(), certAlias);
+                }
+                if (trustStorePath != null) {
+                    LOGGER.info("createSSLContextObject: using SSL TrustStore path: " + trustStorePath);
```
havetisyan: Same here:  We should use {} instead of + so that we don't create the string if the info level is not enabled.

PR #310, opened 2017-11-04
support building trust tore from a ca cert file

smrutilal2:
1. Enabled Junit test cases in the maven build.
2. Fixed existing test failure.
3. Added support to build trust store from a ca cert file.


yahoocla:

libs/java/cert_refresher/src/main/java/com/oath/auth/TrustStore.java
```diff
@@ -0,0 +1,47 @@
...
+ * or implied. See the License for the specific language governing permissions and limitations under
+ * the License.
+ */
+
+import java.security.KeyStore;
+import javax.net.ssl.TrustManager;
+import javax.net.ssl.TrustManagerFactory;
+
+/**
+ * A trust store built from a jks file by default. Or from the keystore provided by the
```
SeanPMiller: Incomplete doc?

libs/java/cert_refresher/src/main/java/com/oath/auth/KeyStoreProvider.java
```diff
@@ -0,0 +1,74 @@
...
+ * limitations under the License.
+ */
+
+import java.io.FileInputStream;
+import java.io.InputStream;
+import java.security.KeyStore;
+import java.security.cert.CertificateFactory;
+import java.security.cert.X509Certificate;
+
+public interface KeyStoreProvider {
```
SeanPMiller: Please write docs for interface and `provide()` method.

libs/java/cert_refresher/src/test/java/com/oath/auth/SocketTest.java
```diff
@@ -106,7 +106,7 @@ private void runPingServer(int port, KeyRefresher keyRefresher) throws Exception
         }).start();
     }
 
-    @Test
+//    @Test
```
SeanPMiller: Should this test be reenabled?

libs/java/cert_refresher/src/test/resources/gdpr.aws.core.key.pem
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
SeanPMiller: Yikes! Where did this come from?

smrutilal2: This is a test key.

libs/java/cert_refresher/src/test/java/com/oath/auth/SocketTest.java
smrutilal2: FYI.  Tests were not running initially.  This particular test like many others was not complete and need little more time. But yes, we should fix and enable it.

havetisyan:
@smrutilal2 Can you please provide some documentation as Sean pointed out.

Also, I noticed that the PR is changing the KeyRefresh class definition. Was this not already used by other teams or other teams were only using it through Utils.generateKeyRefresher to access the key refresher. @mmannerm can you please confirm that you're not affected by this change. thanks.

libs/java/cert_refresher/src/main/java/com/oath/auth/Utils.java
```diff
@@ -86,9 +78,20 @@ public static KeyStore getKeyStore(final String jksFilePath) throws Exception {
      *  It requires that the proxies are created which are then stored in the KeyRefresher
      */
     public static KeyRefresher generateKeyRefresher(final String trustStorePath, final String athensPublicKey, final String athensPrivateKey) throws Exception {
+//        TrustStore trustStore = new TrustStore(trustStorePath, new JavaKeyStoreProvider(trustStorePath, "secret"));
```
havetisyan: please remove this commented line since it's identical to the next one.

PR #311, opened 2017-11-04
move validation code to server

havetisyan:
we should move the validation code to the server since the library doesn't know what is the current regex allowed for principal components and could get out of sync.


charlesk40:
👍

PR #326, opened 2017-11-13
Add go zpe-updater binary to athenz-utils assembly

tatyano:
For now, the document of zpu setup is guiding us to download the package with assembled tar file.
https://github.com/yahoo/athenz/blob/master/docs/setup_zpu.md#getting-software

I believe it's better to have both versions of java and go, and let users choose the solution.
(Go version has less dependency.)


havetisyan:
Thanks. Soon we'll be updating the documentation to use the go version as it doesn't make sense to maintain both versions and go version is much simpler to install and use as you also noted based on its dependency list.

PR #327, opened 2017-11-13
zms-cli correct handling of user domains when home domain is configur…

havetisyan:
…ed with a different value other than user


hawk259:
👍

PR #345, opened 2017-12-07
ServletRequestUtil should get first ip address from XFF header to determine initial client.

tatyano:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/util/ServletRequestUtil.java
```diff
@@ -34,7 +34,7 @@ public static String getRemoteAddress(final HttpServletRequest request) {
             String xff = request.getHeader(XFF_HEADER);
             if (xff != null) {
                 String[] addrs = xff.split(",");
-                addr = addrs[addrs.length - 1].trim();
+                addr = addrs[0].trim();
```
havetisyan: That is not the intended behavior of this function so using the last IP address is the correct behavior. It's not designed to handle the case where you have "client, proxy1, proxy2, proxy3" case but rather only the case where you might have a proxy server sitting on the same box as zms/zts servers and you don't want to log the loopback address. So there is an if block where it retrieves the xff address only if the address is the loopback. Otherwise, the xff cannot be trusted.

PR #346, opened 2017-12-11
zmssvctoken: fix golint issues and correct comments

dmitris:
* fix all golint issues in libs/go/zmssvctoken except the one related to the exported field (to avoid breaking API compatibility): `token.go:348:2: struct field ZTSBaseUrl should be ZTSBaseURL`
* make comment sentences end with a period
* Use the official name of the Go language (with the capital 'G') - as in https://golang.org/

@havetisyan


havetisyan:
Thanks @dmitris

PR #349, opened 2017-12-11
add package documentation in doc.go files

dmitris:
Objective is to ensure that all entries in https://godoc.org/github.com/yahoo/athenz have a short summary as package or program description.

@havetisyan


havetisyan:
👍

Issue #350, opened 2017-12-13
zts role token restriction on appid move to authz service


havetisyan:
Instead of having a restriction on principals to verify client ids for domains, we should be consistent and use the authorized service concept and restrict that if you come with an authorized service token, then you can request role tokens for that service domain only. The authority can be responsible for making sure the token has the authorized service is set if one is present (current Principal authority already does this).

PR #351, opened 2017-12-13
#350 zts role token restriction on appid move to authz service

havetisyan:


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -1164,31 +1164,27 @@ boolean isAuthorizedProxyUser(Set<String> proxyUsers, String principal) {
         return proxyUsers.contains(principal);
     }
     
-    private void checkRoleTokenApplicationIdRequest(ResourceContext ctx, DomainData domain, String caller) {
+    void checkRoleTokenAuthorizedServiceRequest(final Principal principal,
+            final String domainName, final String caller) {
         
-        final Principal principal = ((RsrcCtxWrapper) ctx).principal();
-        final String applicationId = principal.getApplicationId();
+        final String authorizedService = principal.getAuthorizedService();
```
bluelamar: +1

PR #353, opened 2017-12-14
fix package doc strings

dmitris:
PR fixes the package doc strings (need to add a blank line between the copyright block and the package comment)

@havetisyan


havetisyan:
👍

PR #354, opened 2017-12-14
fix selected golint issues

dmitris:
Fix golint issues such as listed on http://go-lint.appspot.com/github.com/yahoo/athenz/libs/go/zmscli.

Avoided changes in the code that is marked as rdl generated as well as modifying the exported names (such as `Id`) to preserve the API compatibility.

The PR builds on #353 and is cumulative with it (may want to review and merge that one first)

@havetisyan

### DIFF
--- a/clients/go/zms/auth.go
+++ b/clients/go/zms/auth.go
@@ -5,16 +5,15 @@ package zms
 
 import (
 	"fmt"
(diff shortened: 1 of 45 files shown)


havetisyan:
Thanks @dmitris. There were a couple of files what were modified for the import order but those are rdl generated so we'll lose those changes during next PR

PR #357, opened 2017-12-18
Rdl latest

havetisyan:


clients/go/zms/client.go
```diff
@@ -8,15 +8,14 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	rdl "github.com/ardielle/ardielle-go/rdl"
```
dmitris: why this change of imports ordering?  Could we keep the goimports-compatible order of imports (first stdlib, then empty line, then third-party packages) - otherwise if someone has goimports integrated with the code editor and opens/saves the file, causes spurious diffs.

Also in terms of consistency, I though the goimports-compatible imports ordering is already a de facto coding standard for Athenz - see for example https://github.com/yahoo/athenz/blob/master/clients/go/zms/auth.go#L6=L11 and https://github.com/yahoo/athenz/blob/master/clients/go/zms/client.go#L7-L19 that separate stdlib and third-party imports.

dmitris:
👍 
this is done with rdl 1.4.14, right?

Do you think we should mention rdl / ardielle-tools in https://github.com/yahoo/athenz/blob/master/docs/dev_environment.md ?

havetisyan:
@dmitris we left it out since you only need to make sure to install and configure go properly. In our scripts we automatically get the utility and use it so we were trying not complicate the steps but we can add a comment how we're using go and what utilities we use. I'll update the docs when I get a chance. Thanks.

PR #377, opened 2018-01-06
Update instance provider class structure

havetisyan:
move instance confirmation object to the instance-provider jar file and remove dependency on zts core to allow devs to write their own instance provider implementations.


charlesk40:
👍

PR #378, opened 2018-01-06
expiry/etag check

havetisyan:
GetEtagForExistingPolicy - must never return failure - for any error case it should just return "" so the client can fetch the latest policy

expiry timeout - increase by default to 2 days (zts issues expiry of 7 days) instead of just relying on startup delay timeout. this way in case there are failures with retrieving the policy fail, it would not end up failing to fetch new policy files due to expiry time.


charlesk40:
👍

PR #411, opened 2018-02-08
change zts client error to warn there is no conf file

havetisyan:
the primary use case now is to specify the zts url so there may never use a conf file. during the client init we should not flag this as an error instead it should be warn


hawk259:
👍

PR #413, opened 2018-02-12
Adding aws support

gurleen-gks:


hawk259:
👍

utils/zpe-updater/cmd/tools/main.go
```diff
@@ -43,6 +52,12 @@ func main() {
 	}
 
 	if logFile != "" {
+		err := ioutil.WriteFile(logFile, []byte("Starting zpe-updater"), 0755)
```
havetisyan: This will truncate the file every time is called so that's not want we want. You want to make sure you open the file in append mode and not truncate it

utils/zpe-updater/zpu_config.go
```diff
@@ -20,6 +20,7 @@ const (
 	DEFAULT_STARTUP_DELAY = 0
 	MAX_STARTUP_DELAY     = 1440
 	DEFAULT_EXPIRY_CHECK  = 2880
+	DEFAULT_ZTS_URL       = "https://zts.athens.yahoo.com:4443/zts/v1"
```
havetisyan: no internal addresses in opensource!

PR #416, opened 2018-02-21
Fix multiple bugs related to Node.js libraries

tatyano:
- Fixed log messages in auth_core of Node.js
- Fixed Node.js version of zpe-client to match begin-with resource
- Fix new Buffer to Buffer.from
- Fixed log messages in zpe-client of Node.js


havetisyan:
👍  Thanks Chris.

PR #417, opened 2018-02-24
fix support for aws temp creds for roles with paths

havetisyan:
when an aws role contains a path - e.g. sso/admin, this will be send to ZTS server as:

sso%252Fadmin which jax-rs will process first and provide the implementation with sso%2Fadmin and we need to decode it to get back sso/admin before trying to assume that role


hawk259:
👍

PR #418, opened 2018-02-26
Athenz servers should provide the detailed error messages

tatyano:
@havetisyan Hi Henry,

For now, Athenz servers only provide "Invalid credentials" message in the response.
https://github.com/yahoo/athenz/blob/master/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/rest/Http.java#L137

To let users easily determine the cause of the error, I believe it's better to provide detailed error messages in  the response.

Thank you,
Chris


havetisyan:
From security perspective, you don't want to provide additional details for login operations since it would expose internal information about your system - such as the user is valid but the password is correct, the server was expecting some type of header, etc.

So when it comes to authentication failures, your response must always be very generic type like invalid user or password and nothing more. So with http, 401 invalid credentials is the right response. Additional details are logged so the system administrator can help with the debugging process.

tatyano:
Thank you for your advice.
I'll withdraw this request.

PR #420, opened 2018-03-01
Adding X.509 certificate support to connect to zts server

gurleen-gks:


hawk259:
👍

utils/zpe-updater/cmd/tools/main.go
```diff
@@ -43,6 +51,14 @@ func main() {
 	}
 
 	if logFile != "" {
+
+		f, err := os.OpenFile(logFile, os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
+		if err != nil {
+			log.Fatalf("The log file:%v cannot be opened, Error:%v. \n "+
+				"If you donot have write access to the default log file /home/athenz/logs/zpu/zpu.log, use the -logFile flag to overwrite the default value. \n"+
```
havetisyan: donot is missing a space - do not.

I also find this error line confusing. not sure why it has to /home/athenz/logs/zpu/zpu.log where it's trying to open a file called logFile. Instead of using the hardcoded path, it should just replace that with logFile value and remove the reference to default value. something like:

"If you do not have write access to the log file %s, use the -logFile flag to overwrite the value. \n"+

gurleen-gks:
@hawk259 Change done !

PR #421, opened 2018-03-09
InstanceConfirmation request for refresh should only occur when attestation data is passed

tatyano:
The current implementation design provides the InstanceConfirmation request for the use of refresh.
However, this might bring an overloading concern since the provider web api will be called every time instances refresh their x509/ssh certificates or ntoken.
I considered adding another flag in the system property to disable this request, but by looking at the process for validation of attestation data, it seems InstanceConfirmation request is only necessary when the attestation is passed.

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -2030,32 +2030,35 @@ public class ZTSImpl implements KeyStore, ZTSHandler {
             throw requestError("dnsName attribute mismatch in CSR", caller, domain);
         }
         
-        // validate attestation data is included in the request
-        
-        InstanceProvider instanceProvider = instanceProviderManager.getProvider(provider);
-        if (instanceProvider == null) {
-            throw requestError("unable to get instance for provider: " + provider, caller, domain);
-        }
-        
         InstanceConfirmation instance = generateInstanceConfirmObject(ctx, provider,
                 domain, service, info.getAttestationData(), certReq);
         
-        // make sure to close our provider when its no longer needed
-
-        try {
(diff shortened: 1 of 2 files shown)


havetisyan:
That would introduce a security hole in the implementation. Take a look at the following use case:

a) the service deploys a new instance in AWS. it is started with its own Athenz service identity - generates a private key and gets the x.509 cert from Athenz.
b) the hacker gets access to the instance, copies the key and cert, and terminates the instance.
c) Most likely the auto scaling group will just bring up another instance for the terminated one.

However, at this point the hacker has access to the key/cert and can indefinitely refresh the certs without passing any attestation data.

**Every refresh request must be validated by the provider service.** 

In this AWS case, the refresh request must be accompanied with the temp credentials for the athenz service role so it can be validated by the provider. The provider can also validate that the instance is still running.

As a provider, you have the option not to check anything and just return 200 OK, but from ZTS perspective, it must contact and validate the refresh request.

tatyano:
> Every refresh request must be validated by the provider service.

I quite understand this idea, however if the hacker gets access to the instance referred to in b), I was afraid the attestation data would also be copied or received by making a request from the instance.
If the hacker has an attestation data to spoof as the instance to request ZTS, I wondered how could we prevent it by making attestation data necessary for the request.

havetisyan:
The attestation data cannot be copied. As a provider you must only accept attestation data that cannot be replayed after some short period of time. For example:

a) AWS - you must submit a temp credentials for the athenz role you assume. those credentials are valid for max 1 hour only. So if you copied the temp credentials, you can't refresh after 1 hour.

b) Openstack - the provider verifies using nova api that the instance (identified by its uuid) is still running thus you cannot have an instance that is no longer active asking for a refresh.

ZTS relies on the provider to validate the request before issuing certs. It is the provider's job to validate both initial registration and subsequent refresh operations.

PR #428, opened 2018-03-20
expose domain and service with SSLContext for zts java client

charlesk40:


clients/java/zts/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
```diff
@@ -410,16 +410,32 @@ public ZTSClient(String ztsUrl, SSLContext sslContext) {
...
+     * change these values by using the athenz.zts.client.read_timeout and
+     * athenz.zts.client.connect_timeout system properties. The values
+     * specified for timeouts must be in milliseconds.
+     * @param ztsUrl ZTS Server's URL
+     * @param proxyUrl Proxy Server's URL
+     * @param sslContext SSLContext that includes service's private key and x.509 certificate
+     * @param domainName name of the domain
+     * @param serviceName name of the service
+     */
+    public ZTSClient(String ztsUrl, String proxyUrl, SSLContext sslContext, String domainName, String serviceName) {
```
havetisyan: I'm not sure if this is the right thing to do. We don't verify what's in the sslcontext so just to blindly take the domain name and service name and assume the sslcontext is for that service is not correct. Let me take a look at the original lira ticket to see what was the issue reported.

PR #430, opened 2018-03-23
Return DENY_ROLETOKEN_EXPIRED instead of DENY_ROLETOKEN_INVALID when the token is expired

tatyano:
Since the conditions are set like this, the access check status 
 of DENY_ROLETOKEN_EXPIRED is never used.
https://github.com/yahoo/athenz/blob/4ddb8076490381099ea70468daed775215186857/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java#L433-L448

I suppose setting allowNoExpiry or ATHENZ_TOKEN_NO_EXPIRY as true means allowing the token to not expire and stay valid.

### DIFF
--- a/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
+++ b/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
@@ -386,6 +386,24 @@ public class AuthZpeClient {
             if (rToken.validate(getZtsPublicKey(rToken.getKeyId()), allowedOffset, false, null) == false) {
                 LOG.error("allowAccess: Authorization denied. Authentication of token failed for token="
                         + rToken.getSignedToken());
+
+                // check the token expiration
+                long now    = System.currentTimeMillis() / 1000;
+                long expiry = rToken.getExpiryTime();
+                if (expiry != 0 && expiry < now) {
+                    String signedToken = rToken.getSignedToken();
+                    LOG.error("allowAccess: Authorization denied. Token expired. now=" +
+                            now + " expiry=" + expiry + " token=" + signedToken);
+                    try {
+                        tokenCache.remove(signedToken);
+                    } catch (Exception exc) {
+                        LOG.error("allowAccess: token cache failure, exc: " + exc.getMessage());
+                    }
+
+                    zpeMetric.increment(ZpeConsts.ZPE_METRIC_NAME_EXPIRED_TOKEN, rToken.getDomain());
+                    return AccessCheckStatus.DENY_ROLETOKEN_EXPIRED;
+                }
+
                 zpeMetric.increment(ZpeConsts.ZPE_METRIC_NAME_INVALID_TOKEN, rToken.getDomain());
                 return AccessCheckStatus.DENY_ROLETOKEN_INVALID;
             }
--- a/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
+++ b/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
@@ -389,15 +389,8 @@ public class AuthZpeClient {
                 long now    = System.currentTimeMillis() / 1000;
                 long expiry = rToken.getExpiryTime();
                 if (expiry != 0 && expiry < now) {
-                    String signedToken = rToken.getSignedToken();
                     LOG.error("allowAccess: Authorization denied. Token expired. now=" +
-                            now + " expiry=" + expiry + " token=" + signedToken);
-                    try {
-                        tokenCache.remove(signedToken);
-                    } catch (Exception exc) {
-                        LOG.error("allowAccess: token cache failure, exc: " + exc.getMessage());
-                    }
-
+                            now + " expiry=" + expiry + " token=" + rToken.getSignedToken());
                     zpeMetric.increment(ZpeConsts.ZPE_METRIC_NAME_EXPIRED_TOKEN, rToken.getDomain());
                     return AccessCheckStatus.DENY_ROLETOKEN_EXPIRED;
                 }
--- a/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
+++ b/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
@@ -384,8 +384,6 @@ public class AuthZpeClient {
 
             // validate the token
             if (rToken.validate(getZtsPublicKey(rToken.getKeyId()), allowedOffset, false, null) == false) {
-                LOG.error("allowAccess: Authorization denied. Authentication of token failed for token="
-                        + rToken.getSignedToken());
 
                 // check the token expiration
                 long now    = System.currentTimeMillis() / 1000;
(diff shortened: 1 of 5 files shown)


libs/java/auth_core/src/test/java/com/yahoo/athenz/auth/token/TokenTest.java
```diff
@@ -116,15 +116,15 @@ public void testTokenValidateNoExpiryTimestamp() throws CryptoException {
...
         PrincipalToken token = new PrincipalToken.Builder(svcVersion, svcDomain, svcName)
             .host(host).salt(salt).issueTime(timestamp).build();
-        token.expiryTime = 0;
+        token.expiryTime = 10;
         token.sign(servicePrivateKeyStringK0);
 
         Token.ATHENZ_TOKEN_NO_EXPIRY = false;
         assertFalse(token.validate(servicePublicKeyStringK0, 20, false));
-        assertFalse(token.validate(servicePublicKeyStringK0, 20, true));
+        assertTrue(token.validate(servicePublicKeyStringK0, 20, true));
```
tatyano: I suppose the result at here should be true since allowNoExpiry is set as true.

libs/java/auth_core/src/test/java/com/yahoo/athenz/auth/token/TokenTest.java
```diff
@@ -116,15 +116,15 @@ public void testTokenValidateNoExpiryTimestamp() throws CryptoException {
...
         
         Token.ATHENZ_TOKEN_NO_EXPIRY = true;
-        assertFalse(token.validate(servicePublicKeyStringK0, 20, false));
+        assertTrue(token.validate(servicePublicKeyStringK0, 20, false));
```
tatyano: I suppose the result at here should be true since ATHENZ_TOKEN_NO_EXPIRY is set as true.

libs/java/auth_core/src/test/java/com/yahoo/athenz/auth/token/TokenTest.java
```diff
@@ -116,15 +116,15 @@ public void testTokenValidateNoExpiryTimestamp() throws CryptoException {
         long timestamp = System.currentTimeMillis() / 1000;
         PrincipalToken token = new PrincipalToken.Builder(svcVersion, svcDomain, svcName)
             .host(host).salt(salt).issueTime(timestamp).build();
-        token.expiryTime = 0;
+        token.expiryTime = 10;
```
tatyano: Setting expiryTime as 0 will make the condition to be always true so I changed it to 10.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/Token.java
```diff
@@ -133,7 +133,7 @@ public boolean validate(PublicKey publicKey, int allowedOffset, boolean allowNoE
         // explicitly enabled that option for our system. by default
         // they should have an expiration date of less than 30 days
         
-        if (expiryTime != 0 || !ATHENZ_TOKEN_NO_EXPIRY || !allowNoExpiry) {
+        if (expiryTime != 0 && !ATHENZ_TOKEN_NO_EXPIRY && !allowNoExpiry) {
```
havetisyan: We cannot change ||'s to &&. That would introduce a security issue where the library will no longer validate expiry times. That's not the job of subclasses. The implementation requires that if an expiry time is specified (regardless of what other options are configured), then it must be validated. If you make those ||'s to &&'s then if I specify an expired token with the option that the caller is allowed to accept tokens with no expiry, then the token will be returned as valid. This method is used by all authorities, etc. So it's not just in use by ZPE. So the correct implementation is correct and will not be changed.

havetisyan:
the referenced method already assumes the caller has validated the role token however, if the caller caches the validated role tokens, the zpe library still verifies that the token is valid. So I believe the code is correct.

I believe what you're looking for is a change to the following line:
https://github.com/yahoo/athenz/blob/4ddb8076490381099ea70468daed775215186857/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java#L390

Instead of returning invalid token, we should check to see if the token has expired and return that code instead of invalid token response.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/Token.java
tatyano: I was thinking of fixing each authority coinciding with this opportunity, but after seeing your proposal, I've decided to leave it as is.

clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
```diff
@@ -386,6 +386,24 @@ public static AccessCheckStatus allowAccess(String roleToken, String angResource
...
                         + rToken.getSignedToken());
+
+                // check the token expiration
+                long now    = System.currentTimeMillis() / 1000;
+                long expiry = rToken.getExpiryTime();
+                if (expiry != 0 && expiry < now) {
+                    String signedToken = rToken.getSignedToken();
+                    LOG.error("allowAccess: Authorization denied. Token expired. now=" +
+                            now + " expiry=" + expiry + " token=" + signedToken);
+                    try {
```
havetisyan: This cache remove code is not necessary here since we know that the token is not in the cache. We only get within this if block if the entry is not in the cache.

clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
```diff
@@ -386,6 +386,24 @@ public static AccessCheckStatus allowAccess(String roleToken, String angResource
             if (rToken.validate(getZtsPublicKey(rToken.getKeyId()), allowedOffset, false, null) == false) {
                 LOG.error("allowAccess: Authorization denied. Authentication of token failed for token="
                         + rToken.getSignedToken());
+
+                // check the token expiration
+                long now    = System.currentTimeMillis() / 1000;
+                long expiry = rToken.getExpiryTime();
+                if (expiry != 0 && expiry < now) {
+                    String signedToken = rToken.getSignedToken();
+                    LOG.error("allowAccess: Authorization denied. Token expired. now=" +
```
havetisyan: Let's move the LOG.error from line 387 after this expiry block check. Otherwise, we'll be logging the failure twice.

clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
tatyano: When we think about calling the method more than once, if the rToken in tokenCache expires, I believe we should remove the token from the cache since it's no longer valid.
Not only that, I thought this was necessary to keep the flow consistent since we also have it here:
https://github.com/yahoo/athenz/blob/4ddb8076490381099ea70468daed775215186857/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java#L441

havetisyan: If the token is in the cache, then that code is not executed. If you look at the function it has:

// look in the cache
RoleToken rToken = null;

// if not in the cache 

if (roleToken == null) {
   // then we parse and validate
   // new code is within this block
}

the new code is inside that if block where we already determined that the token is not in the cache. So there is no point of calling to remove it from the cache.

havetisyan:
Thanks Chris.

PR #431, opened 2018-03-25
Fixed some bugs and improved tests for Node.js libraries

tatyano:
- ad4cbb86 Check the expiration of the token before returning DENY_ROLETOKEN_INVALID (regarding #430)
- c3d70d84 Improved tests for Node.js libraries
- b1ff9571 Added fs.watch() to zpeClient of Node.js
- a5392f60 Added interface to configure ZPEUpdater class in zpe-client of Node.js
- 8f438deb Added close() function to zpe-client of Node.js


havetisyan:
👍

PR #433, opened 2018-03-27
auto cleanup of expired cert records

havetisyan:
ZTS needs to auto cleanup any cert records that are longer than the configured expiry time for the http signer. Those are already expired so there is no point of keeping them and increasing the db size. It will have a daily thread to delete all expired cert records.


charlesk40:
👍

PR #435, opened 2018-03-27
reuse s3 client during initial load

havetisyan:
During initial load of domains from S3, we don't need to create a new s3 client for every domain. By reusing the client, loading of 690 domains goes down from 49 seconds to 22 seconds.


mujibur:
👍

PR #437, opened 2018-03-30
correct terminology and timeout configurable

havetisyan:
replace athensPrivateKey with athenzPrivateKey to correct product name

replace athensPublicKey with athenzPublicCert to correct termonology

make key wait time configurable


hawk259:
👍

PR #439, opened 2018-04-03
add read only mode of operation to ZTS

havetisyan:
if athenz.zts.read_only_mode property is set to true then all post operations for x.509 certificates will be rejected since those require write access to the certificate record database.

this is useful when there are db operations that need to be carried that would require zts to be in read only mode (we already have same functionality in zms server).


hawk259:
👍

PR #445, opened 2018-04-13
ZTS should not send empty policy data since the client must expand the expiration periodically.

tatyano:
I believe ZTS should not send empty policy data since the client must expand the expiration periodically.

When ZPU receives an empty data, it doesn't update the policy file, nor the expiration.
https://github.com/yahoo/athenz/blob/86165415e025ce5f3c6ba862c04e2cb3c2ec1b53/utils/zpe-updater/zpu_client.go#L90-L93

I think it's ok to send empty policy data with "HTTP/1.1 304 Not Modified" if the zpe client is not checking the expiration of the policy file.
And the current implementation of this open-source seems to not be checking it.
https://github.com/yahoo/athenz/search?utf8=%E2%9C%93&q=DENY_DOMAIN_EXPIRED&type=

However, the previous implementation does so, I think we should keep the backward compatibility.

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -885,14 +885,6 @@ public class ZTSImpl implements KeyStore, ZTSHandler {
         EntityTag eTag = new EntityTag(modified.toString());
         String tag = eTag.toString();
         
-        // Set timestamp for domain rather than youngest policy.
-        // Since a policy could have been deleted, and can only be detected
-        // via the domain modified timestamp.
-        
-        if (matchingTag != null && matchingTag.equals(tag)) {
-            signedPoliciesResult.done(ResourceException.NOT_MODIFIED, matchingTag);
-        }
-        
         // first get our PolicyData object
         
         PolicyData policyData = new PolicyData()


havetisyan:
No, this is not a correct change. ZTS does not need to sign and send back the data every time. There are clients that periodically check for any updates like every minute so we can't sign and return the data every time a request comes in.

It seems to be you're referring to a possibly a wrong logic in zpu in which case that needs to be fixed and not this change in ZTS.

tatyano:
Yes, I understand that we shouldn't sign and return the new data every time a request comes in.

However, some ZPE clients are checking the "expiry" field in the policy file during the authorization process, hence we also must update the policy file's expiration before it really expires.
In order to prevent the policy file from expiration, the only way I could find was to update the entire policy file frequently.
At this point, I also found another obstacle, that there is no way for ZPU to update the policy file's expiration if ZTS sends empty data.

Hence, I figured the only two ways to prevent the unauthorized response from ZPE client are:
- To fix ZPE client not to check the policy file's expiration during the authorization check process.
- To fix ZTS to send the entire policy data(including expiration) every time so ZPU can update the file.

And, as an open-source, the only option I think we have is the second one.

havetisyan:
It is ZPU's responsibility to update the policy before it expires and I believe it does it correctly. Before asking ZTS Server for an update, it checks to see what is the ETag to be sent to the server:

https://github.com/yahoo/athenz/blob/master/utils/zpe-updater/zpu_client.go#L143

The default config expiry check is set to 2 days. So the policy expires in 7 days. Even if zpu is configured to run daily, 2 days before the policy expires, it will send the request to ZTS with no Etag which will cause the full policy file with a new expiration date to be sent back to the client. And this should cause ZPE to reload the file with a new expiration date.

So which part is not working in your enviornment?

tatyano:
Thank you for your prompt reply.
After reading it, I see my concern was for nothing.

Now I know that the default config expiry check is set to 2 days, the only question I have left: Is it intentional for the implementation to lock the config expiry check to 2 days?
https://github.com/yahoo/athenz/blob/master/utils/zpe-updater/zpu_config.go#L125
Since ZPU can be run in any frequency that client configures(crontab or etc.), I thought it would be more useful for the users to be able to configure the frequency of the expiry check.

PR #447, opened 2018-04-14
cert refresher fixes

havetisyan:
2 fixes for the cert refresher logic when checking for file updates

a) it was correctly storing the checksum so it would treat the file as changed
b) if the file is a relative path, it came from resources and as such it should not check for any updates.


charlesk40:
👍

PR #448, opened 2018-04-17
aws temporary creds enhancements

havetisyan:
2 enhancements to the AWS Temporary credentials:

a) if the assume role statement in aws is configued with an external id condition, ZTS server now support this. the zts java and go client apis have been extended to support an additional externalId argument that the caller can specify for improved security of their accounts.

b) aws now supports temp creds of up to 12 hours instead of the previous 1 hour. You can configure this extended limit in your IAM role configuration page. Similarly, ZTS now allows the caller to specify the maximum token timeout value that must be equal or less than the configured IAM value (otherwise STS will reject the request).

### DIFF
--- a/clients/go/zts/client.go
+++ b/clients/go/zts/client.go
@@ -687,9 +687,9 @@ func (client ZTSClient) PostInstanceRefreshRequest(domain CompoundName, service 
 	}
 }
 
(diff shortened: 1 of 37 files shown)


hawk259:
👍

Issue #455, opened 2018-04-29
remove use of system environment variables in servers


havetisyan:
Environment variables are a notorious source of risk and their use should be minimized or
eliminated wherever possible. Where their use is absolutely necessary, the values
of environment variables should be treated as untrusted so they must sanitized and validated
prior use.

havetisyan:
#460

PR #459, opened 2018-05-01
 #455 remove use of system environment variables in servers

havetisyan:
plus more minor cleanup of variables scopes to remove static code coverage check warnings from InteilliJ


charlesk40:
👍

PR #471, opened 2018-05-19
Use apache connector provider for zts client

havetisyan:
Before we were using apache connector only when proxy option was set but we should use it always.


mujibur:
👍

PR #493, opened 2018-06-06
rate limit handling for get zms domain data

havetisyan:
when zms server has rate limiting enabled and rejects get domain data request from zts with 429, instead of just failing and skipping the domain, zts should retry the operation after a second and keep retrying until it is successful.


charlesk40:
👍

PR #495, opened 2018-06-11
mark sig/key id as optional

havetisyan:
when the caller requests get signed api with metaonly flag set to true, zms server does not sign any data since there is no data to sign. Thus, for signature and key id, it returns null values. However, those fields are set as required in the rdl so the go client which validates data received from the server, rejects the response as invalid since those values are not preset. those two fields should be marked as optional in the response.

### DIFF
--- a/core/zms/src/main/rdl/SignedDomains.rdli
+++ b/core/zms/src/main/rdl/SignedDomains.rdli
@@ -49,11 +49,13 @@ type DomainData Struct {
     String applicationId (optional); // associated application id
 }
 
-//A domain object signed with server's private key 
+//A domain object signed with server's private key. The signature
+//and keyid are optional if the metaonly flag is set to true in the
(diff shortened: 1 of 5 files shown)


core/zms/src/main/rdl/SignedDomains.rdli
```diff
@@ -49,11 +49,13 @@ type DomainData Struct {
     String applicationId (optional); // associated application id
 }
 
-//A domain object signed with server's private key 
+//A domain object signed with server's private key. The signature
+//and keyid are optional if the metaonly flag is set to true in the
```
xythian: They're never present when metaonly is true, correct?

havetisyan: yes, if metonly is true, they're never present.

PR #497, opened 2018-06-12
provide property for jdbc app name

havetisyan:
instead of hardcoding the value for appname when fetching db app secret to JDBC, we now provide the capability to extract the appname from a system property.


mujibur:
👍

PR #498, opened 2018-06-13
SolutionTemplate should be able to use placeholders in names

tatyano:
As in the current implementation, we cannot use placeholders in the names of role, policy or service. 
For example "builders" cannot be defined with a placeholder. (If we do, the API just doesn't work.)
https://github.com/yahoo/athenz/blob/master/servers/zms/conf/solution_templates.json#L14

However, the intended behavior as the implementation seems to be able to resolve the placeholders in the name field.
https://github.com/yahoo/athenz/blob/c34e24c8ba4c1b23c83faf46dbd9b66be62428ba/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java#L2292-L2298

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
@@ -2139,7 +2139,7 @@ public class DBService {
                 Role templateRole = updateTemplateRole(role, domainName, roleName, templateParams);
                 firstEntry = auditLogSeparator(auditDetails, firstEntry);
                 auditDetails.append(" \"add-role\": ");
-                if (!processRole(con, originalRole, domainName, roleName, templateRole,
+                if (!processRole(con, originalRole, domainName, ZMSUtils.removeDomainPrefix(templateRole.getName(), domainName, ROLE_PREFIX), templateRole,
                         admin, auditRef, true, auditDetails)) {
                     return false;
                 }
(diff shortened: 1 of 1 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2139,7 +2139,7 @@ boolean addSolutionTemplate(ObjectStoreConnection con, String domainName, String
                 Role templateRole = updateTemplateRole(role, domainName, roleName, templateParams);
                 firstEntry = auditLogSeparator(auditDetails, firstEntry);
                 auditDetails.append(" \"add-role\": ");
-                if (!processRole(con, originalRole, domainName, roleName, templateRole,
+                if (!processRole(con, originalRole, domainName, ZMSUtils.removeDomainPrefix(templateRole.getName(), domainName, ROLE_PREFIX), templateRole,
```
havetisyan: I don't believe the change is sufficient to correctly handle variable substitutions in the name. As part of the process command we pass the original role object that was retreived without taking into account the substitution. So while the first template apply command will work fine because the original role does not exist thus it's a new add operation, if you try to re-apply the same template again, we'll pick up and process a wrong original name thus we'll try to execute an add operation again instead of modify. So we need to take into account the substituted name when we retrieve the original role as well.

Same change should be applied to both policy and service blocks as well.

havetisyan:
@tatyano didn't see any updates. #510 will address this issue.

tatyano:
Sorry for not replying.
Thank you so much for taking care of it.

PR #504, opened 2018-06-28
Add key service check for public key fetch

mcieplak:
This PR is to keep the public key fetch inline with the java codebase here: https://github.com/yahoo/athenz/blob/master/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/PrincipalAuthority.java#L229-L240. This will allow project tokens to fetch the public key from zts from the sys.auth domain.

### DIFF
--- a/libs/go/zmssvctoken/token.go
+++ b/libs/go/zmssvctoken/token.go
@@ -31,6 +31,7 @@ const (
 	tagExpireTime     = "e"
 	tagSalt           = "a"
 	tagSignature      = "s"
+	tagKeyService     = "z"
 )
 
 // NToken provides access to useful fields in an ntoken.
@@ -39,6 +40,7 @@ type NToken struct {
 	Domain         string    // domain for which token is valid
 	Name           string    // local principal name
 	KeyVersion     string    // key version as registered in Athenz
+	KeyService     string    // optional key service
 	Hostname       string    // optional hostname
 	IPAddress      string    // optional IP address
 	GenerationTime time.Time // time token was generated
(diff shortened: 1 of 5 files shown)


libs/go/zmssvctoken/token.go
```diff
@@ -229,8 +237,9 @@ type tokenBuilder struct {
 
 // NewTokenBuilder returns a TokenBuilder implementation for the specified
 // domain/name, with a private key (PEM format) and its key-version. The key-version
-// should be the same string that was used to register the key with Athenz.
-func NewTokenBuilder(domain, name string, privateKeyPEM []byte, keyVersion string) (TokenBuilder, error) {
+// should be the same string that was used to register the key with Athenz. The key-service
+// (optional) should be the service name where the public key is stored.
+func NewTokenBuilder(domain, name string, privateKeyPEM []byte, keyVersion, keyService string) (TokenBuilder, error) {
```
havetisyan: Why was keyService added to the NewTokenBuilder api? This has 2 issues that we should avoid:
a) breaks existing clients who already use the api. From client's perspective, they're always going to pass "" since no client has access to the zms/zts keys to sign anything
b) it's an optional component so it must be treated like all optional components in the ntoken with the appropriate SetKeyService endpoint. (similar to SetHostname/SetIPAddress,etc)

mcieplak: @havetisyan That's a good point. I have updated the code so that this parameter is considered optional and have removed it from the NewTokenBuilder api. Can you please take a look again? Thanks.

havetisyan:
Thanks @mcieplak

PR #505, opened 2018-06-28
produce json for error conditions

havetisyan:
We should have @Produces(MediaType.APPLICATION_JSON) for all our endpoints even if they produce no content because all errors are reported as json so if we don't have annotation, then jetty tries to generate text/plain output and fails with "MessageBodyWriter not found for media type=text/plain, type=class" error message. So for all errors we get back 500 instead of the actual error code.


hawk259:
👍

PR #506, opened 2018-06-29
handle cache when duration is <=0

havetisyan:
if duration seconds specified for aws temp creds is either 0 or < 0 then we always default back to 1 hour so handle that case for cache support as well.


mujibur:
👍

PR #512, opened 2018-07-16
text handling of error response from providers

havetisyan:
simplify error handling from providers. Any response text returned from provider - treat as regular text and log error instead of deserializing into json object - a provider may not return the data in proper format and server can get exceptions when deserializing thus returning confusing error back to the client.


charlesk40:
👍

PR #513, opened 2018-07-16
allow config option to reset cert refresh db

havetisyan:
During certificate refresh operations zts server looks up the original certificate details (serial number, timestamp, etc) to detect compromise. If this database is lost, then server will not be able to refresh any certs, so we provide an option to regenerate the db based on requests rather than rejecting all. So while the certs are refreshed, compromise will not be detected during the first refresh, but during the second one it will be detected. The value of the setting is the number of milliseconds since epoch. Any refresh request where the cert has a timestamp before this date will be handled successfully if the db record is not found.


psasidhar:
👍

PR #515, opened 2018-07-21
config option for reserved service names

havetisyan:


servers/zms/conf/zms.properties
```diff
@@ -350,7 +350,17 @@ athenz.zms.solution_templates_fname=/home/athenz/conf/zms_server/solution_templa
...
 # to request authorize service tokens. This is only used for the
 # optionsUserToken method where we return CORS headers
 #athenz.zms.cors_origin_list
+
+# Comma separated list of service names that are reserved. The default
+# list includes most common gTLDs so a service cannot obtain an identity
+# x.509 certificate with a name that matches an actual domain e.g. yahoo.com
+#athenz.zms.reserved_service_names=com,net,org,edu,biz,gov,mil,info,name,mobi,cloud
+
+# Integer value specifying the min length of any service names. The default
```
drewfish: Is "length" number of characters or number of words on the DNS name?

havetisyan: Service name in Athenz is SimpleName which is defined as a single component so this setting only applies to the number of characters in a service name. e.g. us is not a valid service name but use will be a valid name. The idea is that if you have a top level domain called yahoo, you should not create a service called us and then request a service x.509 certiifcate which would include the subject cn=yahoo.us

PR #516, opened 2018-07-26
Add additional counters for cert and auth operations

havetisyan:
keep track of additional counters for provider register and refresh operation along with certificate sign requests. also keep track of any auth n/z failures since those are not reflected in the server get/post request/method counters.


charlesk40:
👍

PR #517, opened 2018-07-26
ignore invalid authorities that provider no header

havetisyan:
avoid npe when the administrator has implemented an authority based on a specific header but the header method returns null for the header name.


charlesk40:
👍

PR #522, opened 2018-08-07
DynamoDB implementation for CertRecordStore interface

havetisyan:
new AWS dynamodb implementation with factory com.yahoo.athenz.zts.cert.impl.DynamoDBCertRecordStoreFactory for CertRecordStore interface.

The table must be created with the following requirements:

- primary field - instanceId,
- sort key - provider.
- Enable TTL and call the attribute as ttl.


hawk259:
👍

PR #523, opened 2018-08-09
Correcting aws jars dependencies

gurleen-gks:


containers/jetty/pom.xml
```diff
@@ -134,6 +134,20 @@
...
+        </exclusion>
+        <exclusion>
+          <groupId>com.amazonaws</groupId>
+          <artifactId>aws-java-sdk-kinesis</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>com.amazonaws</groupId>
+      <artifactId>aws-java-sdk-kinesis</artifactId>
```
havetisyan: since you're excluding both sdk-core and sdk-kinesis, let's add both as dependency instead of just kinesis (which I'm assuming has dependency on core but let's specify it explicitely).

havetisyan:
👍

PR #549, opened 2018-09-13
Address static code analysis recommendations from IntelliJ

havetisyan:
mostly removing unnecessary init values, removing duplicates (we had a couple of test cases identical).


pratikgote:
👍

PR #551, opened 2018-09-17
fix binary release links + jdbc class

havetisyan:
1. fix the binary release links from github to bintray
2. include the object store factory class property update when switching to use mysql/jdbc implementation for production use cases.


charlesk40:
👍

PR #554, opened 2018-09-21
add go.mod and go.sum files

dmitris:
add go.mod and go.sum files to document dependencies.
Also minor fix of incorrect Print in an example

@havetisyan


havetisyan:
👍

PR #578, opened 2018-10-12
Generate UUID (type 1) for audit logs

charlesk40:
Add (by default) UUID type 1 (MAC address + timestamp) to audit logs, so that they are uniquely identifiable.

See https://github.com/cowtowncoder/java-uuid-generator


havetisyan:
👍

PR #603, opened 2018-11-30
Support separate private key for role csr

prabushyam:
@havetisyan 
This enables zts-rolecert utility to accept a different private key for signing the role cert CSR (defaults to using service identity key)


havetisyan:
Thanks @prabushyam

PR #617, opened 2019-01-16
Make setEnablePrefetch public

jacobtolar:
Making this method public allow users to disable prefetch at the client level (currently, it can only be disabled globally by a system property/static method).

@havetisyan


codingwhatever:
👍 Thanks @jacobtolar

PR #631, opened 2019-02-14
Update Go dependencies

jeffreytolar:
Also switch to using lumberjack's preferred import path on gopkg.in - see the note at the top of https://github.com/natefinch/lumberjack

I'm not sure where some of the modules that this removes came from; perhaps from a stray `go get` run from the Athenz repo?


yahoocla:

PR #651, opened 2019-03-22
docs: minor markdown formatting changes

dmitris:
PR makes the link to https://github.com/yahoo/athenz/tree/master/examples/go/centralized-use-case/server be actual HTML link (rather than just piece of text) and couple of other minor editorial changes ("thru" -> "through", `cd` in backticks).  A number of whitespace changes (truncations) as "collateral damage".


havetisyan:
👍

PR #660, opened 2019-04-14
move oauth2 token classes to auth_core from zts_core

havetisyan:
For structured JWT access token format - we're now following - https://tools.ietf.org/html/draft-bertocci-oauth-access-token-jwt-00


psasidhar:
👍

PR #684, opened 2019-05-07
Validate Audit Reference for Audit Enabled Domains

pratikgote:


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSConsts.java
```diff
@@ -96,6 +96,8 @@
     public static final String ZMS_PROP_AUDIT_LOGGER_FACTORY_CLASS      = "athenz.zms.audit_logger_factory_class";
     public static final String ZMS_AUDIT_LOGGER_FACTORY_CLASS           = "com.yahoo.athenz.common.server.log.impl.DefaultAuditLoggerFactory";
 
+    public static final String ZMS_PROP_AUDIT_REF_FACTORY_CLASS      = "athenz.zms.audit_logger_factory_class";
```
havetisyan: this has the same value as the ZMS_PROP_AUDIT_LOGGER_FACTORY_CLASS.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -700,6 +707,26 @@ void loadAuditLogger() {
         
         auditLogger = auditLogFactory.create();
     }
+
+    void loadAuditRefValidator() {
+        String auditRefValidatorClass = System.getProperty(ZMSConsts.ZMS_PROP_AUDIT_REF_FACTORY_CLASS);
```
havetisyan: make the String final.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/audit/AuditReferenceFactory.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.yahoo.athenz.zms.store.audit;
+
+public interface AuditReferenceFactory {
```
havetisyan: Let's call this as AuditReferenceValidatorFactory to be consistent with the object it creates.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/audit/AuditReferenceValidator.java
```diff
@@ -0,0 +1,28 @@
...
+public interface AuditReferenceValidator {
+
+    /**
+     * Validate the Audit Reference for audit enabled domains
+     * @param auditRef - Audit Reference provided for the change requested
+     * @param principal - Domain principal for which the Audit Reference will be validated
+     * @param operation - Operation requiring Audit Reference Validation
+     * @return true if valid, else false
+     */
+    boolean validateAudit(String auditRef, String principal, String operation);
```
havetisyan: let's call this validateReference as opposed to validateAudit to be more clear what it does.

PR #705, opened 2019-06-14
adding support for role level system meta attributes

abvaidya:
added a new column in db schema for role table
modified rdl and generated stubs are committed
modified zmscli go lib to reflect new command to set role audit enabled
added implementation for new api
added wrapper method in zms-java-client


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -3454,4 +3476,61 @@ public Quota getQuota(String domainName) {
...
+
+                // first verify that auditing requirements are met
+
+                Domain domain = checkDomainAuditEnabled(con, domainName, auditRef, caller, getPrincipalName(ctx));
+
+                if (!domain.getAuditEnabled()) {
+                    throw ZMSUtils.requestError(caller + ": auditenabled flag not set for domain: " + domainName + " to add it on the role: " + roleName, caller);
+                }
+
+                Role rolefromdb = getRole(domainName, roleName, false, true);
```
havetisyan: the last argument should be false as we don't need to expand any users

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -3454,4 +3476,61 @@ public Quota getQuota(String domainName) {
...
+
+                // now process the request. first we're going to make a
+                // copy of our domain
```
havetisyan: comment - copy of our role

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
abvaidya: fixed

PR #711, opened 2019-06-19
ZMS cannot start after DDL update

WindzCUHK:
SQL column name does not match. ZMS cannot start up.
https://github.com/yahoo/athenz/blob/master/servers/zms/schema/zms_server.sql#L54

> I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.


havetisyan:
Thanks @WindzCUHK. We noticed the same failure in our functional test pipeline as well, but your PR came first.

PR #715, opened 2019-06-22
adding feature to ignore unknown properties in zms & zts clients

abvaidya:
only setting auditEnabled on Role object when DB returns true

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
@@ -1050,14 +1050,11 @@ public class JDBCConnection implements ObjectStoreConnection {
             ps.setString(2, roleName);
             try (ResultSet rs = executeQuery(ps, caller)) {
                 if (rs.next()) {
-                    Role r = new Role();
-                    r.setName(ZMSUtils.roleResourceName(domainName, roleName))
+                    return new Role().setName(ZMSUtils.roleResourceName(domainName, roleName))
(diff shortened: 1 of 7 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -1050,10 +1050,14 @@ public Role getRole(String domainName, String roleName) {
             ps.setString(2, roleName);
             try (ResultSet rs = executeQuery(ps, caller)) {
                 if (rs.next()) {
-                    return new Role().setName(ZMSUtils.roleResourceName(domainName, roleName))
-                            .setModified(Timestamp.fromMillis(rs.getTimestamp(ZMSConsts.DB_COLUMN_MODIFIED).getTime()))
-                            .setTrust(saveValue(rs.getString(ZMSConsts.DB_COLUMN_TRUST)))
-                            .setAuditEnabled(rs.getBoolean(ZMSConsts.DB_COLUMN_AUDIT_ENABLED));
+                    Role r = new Role();
```
havetisyan: I'd like to propose a slightly different implementation to make our intentions clear. I would keep the original code but replace the last line:

.setAuditEnabled(rs.getBoolean(ZMSConsts.DB_COLUMN_AUDIT_ENABLED));

with

.setAuditEnabled(nullIfDefaultValue(rs.getBoolean(ZMSConsts.DB_COLUMN_AUDIT_ENABLED), false));

and then have:

    Boolean nullIfDefaultValue(boolean flag, boolean defaultValue) {
        return flag == defaultValue ? null : flag;
    }

this way a reader would know that our intentions is to use null when we have the default value as opposed to trying to understand why we're not setting the value false.

PR #732, opened 2019-07-22
role membership for audit enabled and selfserve roles

abvaidya:
new active state in role_member table
new api to approve or reject membership
new api to set role level meta object


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -3537,4 +3548,137 @@ public void executePutRoleSystemMeta(ResourceContext ctx, String domainName, Str
...
+        // and throws the exception it had received when the retry
+        // count reaches 0
+
+        for (int retryCount = defaultRetryCount; ; retryCount--) {
+
+            try (ObjectStoreConnection con = store.getConnection(true, true)) {
+
+                String principal = getPrincipalName(ctx);
+
+                // process our insert role member support. since this is a "single"
```
havetisyan: let's fix the comment about insert role member

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -3537,4 +3548,137 @@ public void executePutRoleSystemMeta(ResourceContext ctx, String domainName, Str
...
+
+                // update our role and domain time-stamps, and invalidate local cache entry
+
+                con.updateRoleModTimestamp(domainName, roleName);
+                con.updateDomainModTimestamp(domainName);
+                cacheStore.invalidate(domainName);
+
+                // audit log the request
+
+                auditLogRequest(ctx, domainName, auditRef, caller, ZMSConsts.HTTP_PUT,
```
havetisyan: this doesn't seem sufficient since it doesn't say if the request was approved or denied.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6644,4 +6657,198 @@ public void putRoleSystemMeta(ResourceContext ctx, String domainName, String rol
...
+        roleMember.setActive(membership.getActive());
+
+        dbService.executePutMembershipDecision(ctx, domainName, roleName, getNormalizedMember(roleMember), auditRef, caller);
+
+        metric.stopTiming(timerMetric, domainName, principalDomain);
+    }
+
+    private boolean isAllowedPutMembershipDecision(Principal principal, String domainName, String roleName) {
+        AthenzDomain domain = getAthenzDomain(domainName, false);
+        if (domain != null) {
```
havetisyan: we shouldn't need this check. we should get the athenz domain with the option to throw exceptions if the domain is not found.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6644,4 +6657,198 @@ public void putRoleSystemMeta(ResourceContext ctx, String domainName, String rol
...
+    }
+
+    boolean isAllowedPutMembershipAccess(Principal principal, final String reqDomain, final String roleName, final AthenzDomain domain) {
+
+
+        if (domain != null) {
+            // evaluate our domain's roles and policies to see if access
+            // is allowed or not for the given operation and resource
+            // our action are always converted to lowercase
+            String resource = reqDomain + ":role." + roleName;
```
havetisyan: we should mark the resource as final

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -110,9 +111,11 @@
     private static final String SQL_LIST_PRINCIPAL = "SELECT * FROM principal;";
     private static final String SQL_LIST_PRINCIPAL_DOMAIN = "SELECT * FROM principal WHERE name LIKE ?;";
     private static final String SQL_LAST_INSERT_ID = "SELECT LAST_INSERT_ID();";
-    private static final String SQL_INSERT_ROLE_MEMBER = "INSERT INTO role_member (role_id, principal_id, expiration) VALUES (?,?,?);";
+    private static final String SQL_INSERT_ROLE_MEMBER = "INSERT INTO role_member (role_id, principal_id, expiration, active) VALUES (?,?,?,?);";
     private static final String SQL_DELETE_ROLE_MEMBER = "DELETE FROM role_member WHERE role_id=? AND principal_id=?;";
+    private static final String SQL_DELETE_INACTIVE_ROLE_MEMBER = "DELETE FROM role_member WHERE role_id=? AND principal_id=? AND active=FALSE;";
```
havetisyan: we should be consistent and use false since we've used lowercase everywhere else

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3355,6 +3363,89 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+        String principal = roleMember.getMemberName();
+        int domainId = getDomainId(domainName);
+        if (domainId == 0) {
+            throw notFoundError(caller, ZMSConsts.OBJECT_DOMAIN, domainName);
+        }
+        int roleId = getRoleId(domainId, roleName);
+        if (roleId == 0) {
+            throw notFoundError(caller, ZMSConsts.OBJECT_ROLE, ZMSUtils.roleResourceName(domainName, roleName));
+        }
+        if (!validatePrincipalDomain(principal)) {
```
havetisyan: if we're confirming a role member, there is no need to validate domain nor insert the principal if one doesn't exist since those are checks done by the insert call. so 3382-3384 and 3387 should not be necessary

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3355,6 +3363,89 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+            ps.setInt(2, principalId);
+            try (ResultSet rs = executeQuery(ps, caller)) {
+                if (rs.next()) {
+                    roleMemberExists = true;
+                }
+            }
+        } catch (SQLException ex) {
+            throw sqlError(ex, caller);
+        }
+        java.sql.Timestamp expiration = null;
```
havetisyan: at this time we should check if roleMemberExists or not and not on line 3411. if it doesn't exist, then throw an exception but not an internal error - it's a 404 so a not found error

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3355,6 +3363,89 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+        if (roleMember.getExpiration() != null) {
+            expiration = new java.sql.Timestamp(roleMember.getExpiration().toDate().getTime());
+        }
+        if (roleMemberExists) {
+
+            if (roleMember.getActive() != null && roleMember.getActive() == Boolean.TRUE) {
+
+                try (PreparedStatement ps = con.prepareStatement(SQL_APPROVE_ROLE_MEMBER)) {
+                    ps.setTimestamp(1, expiration);
+                    ps.setBoolean(2, roleMember.getActive());
```
havetisyan: since this is an approve call and we've verified getActive() is true, why have this as argument. The SQL_APPROVE_ROLE_MEMBER command should hardcode the value to be true and not take it as an argument

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3355,6 +3363,89 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+                    ps.setInt(3, roleId);
+                    ps.setInt(4, principalId);
+                    affectedRows = executeUpdate(ps, caller);
+                } catch (SQLException ex) {
+                    throw sqlError(ex, caller);
+                }
+                result = (affectedRows > 0);
+                if (result) {
+                    result = insertRoleAuditLog(roleId, admin, principal, "UPDATE", auditRef);
```
havetisyan: we should call the action APPROVE instead of UPDATE

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3355,6 +3363,89 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+                try (PreparedStatement ps = con.prepareStatement(SQL_DELETE_INACTIVE_ROLE_MEMBER)) {
+                    ps.setInt(1, roleId);
+                    ps.setInt(2, principalId);
+                    affectedRows = executeUpdate(ps, caller);
+                } catch (SQLException ex) {
+                    throw sqlError(ex, caller);
+                }
+                result = (affectedRows > 0);
+                if (result) {
+                    result = insertRoleAuditLog(roleId, admin, principal, "DELETE", auditRef);
```
havetisyan: we should call the action as REJECT instead of DELETE

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3355,6 +3363,89 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+        } catch (SQLException ex) {
+            throw sqlError(ex, caller);
+        }
+        java.sql.Timestamp expiration = null;
+        if (roleMember.getExpiration() != null) {
+            expiration = new java.sql.Timestamp(roleMember.getExpiration().toDate().getTime());
+        }
+        if (roleMemberExists) {
+
+            if (roleMember.getActive() != null && roleMember.getActive() == Boolean.TRUE) {
```
havetisyan: no need for the != null check since == Boolean.TRUE will be sufficient

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -1501,4 +1501,46 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+
+        DomainStruct domainStruct = getDomainStruct(domainName);
+        if (domainStruct == null) {
+            throw ZMSUtils.error(ResourceException.NOT_FOUND, "domain not found", "confirmRoleMember");
+        }
+        Role role = getRoleObject(domainStruct, roleName);
+        if (role == null) {
+            throw ZMSUtils.error(ResourceException.NOT_FOUND, "role not found", "confirmRoleMember");
+        }
+        if (!validatePrincipalDomain(member.getMemberName())) {
```
havetisyan: no need for this check here since we're going to look for the inactive member before making any changes.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -1501,4 +1501,46 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+            // need to check if the member already exists
+
+            Iterator<RoleMember> rmIter = role.getRoleMembers().iterator();
+            RoleMember roleMember;
+            while (rmIter.hasNext()) {
+                roleMember = rmIter.next();
+                // check whether the member exists and is in inactive state
+                if (roleMember.getMemberName().equals(member.getMemberName()) && !roleMember.getActive()) {
+                    //if membership is approved, set rolemember to active
+                    if (member.getActive() != null && member.getActive() == Boolean.TRUE) {
```
havetisyan: no need to check for null, == Boolean.TRUE is sufficient

PR #746, opened 2019-08-09
Adding a new API to get pending members for audit enabled or selfserv…

abvaidya:
…e roles


libs/go/zmscli/cli.go
```diff
@@ -1530,6 +1550,28 @@ func (cli Zms) HelpSpecificCommand(interactive bool, cmd string) string {
...
+		buf.WriteString("   " + domain_param + " put-membership-decision role member expiration approval \n")
+		buf.WriteString("   " + "or \n")
+		buf.WriteString("   " + domain_param + " put-membership-decision role member approval \n")
+		buf.WriteString(" parameters:\n")
+		if !interactive {
+			buf.WriteString("   domain        : name of the domain that role belongs to\n")
+		}
+		buf.WriteString("   role    : name of the role to be modified\n")
+		buf.WriteString("   member  : name of the member\n")
+		buf.WriteString("   approval: true/false depcting whether membership is approved or rejected\n")
```
havetisyan: depicting missing i

libs/go/zmscli/role.go
```diff
@@ -313,3 +313,42 @@ func (cli Zms) SetRoleSelfserve(dn string, rn string, selfserve bool) (*string,
...
+	var member zms.Membership
+	member.MemberName = zms.MemberName(validatedUser)
+	member.RoleName = zms.ResourceName(rn)
+	member.Expiration = &expiration
+	member.Active = &approval
+	err := cli.Zms.PutMembershipDecision(zms.DomainName(dn), zms.EntityName(rn), zms.MemberName(validatedUser), cli.AuditRef, &member)
+	if err != nil {
+		return nil, err
+	}
+	s := "[domain " + dn + " role " + rn
```
havetisyan: any reason why we didn't include the rest of the common data " member " + mbr + " successfully " ?

libs/go/zmscli/role.go
```diff
@@ -313,3 +313,42 @@ func (cli Zms) SetRoleSelfserve(dn string, rn string, selfserve bool) (*string,
...
+	validatedUser := cli.validatedUser(mbr)
+	var member zms.Membership
+	member.MemberName = zms.MemberName(validatedUser)
+	member.RoleName = zms.ResourceName(rn)
+	member.Active = &approval
+	err := cli.Zms.PutMembershipDecision(zms.DomainName(dn), zms.EntityName(rn), zms.MemberName(validatedUser), cli.AuditRef, &member)
+	if err != nil {
+		return nil, err
+	}
+	s := "[domain " + dn + " role " + rn
```
havetisyan: same comment as above regarding common text

libs/java/client_common/src/main/java/com/yahoo/athenz/common/utils/SignUtils.java
```diff
@@ -70,6 +70,7 @@
     private static final String ATTR_CONTENTS = "contents";
     private static final String ATTR_CERT_DNS_DOMAIN = "certDnsDomain";
     private static final String ATTR_AUDIT_ENABLED = "auditEnabled";
+    private static final String ATTR_SELF_SERVE = "selfserve";
```
havetisyan: we should keep the camel case format - selfServe

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6844,4 +6844,24 @@ boolean isAllowedPutMembership(Principal principal, final AthenzDomain domain, f
...
+    public DomainRoleMembership getPendingDomainRoleMembersList(ResourceContext ctx) {
+        final String caller = "getpendingdomainrolememberslist";
+        metric.increment(ZMSConsts.HTTP_GET);
+        metric.increment(ZMSConsts.HTTP_REQUEST);
+        metric.increment(caller);
+        final String principalDomain = getPrincipalDomain(ctx);
+        Object timerMetric = metric.startTiming("getpendingdomainrolememberslist_timing", null, principalDomain);
+        logPrincipal(ctx);
+        validateRequest(ctx.request(), caller);
+        Principal ctxPrincipal = ((RsrcCtxWrapper) ctx).principal();
```
havetisyan: since we get the principal object only for debug we should included it within the if block or just reference it directly in the debug call

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3441,6 +3441,72 @@ public boolean confirmRoleMember(String domainName, String roleName, RoleMember
...
+        MemberRole memberRole;
+
+        try (PreparedStatement ps = con.prepareStatement(SQL_PENDING_DOMAIN_ROLE_MEMBER_LIST)) {
+            ps.setInt(1, principalId);
+            ps.setInt(2, auditDomId);
+            try (ResultSet rs = executeQuery(ps, caller)) {
+                while (rs.next()) {
+                    domain = rs.getString(1);
+                    domainRoleMembers = domainRoleMembersMap.get(domain);
+                    if (domainRoleMembers != null) {
```
havetisyan: there seems to be a lot of code repetition here. you can use the following pattern:
if (!domainRoleMembersMap.contains(domain)) {
    // create and insert a new List
}

then just get the list and add a new member to it.

libs/go/zmscli/role.go
abvaidya: Its included below based on the approval flag being true / false

libs/go/zmscli/role.go
abvaidya: Its included below based on the approval flag being true / false

PR #753, opened 2019-08-16
Role optional

abvaidya:


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1855,7 +1855,7 @@ Role getRole(ObjectStoreConnection con, String domainName, String roleName,
                 // if we have no trust field specified then we need to
                 // retrieve our standard group role members
                 
-                role.setRoleMembers(con.listRoleMembers(domainName, roleName));
+                role.setRoleMembers(con.listRoleMembers(domainName, roleName, pending));
```
havetisyan: after this call, for backward support we set the members list as well (which is just a list of member names). when pending value is set to true we can't include the members in the list since that gives the wrong impression so we should always exclude the pending members from the convertRoleMembersToMembers method.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -525,14 +525,32 @@ public boolean deletePrincipal(String principalName, boolean subDomains) {
...
+    Role getRoleObject(DomainStruct domain, String roleName, Boolean pending) {
+
+        Role role = getRoleObject(domain, roleName);
+        if (pending == Boolean.FALSE) {
+            if (role != null && role.getRoleMembers() != null && !role.getRoleMembers().isEmpty()) {
+                Iterator<RoleMember> roleit = role.getRoleMembers().iterator();
+                RoleMember rm;
+                while (roleit.hasNext()) {
+                    rm = roleit.next();
+                    if (rm != null && rm.getActive() != null && rm.getActive() == Boolean.FALSE) {
```
havetisyan: we don't really need rm.getActive() != null part since we're checking for m.getActive() == Boolean.FALSE

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -90,10 +90,10 @@
             + "WHERE role.role_id=? AND principal.name=? AND role_member.active=true;";
     private static final String SQL_GET_ROLE_MEMBER_EXISTS = "SELECT principal_id FROM role_member WHERE role_id=? AND principal_id=?;";
     private static final String SQL_GET_ROLE_MEMBER_INACTIVE_EXISTS = "SELECT principal_id FROM role_member WHERE role_id=? AND principal_id=? AND active=false;";
-    private static final String SQL_LIST_ROLE_MEMBERS = "SELECT principal.name, role_member.expiration FROM principal "
+    private static final String SQL_LIST_ROLE_MEMBERS = "SELECT principal.name, role_member.expiration, role_member.active FROM principal "
```
havetisyan: why do we need to make this a conditional? don't we already have active set to 1 by default? so isn't it as simple as:

WHERE role.role_id=? AND role_member.active=?

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -1288,7 +1290,15 @@ public int countRoles(String domainName) {
             throw notFoundError(caller, ZMSConsts.OBJECT_ROLE, ZMSUtils.roleResourceName(domainName, roleName));
         }
         List<RoleMember> members = new ArrayList<>();
-        try (PreparedStatement ps = con.prepareStatement(SQL_LIST_ROLE_MEMBERS)) {
+
+        String query = SQL_LIST_ROLE_MEMBERS;
```
havetisyan: see comment about the query string.

PR #771, opened 2019-09-04
Remove redundant log for filter to make it easy to debug

MartinTrojans:
Remove redundant log, because this debug info will come out for every request and it is not necessary. It will make it easy to debug while turn debug mode on and use this NoOpRateLimit filter


havetisyan:
👍

PR #777, opened 2019-09-24
Notification API

abvaidya:


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1838,7 +1838,7 @@ DomainRoleMembers listDomainRoleMembers(String domainName) {
         }
     }
 
-    Role getRole(String domainName, String roleName, Boolean auditLog, Boolean expand, Boolean pending) {
+    public Role getRole(String domainName, String roleName, Boolean auditLog, Boolean expand, Boolean pending) {
```
havetisyan: any reason we made this and the getAthenzDomain method public?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSConsts.java
```diff
@@ -229,5 +231,15 @@
     public static final int STRING_BLDR_SIZE_DEFAULT = 512;
 
     public static final String ZMS_JSON_PARSER_ERROR_RESPONSE = "{\"code\":400,\"message\":\"Invalid Object: checkout https://github.com/yahoo/athenz/core/zms for object defintions\"}";
+
+    public static final String NOTIFICATION_TYPE_MEMBERSHIP_APPROVAL = "MEMBERSHIP_APPROVAL";
```
havetisyan: lines 235 - 240: let's move these as public strings in the Notification interface class. This way the implementation can just reference and use them directly instead of having their own values and possibly misspelling, etc.

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/NotificationManager.java
```diff
@@ -0,0 +1,162 @@
...
+
+    NotificationServiceFactory notificationServiceFactory;
+    NotificationService notificationService;
+    ScheduledExecutorService scheduledExecutor;
+    private DBService dbService;
+
+    public NotificationManager(DBService dbService) {
+
+        this.dbService = dbService;
+        String notificationServiceFactoryClass = System.getProperty(ZMSConsts.ZMS_PROP_NOTIFICATION_SERVICE_FACTORY_CLASS);
```
havetisyan: 1) make this final

2) notification manager is not a required attribute thus notificationServiceFactoryClass can be null. So therefore we need to handle this correctly where we're running without any notification support.

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/NotificationManager.java
```diff
@@ -0,0 +1,162 @@
...
+
+    void shutdown() {
+        if (scheduledExecutor != null) {
+            scheduledExecutor.shutdownNow();
+        }
+    }
+
+    public void generateAndSendPostPutMembershipNotification(String domain, String org, Boolean auditEnabled, Boolean selfserve, Map<String, String> details) {
+
+        Set<String> recipients = new HashSet<>();
```
havetisyan: it's possible that we don't have any notification service configured. So before doing anything else we must check if the service is null or not and it's null, return without doing any work.

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/NotificationManager.java
```diff
@@ -0,0 +1,162 @@
...
+                recRole = recipient.substring(recipient.indexOf(":") + 6);
+                domain = dbService.getAthenzDomain(recDomain, false);
+                for (Role role : domain.getRoles()) {
+                    if (role.getName().equalsIgnoreCase(domain.getName() + ":role." + recRole)) {
+                        for (RoleMember member : role.getRoleMembers()) {
+                            notification.addRecipient(member.getMemberName());
+                        }
+                    }
+                }
+            } else if (recipient.startsWith("user.")) {
```
havetisyan: do not hardcode user. since the service might be configured with a different user domain. use the variables defined for home domain

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/NotificationManager.java
```diff
@@ -0,0 +1,162 @@
...
+        Notification notification = new Notification(notificationType);
+        notification.setDetails(details);
+        for (String recipient : recipients) {
+            if (recipient.contains(":role.")) {
+                //recipient is of type role. Extract role members
+                recDomain = recipient.substring(0, recipient.indexOf(":"));
+                recRole = recipient.substring(recipient.indexOf(":") + 6);
+                domain = dbService.getAthenzDomain(recDomain, false);
+                for (Role role : domain.getRoles()) {
+                    if (role.getName().equalsIgnoreCase(domain.getName() + ":role." + recRole)) {
```
havetisyan: isn't domain.getName() + ":role." + recRole is the same as recipient?

and if that's the case why are we extracting the recRole value?

any reason for equalsIgnoreCase since internally we already lower cased everything?

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/NotificationManager.java
```diff
@@ -0,0 +1,162 @@
...
+            }
+        }
+        if (notification.getRecipients() == null || notification.getRecipients().isEmpty()) {
+            throw new ResourceException(400, "Notification requires atleast 1 recipient.");
+        }
+        return notification;
+    }
+
+    public void sendNotification(Notification notification) {
+        notificationService.notify(notification);
```
havetisyan: again we should check if we have a notification service first

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -1620,4 +1620,47 @@ public boolean confirmRoleMember(String domainName, String roleName, RoleMember
         }
         return domainRoleMembersMap;
     }
+
+    @Override
+    public Set<String> getPendingMembershipApproverRoles() {
+
+        String[] fnames = getDomainList();
+        Set<String> roleNames = new HashSet<>();
+        List<String> slist = new ArrayList<>(java.util.Arrays.asList(fnames));
```
havetisyan: Isn't java.util.Arrays.asList(fnames) already give you list? Why calling ArrayList with it?

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
abvaidya: Because this is getting called from within NotificationManager now.  It had package access before.

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/NotificationManager.java
```diff
@@ -0,0 +1,162 @@
...
+            if (recipient.contains(":role.")) {
+                //recipient is of type role. Extract role members
+                recDomain = recipient.substring(0, recipient.indexOf(":"));
+                recRole = recipient.substring(recipient.indexOf(":") + 6);
+                domain = dbService.getAthenzDomain(recDomain, false);
+                for (Role role : domain.getRoles()) {
+                    if (role.getName().equalsIgnoreCase(domain.getName() + ":role." + recRole)) {
+                        for (RoleMember member : role.getRoleMembers()) {
+                            notification.addRecipient(member.getMemberName());
+                        }
```
havetisyan: I'm also assuming that once we have a match we should stop iterating through the roles?

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/NotificationManager.java
```diff
@@ -0,0 +1,162 @@
...
+                    }
+                }
+            } else if (recipient.startsWith("user.")) {
+                notification.addRecipient(recipient);
+            }
+        }
+        if (notification.getRecipients() == null || notification.getRecipients().isEmpty()) {
+            throw new ResourceException(400, "Notification requires atleast 1 recipient.");
```
havetisyan: if we are throwing runtime exceptions, who is properly catching? just because notification failed, it should not fail the full zms operation. It might be that it's better just return null and then handle accordingly where notification was a null object.

libs/java/server_common/src/test/java/com/yahoo/athenz/common/server/notification/NotificationTest.java
```diff
@@ -0,0 +1,66 @@
+package com.yahoo.athenz.common.server.notification;
```
havetisyan: let's include the copyright notice

servers/zms/src/main/java/com/yahoo/athenz/zms/NotificationManager.java
```diff
@@ -0,0 +1,172 @@
...
+public class NotificationManager {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(NotificationManager.class);
+
+
+    private NotificationServiceFactory notificationServiceFactory;
+    private NotificationService notificationService;
+    private ScheduledExecutorService scheduledExecutor;
+    private final DBService dbService;
+    private String userDomain;
```
havetisyan: why are we keeping the userDomain since we don't seem to use it - we only use the userDomainPrefix.

and why are we not passing the userDomainPrefix as part of the constructor since ZMSImpl already has that value?

servers/zms/src/main/java/com/yahoo/athenz/zms/NotificationManager.java
```diff
@@ -0,0 +1,172 @@
...
+        notification.setDetails(details);
+        for (String recipient : recipients) {
+            if (recipient.contains(":role.")) {
+                //recipient is of type role. Extract role members
+                recDomain = recipient.substring(0, recipient.indexOf(":"));
+                domain = dbService.getAthenzDomain(recDomain, false);
+                for (Role role : domain.getRoles()) {
+                    if (role.getName().equals(recipient)) {
+                        for (RoleMember member : role.getRoleMembers()) {
+                            notification.addRecipient(member.getMemberName());
```
havetisyan: we should check to make sure the member is a user and not a service (like we do on line 127).

servers/zms/src/main/java/com/yahoo/athenz/zms/NotificationManager.java
```diff
@@ -0,0 +1,172 @@
...
+
+    Notification createNotification(String notificationType, Set<String> recipients, Map<String, String> details) {
+        String recDomain;
+        AthenzDomain domain;
+        Notification notification = new Notification(notificationType);
+        notification.setDetails(details);
+        for (String recipient : recipients) {
+            if (recipient.contains(":role.")) {
+                //recipient is of type role. Extract role members
+                recDomain = recipient.substring(0, recipient.indexOf(":"));
```
havetisyan: there is no reason for us to call both recipient.contains and recipient.indexOf - first one is a wasted operation. we should something like
`
idx = recipient.indexOf(":role.");
if (idx != -1) {
    recDomain = recipient.substring(0, idx);
...
`

servers/zms/src/main/java/com/yahoo/athenz/zms/NotificationManager.java
```diff
@@ -0,0 +1,172 @@
...
+                            .map(RoleMember::getMemberName).collect(Collectors.toSet()));
+                }
+                if (orgRole != null) {
+                    recipients.addAll(orgRole.getRoleMembers().stream()
+                            .map(RoleMember::getMemberName).collect(Collectors.toSet()));
+                }
+            } else if (selfserve == Boolean.TRUE) {
+                // get admin role from the request domain
+                Role adminRole = dbService.getRole(domain, "admin", false, true, false);
+                recipients.addAll(adminRole.getRoleMembers().stream()
```
havetisyan: so for all three cases - lines 91 95 and here we should check that we're notifying users only.

while 91 and 91 - we control so probably there won't be any services, with user domains, not sure what could be there so we should have the checks.

PR #785, opened 2019-10-07
Notification reminder implementation

abvaidya:


servers/zms/src/test/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnectionTest.java
```diff
@@ -7649,4 +7648,164 @@ public void testGetPendingMembershipApproverRolesSelfServeError() throws Excepti
...
+
+    @Test
+    public void testProcessExpiredPendingMembersError() throws Exception {
+        JDBCConnection jdbcConn = new JDBCConnection(mockConn, true);
+        Mockito.when(mockResultSet.next())
+                .thenReturn(true) // first pending member found
+                .thenReturn(true) // second pending member found
+                .thenThrow(new SQLException("sql error"));
+        try {
+            boolean result = jdbcConn.processExpiredPendingMembers(30, "sys.auth.monitor");
```
havetisyan: we should have fail() after the call if we're expecting the call to throw an exception.

servers/zms/src/test/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnectionTest.java
```diff
@@ -7649,4 +7648,164 @@ public void testGetPendingMembershipApproverRolesSelfServeError() throws Excepti
...
+                .thenReturn("user.user1") //first pending
+                .thenReturn("user.user2"); // second pending
+
+        Mockito.when(mockPrepStmt.executeUpdate())
+                .thenReturn(1) // first audit entry
+                .thenReturn(1) // second audit entry
+                .thenThrow(new SQLException("sql error"));
+
+        try {
+            boolean result = jdbcConn.processExpiredPendingMembers(40, "sys.auth.monitor");
```
havetisyan: same here - we should have fail(). we also should remove "boolean result =" part since it's not used.

servers/zms/src/test/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnectionTest.java
```diff
@@ -7649,4 +7648,164 @@ public void testGetPendingMembershipApproverRolesSelfServeError() throws Excepti
...
+                .thenReturn("user.user1") //first pending
+                .thenReturn("user.user2"); // second pending
+
+        Mockito.when(mockPrepStmt.executeUpdate())
+                .thenReturn(1) // first audit entry
+                .thenReturn(1) // second audit entry
+                .thenReturn(0); // delete members
+
+        try {
+            boolean result = jdbcConn.processExpiredPendingMembers(40, "sys.auth.monitor");
```
havetisyan: same comment as above re fail and result.

servers/zms/src/test/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnectionTest.java
```diff
@@ -7649,4 +7648,164 @@ public void testGetPendingMembershipApproverRolesSelfServeError() throws Excepti
...
+        jdbcConn.close();
+    }
+
+    @Test
+    public void testUpdateLastNotifiedTimestampError() throws Exception {
+        JDBCConnection jdbcConn = new JDBCConnection(mockConn, true);
+        Mockito.when(mockPrepStmt.executeUpdate())
+                .thenThrow(new SQLException("sql error")); // 3 members updated
+        try {
+            boolean result = jdbcConn.updateLastNotifiedTimestamp(30);
```
havetisyan: and here as well.

servers/zms/src/test/java/com/yahoo/athenz/zms/store/file/FileConnectionTest.java
```diff
@@ -603,4 +603,29 @@ public void testgetPendingMembershipApproverRoles() {
         }
     }
 
+    @Test
+    public void testProcessExpiredPendingMembers() {
+        File fileDir = new File("/home/athenz/zms_store");
+        File quotaDir = new File("/home/athenz/zms_quota");
+        try (FileConnection fileconnection = new FileConnection(fileDir, quotaDir)) {
+            try {
+                fileconnection.processExpiredPendingMembers(30, "sys.auth.monitor");
```
havetisyan: we should have fail() after the call if we're expecting the call to throw an exception.

servers/zms/src/test/java/com/yahoo/athenz/zms/store/file/FileConnectionTest.java
```diff
@@ -603,4 +603,29 @@ public void testgetPendingMembershipApproverRoles() {
...
+        }
+    }
+
+    @Test
+    public void testUpdateLastNotifiedTimestamp() {
+        File fileDir = new File("/home/athenz/zms_store");
+        File quotaDir = new File("/home/athenz/zms_quota");
+        try (FileConnection fileconnection = new FileConnection(fileDir, quotaDir)) {
+            try {
+                fileconnection.updateLastNotifiedTimestamp(45);
```
havetisyan: same as above

servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
```diff
@@ -4311,4 +4314,22 @@ public void testGetPendingMembershipNotificationsEdge() {
...
         zms.dbService.store = saveStore;
     }
+
+    @Test
+    public void testProcessExpiredPendingMembers() {
+        Mockito.when(mockObjStore.getConnection(true, true)).thenReturn(mockFileConn);
+        Mockito.when(mockFileConn.processExpiredPendingMembers(30, "sys.auth.monitor")).thenReturn(true);
+
+        zms.dbService.processExpiredPendingMembers(30, "sys.auth.monitor");
+        assertTrue(true);
```
havetisyan: don't need this here since in case of exception the test case will fail anyway.

servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
```diff
@@ -4311,4 +4314,22 @@ public void testGetPendingMembershipNotificationsEdge() {
...
+    }
+
+    @Test
+    public void testUpdateLastNotifiedTimestamp() {
+        Mockito.when(mockObjStore.getConnection(true, true)).thenReturn(mockFileConn);
+        Mockito.when(mockFileConn.updateLastNotifiedTimestamp(30)).thenReturn(true);
+
+        zms.dbService.updateLastNotifiedTimestamp(30);
+        assertTrue(true);
```
havetisyan: same as above.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3701,6 +3709,48 @@ private void populateDomainRoleMembersMapFromResultSet(Map<String, List<DomainRo
         return targetRoles;
     }
 
+    @Override
+    public boolean processExpiredPendingMembers(int pendingRoleMemberLifespan, String monitorIdentity) {
```
havetisyan: The issue with this implementation is that we're not logging these expiry in our official audit log file. We keep track of them in the per-role but those are not permanent - e.g. if I delete the role, the history is gone as well.

So we need to make sure to log these in the audit log file. So the implementation should just fetch the expired pending members from DBService class and then go through each one and execute an expired operation - having the system monitor should be sufficient to identify that this was an expiry operation.

servers/zms/src/main/java/com/yahoo/athenz/zms/NotificationManager.java
```diff
@@ -175,8 +180,9 @@ public void run() {
         }
 
         private void sendPendingMembershipApprovalReminders() {
+            dbService.updateLastNotifiedTimestamp(pendingRoleMemberLifespan);
```
havetisyan: Not sure if I follow how we're protecting multiple hosts processing the same data multiple times. 

host1 - updates all the timestamps.
host2 - tries the same but nothing is updated.
host1 - gets all pending roles
host2 - gets the same pending roles
and then both process notifications.

we might want to consider introducing another column that maintains what host is updating the entry thus that host will get those during get call to process.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3696,14 +3707,69 @@ private void populateDomainRoleMembersMapFromResultSet(Map<String, List<DomainRo
...
+            } catch (SQLException ex) {
+                throw sqlError(ex, caller);
+            }
+        }
+        return membersMapList;
+    }
+
+    @Override
+    public boolean updateLastNotifiedTimestamp(String server, java.sql.Timestamp timestamp) {
+        String caller = "updateLastNotifiedTimestamp";
```
havetisyan: should be final

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3696,14 +3707,69 @@ private void populateDomainRoleMembersMapFromResultSet(Map<String, List<DomainRo
...
 
         return targetRoles;
     }
 
-    private void getRecipientRoleForSelfServeMembershipApproval(String caller, Set<String> targetRoles) {
+    @Override
+    public List<Map<String, String>> processExpiredPendingMembers(int pendingRoleMemberLifespan, String monitorIdentity) {
+
+        boolean result = true;
+        String caller = "processExpiredPendingMembers";
```
havetisyan: should be final

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3696,14 +3707,69 @@ private void populateDomainRoleMembersMapFromResultSet(Map<String, List<DomainRo
...
+            ps.setInt(1, pendingRoleMemberLifespan);
+            try (ResultSet rs = executeQuery(ps, caller)) {
+                while (rs.next()) {
+                    membersMap = new HashMap<>();
+                    membersMap.put("domain", rs.getString(4));
+                    membersMap.put("role", rs.getString(3));
+                    member = rs.getString(2);
+                    membersMap.put("member", member);
+                    membersMapList.add(membersMap);
+                    result = result && insertRoleAuditLog(rs.getInt(1), monitorIdentity, member, "REJECT", "Expired");
```
havetisyan: not sure if I follow this logic. we're processing all the entries however, if any one of the entries fails, we mark the result as failed and don't execute the delete command. However, we have already inserted all those log entries claiming that the requests were rejected due to being expired. so the correct options would be:
a) either wrap this in a transaction where in case of any errors from insertRoleAuditLog we return right away and in DBService property rollback the changes (the problem with this method is that would we ever get into a case where we can never cleanup because one entry keeps failing?).
b) remove the logic to remove all entries at once and instead return the list to DBService and there we can have one delete operation (just like delete member) per entry wrapped in transaction.

PR #795, opened 2019-10-24
updated dependencies, 1 bc api change

havetisyan:


mnagaya:
bouncycastle-1.63 or later is supported multi release JAR files.

https://www.bouncycastle.org/latest_releases.html

So some apps may break compatibility.

https://github.com/eclipse/jetty.project/issues/1797

This is just FYI, I will update jetty to latest.

PR #797, opened 2019-10-28
Triple brace = no escape

JesseClarkND:
Handlebars HTML-escapes values returned by a {{expression}}. If you don't want Handlebars to escape a value, use the "triple-stash", {{{.

The use of {{{ for  athenzParams = {{{json allParams}}} was leading to an xss on almost every page.

### DIFF
--- a/ui/views/layouts/main.handlebars
+++ b/ui/views/layouts/main.handlebars
@@ -20,7 +20,7 @@
           </div>
         {{/if}}
         <script type="text/javascript">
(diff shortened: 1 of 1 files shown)


havetisyan:
This change breaks the display of all users in the roles. Since the UI is presented as a reference implementation where during a development phase it can be used to add roles, see members, etc, this will break that functionality thus will not be accepted. the expectation is that the property will integrate with their own authentication system, verify and fix any csrf/xss issues before deploying this in their production.

PR #800, opened 2019-10-29
fixing xss

plmaltais:
Fix for https://hackerone.com/reports/724621


havetisyan:
a side effect of this fix is that the the admin names are now showing up with their http links as inline text. #821 addresses this issue with removing the http links and only showing the admin names.

PR #807, opened 2019-11-02
Add authorization check for update role membership

c0rvax:


havetisyan:
Thanks for the PR. However both of the changes are invalid and, in fact, will cause, auditing and governance functionality present in the server to break. 

First, ZMSResources.java is auto-generated and so no manual changes are allowed (see the comments on top of the file). If you were to make such a change, then you would update the RDL: https://github.com/yahoo/athenz/blob/master/core/zms/src/main/rdl/Role.rdli#L129 and then ZMSResources class will be automatically updated.

One limitation with RDL (this is another open source project from Yahoo: https://github.com/ardielle) it only supports having a single authorization check within an API call. in the case of role membership the logic is different:

a) if the role is marked for auditing/sox compliance then the domain admins can't add members. instead they can only add pending memory and designated users can approve those requests.
b) if the role is marked as self-serve then the users themselves can add them into the role but they'll be added in pending state and one of the domain admins must approve the request.
c) if it's a standard role, then the domain admins are authorized to add members.

So all this logic is in the isAllowedPutMembership method: https://github.com/yahoo/athenz/blob/master/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java#L3112

Similar logic exists for putMembershipDecision api as well.

PR #849, opened 2019-12-16
moving from simple mail to MIME

abvaidya:
bug fix to avoid adding duplicate rows for pending members


libs/java/server_common/src/main/resources/messages/membership-approval.html
```diff
@@ -0,0 +1,31 @@
...
+        <hr>
+        <table id="t01">
+            <tr><td class="ch">DOMAIN</td><td class="cv">{0}</td></tr>
+            <tr><td class="ch">ROLE</td><td class="cv">{1}</td></tr>
+            <tr><td class="ch">MEMBER</td><td class="cv">{2}</td></tr>
+            <tr><td class="ch">REASON</td><td class="cv">{3}</td></tr>
+            <tr><td class="ch">REQUESTOR</td><td class="cv">{4}</td></tr>
+        </table>
+        <hr>
+        <div class="bt">Please visit the <a href="{5}">Athenz Workflow Sceen</a> to act on the request.</div>
```
havetisyan: Screen

libs/java/server_common/src/main/resources/messages/membership-approval.html
```diff
@@ -0,0 +1,31 @@
...
+<head>
+    <meta charset="utf-8">
+    <title></title>
+    <style></style>
+</head>
+<body style="margin: 0; background: rgba(231, 232, 231, 1.0);">
+<div class="wrapper">
+    <div class="mbrapproval">
+        <div class="logo">
+            <!-- <img src="athenz-logo-white.png" class="athenzlogowhite" alt="Athenz logo"/> -->
```
havetisyan: why a comment? can we remove this?

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
```diff
@@ -134,7 +148,10 @@ String getBody(String type, Map<String, String> details) {
                 body = getPrincipalExpiryBody(details);
                 break;
         }
-        body = body + getFooter();
+        if (addFooter) {
+            body = body + getFooter();
+        }
+        body = body.replace("<style></style>", "<style>" + readContentFromFile(EMAIL_TEMPLATE_CSS) + "</style>");
```
havetisyan: here and also on line 193, we seem to be calling readContentsFromFile every time we generate a message. Can't we read those contents only once during class constructor and then keep reusing them?

libs/java/server_common/src/main/resources/messages/membership-approval.html
abvaidya: Will fix the typo. 
For now i did the changes only for membership notification. I can use same template for expiry notifications as well.

PR #855, opened 2019-12-23
add extract domain/service/role name functions to AthenzUtils

havetisyan:

### DIFF
--- a/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/AthenzUtils.java
+++ b/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/AthenzUtils.java
@@ -112,7 +112,7 @@ public class AthenzUtils {
      * which includes the role name. The format of the role name
      * is {domain}:role.{role-name}
      * @param roleName the full arn of the role
-     * @return role name, null if it's not expected full arn format
+     * @return domain name, null if it's not expected full arn format
      */
     public static String extractRoleDomainName(final String roleName) {
         int idx = roleName.indexOf(ROLE_SEP);
@@ -127,7 +127,7 @@ public class AthenzUtils {
      * Extract the domain name from the principal name which
      * is in the format {domain}.{service}
      * @param principalName full name of the principal
-     * @return domain name, null if it's not the expected full service name
+     * @return domain name, null if it's not the expected full principal service name
      */
     public static String extractPrincipalDomainName(final String principalName) {
         int idx = principalName.lastIndexOf('.');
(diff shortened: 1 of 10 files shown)


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/AthenzUtils.java
```diff
@@ -90,4 +90,66 @@ public static boolean isRoleCertificate(X509Certificate x509Cert) {
...
+        } else {
+            return roleName.substring(0, idx);
+        }
+    }
+
+    /**
+     * Extract the domain name from the principal name which
+     * is in the format {domain}.{service}
+     * @param principalName full name of the principal
+     * @return domain name, null if it's not the expected full service name
```
abvaidya: if it's not the expected full domain name

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/AthenzUtils.java
```diff
@@ -90,4 +90,66 @@ public static boolean isRoleCertificate(X509Certificate x509Cert) {
...
+            return roleName.substring(idx + ROLE_SEP.length());
+        }
+    }
+
+    /**
+     * Extract the domain name from the full Athenz Role Name (arn)
+     * which includes the role name. The format of the role name
+     * is {domain}:role.{role-name}
+     * @param roleName the full arn of the role
+     * @return role name, null if it's not expected full arn format
```
abvaidya: @return domain name

havetisyan: will fix it now

servers/zts/src/main/java/com/yahoo/athenz/zts/cache/DataCache.java
```diff
@@ -17,9 +17,9 @@
 
 import java.util.*;
```
pratikgote: can we expand the * ?

servers/zts/src/main/java/com/yahoo/athenz/zts/cache/DataCache.java
```diff
@@ -17,9 +17,9 @@
...
 
+import com.yahoo.athenz.auth.util.AthenzUtils;
 import com.yahoo.athenz.zms.*;
```
pratikgote: can we expand the * ?

havetisyan:
@pratikgote IntelliJ automatically switches to * when it determines that we're doing too many imports. So I've been ok with it as I would expect the default behavior is most likely what's correct for most projects and no need to keep overriding it.

PR #856, opened 2019-12-24
updated dependency packages

havetisyan:

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -69,22 +69,22 @@
   </ciManagement>
 
   <properties>
-    <jetty.version>9.4.24.v20191120</jetty.version>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -222,7 +222,7 @@
     <dependency>
       <groupId>org.testng</groupId>
       <artifactId>testng</artifactId>
-      <version>7.0.0</version>
+      <version>7.1.1</version>
```
abvaidya: should be 7.1.0

havetisyan: 7.1.1 is the latest - it’s just hasn’t been pushed yet from bintray. Once it’s synced then we can murge this pr.

PR #861, opened 2020-01-04
properly reading resources from classpath

abvaidya:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
```diff
@@ -117,14 +118,30 @@
...
     }
 
+    byte[] readBinaryFromFile(String fileName) {
+
+        byte[] fileByteArray = null;
+        URL resource = getClass().getClassLoader().getResource(fileName);
+        if (resource != null) {
+            try (InputStream fileStream = getClass().getClassLoader().getResourceAsStream(fileName)) {
+                //convert to byte array
+                fileByteArray = IOUtils.toByteArray(Objects.requireNonNull(fileStream));
```
havetisyan: Both here and in the readContentFromFile method why are we calling Objects.requireNonNull if we have already verified that the resource is valid - there is resource != null check? was the plan to remove the initial resource check completely?

PR #877, opened 2020-02-10
system admin option to reset domain last modified timestamp

havetisyan:
if we need to make sure a domain is synced from zts to zms, this option would allow system admins to change the last modified timestamp thus marking the domain as changed rather than asking the domain admin to make some change that is not necessary in the domain.


psasidhar:
👍

PR #878, opened 2020-02-10
If user is expired, prevent sending notifications

OferLevi85:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/NotificationManager.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/NotificationManager.java
@@ -145,11 +145,19 @@ public class NotificationManager {
         for (Role role : domain.getRoles()) {
             if (role.getName().equals(roleName)) {
                 notification.getRecipients().addAll(role.getRoleMembers().stream()
-                        .filter(m -> m.getMemberName().startsWith(userDomainPrefix))
+                        .filter(this::isUnexpiredUser)
                         .map(RoleMember::getMemberName).collect(Collectors.toSet()));
                 return;
             }
         }
+    }
+
+    private boolean isUnexpiredUser(RoleMember roleMember) {
+        if (!roleMember.getMemberName().startsWith(userDomainPrefix)) {
+            return false;
+        }
+
+        return (roleMember.getExpiration() == null) || (roleMember.getExpiration().millis() > System.currentTimeMillis());
     }
 
     void addNotificationRecipient(Notification notification, final String recipient, boolean ignoreService) {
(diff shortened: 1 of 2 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/NotificationManager.java
```diff
@@ -145,13 +145,19 @@ void addDomainRoleRecipients(Notification notification, final String domainName,
...
                 notification.getRecipients().addAll(role.getRoleMembers().stream()
-                        .filter(m -> m.getMemberName().startsWith(userDomainPrefix))
+                        .filter(this::isUnexpiredUser)
                         .map(RoleMember::getMemberName).collect(Collectors.toSet()));
                 return;
             }
         }
     }
 
+    private boolean isUnexpiredUser(RoleMember roleMember) {
```
havetisyan: while the logic is correct, I don't want us to carry out additional work if not necessary. e.g. if the principal is not user there is no reason for us to check the expiry (it's just a wasted operation).

we should have logic implemented like this:

if (!roleMember.getMemberName().startsWith(userDomainPrefix)) {
    return false;
}
return (roleMember.getExpiration() == null) || (roleMember.getExpiration().millis() > System.currentTimeMillis());

OferLevi85: Thanks for the review I'll fix it shortly.
I do want to provide an explanation as to why I did it this way (as I considered the option of returning early if possible). I thought that the extra operation is acceptable as the first check that compares characters is already quite expensive and it won't make a noticeable difference. Readability seemed like the bigger concern so I tried to minimize the number of return statements in the function (as mentioned in "Code Complete"). If this extra operation is expensive though, instead of calling System.currentTimeMiils() in each iteration, I can get the current time as a parameter so it will only be calculated once (this will also have some effect on readability). What do you think?

havetisyan: The problem is not with the System.currentTimeMiils but rather with the Timestamp object from rdl. Take a look at that class's millis implementation (if it hasn't been done before, it actually parses the full string to get a date object and then returns the getTime method - not that efficient so we should try to avoid that as much as possible. if you want to minimize the return statements then you can just combine the values with && in the return - the jdk should optimize the call and not call the second part of the && if the first is already false - something like:

boolean isUser = roleMember.getMemberName().startsWith(userDomainPrefix);
return (isUser && ((roleMember.getExpiration() == null) || (roleMember.getExpiration().millis() > System.currentTimeMillis())));

OferLevi85: I wasn't aware of the work done in the rdl to get normalizedMillis. Thanks for pointing it out! Regarding the return statements, I think that it is best to leave it as you first suggested as the full boolean statement with the added parentheses is harder to read then viewing the 2 statements separately.

PR #881, opened 2020-02-17
Notifications for unrefreshed certificates - first stage. Seperating NotificationManager from providers.

OferLevi85:
1. Seperated provider implementation from EmailNotificationService
2. Currently implemented provider: AWSEmailProvider. SonicEmailProvider class was created but not implemented.
3. Renamed NotificationServiceFactoryImpl to EmailNotificationServiceFactory and its create function now gets a “providerName” argument.
4. The NotificationManager will get the providerName argument by reading a system property (“athenz.notification.email_provider"). Currently supported values: “AWS” and “Sonic”
5. The tests checking AWS SES functionality in NotificationManagerTest were extracted to a new test class, AWSEmailProviderTest.
6. Moved Notification constants from interface to dedicated constants file


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/EmailProvider.java
```diff
@@ -0,0 +1,10 @@
+package com.yahoo.athenz.common.server.notification;
+
+import java.util.Collection;
+
+public interface EmailProvider {
+    String HTML_LOGO_CID_PLACEHOLDER = "<logo>";
```
dvirguttman: Better use constants in classes.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationServiceFactory.java
```diff
@@ -0,0 +1,54 @@
...
+                break;
+            case "Sonic":
+                provider = new SonicEmailProvider();
+                break;
+            default:
+                LOGGER.warn("Unknown Email Provider: " + providerName + ". Failed to instantiate NotificationService");
+                return null;
+        }
+
+        return new EmailNotificationService(provider);
```
dvirguttman: Just return the provider in the switch case itself

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/EmailProvider.java
OferLevi85: You're right. Fixed it.

PR #884, opened 2020-02-20
Notifications for unrefreshed certificates - Separating EmailNotificationService from providers

OferLevi85:
1. Seperated provider implementation from EmailNotificationService
2. Currently implemented provider: AWSEmailProvider..
3. The tests checking AWS SES functionality in NotificationManagerTest were extracted to a new test class, AWSEmailProviderTest.
4. Moved Notification constants from interface to dedicated constants file

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/AWSEmailProvider.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/AWSEmailProvider.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright 2020 Verizon Media
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.yahoo.athenz.common.server.notification.impl;
+
+import com.amazonaws.regions.Region;
+import com.amazonaws.regions.Regions;
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailService;
+import com.amazonaws.services.simpleemail.AmazonSimpleEmailServiceClientBuilder;
+import com.amazonaws.services.simpleemail.model.RawMessage;
+import com.amazonaws.services.simpleemail.model.SendRawEmailRequest;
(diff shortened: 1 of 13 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/NotificationServiceFactoryImpl.java
```diff
@@ -26,6 +26,6 @@
 
     @Override
     public NotificationService create() {
-        return new EmailNotificationService();
+        return new EmailNotificationService(new AWSEmailProvider());
```
abvaidya: Instead of hardcoding the AWSEmailProvider here, it should come from properties, so that it can be replaced with another EmailProvider for ZMS vs ZTS

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/AWSEmailProvider.java
```diff
@@ -0,0 +1,116 @@
...
+import static com.yahoo.athenz.common.server.notification.NotificationServiceConstants.HTML_LOGO_CID_PLACEHOLDER;
+
+public class AWSEmailProvider implements EmailProvider {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AWSEmailProvider.class);
+    private final AmazonSimpleEmailService ses;
+
+    @Override
+    public boolean sendEmail(String subject, String body, boolean status, Collection<String> recipients, String from, byte[] logoImage) {
+        try {
+            Session session = Session.getDefaultInstance(new Properties());
```
abvaidya: Since creating MIMEMessage is common across diff email providers, it can be extracted in the EmailProvider implementation as default method

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/NotificationServiceFactoryImpl.java
OferLevi85: Yes, I've considered it and it is actually part of the original design. But that means there would be 2 separate factory types. One for NotificationService - currently only email but in the future could be something else like push messages to phone - and another for providers such as AWS SES, or some SMS messages provider.
The providers and types of Notification Services are coupled as there is no reason to Instantiate a NotificationService of one type (for example, EmailNotificationService) with a provider from another type (for example, some push notifications Provider).
So I've decided on using a single factory. Implementors can implement other factories that create EmailNotificationService with a different provider of their own or create a different type of NotificationService all together with its own providers.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/AWSEmailProvider.java
OferLevi85: Yes I'm planning on doing it in the next couple of pull requests. I just want to implement one more email provider before doing it so I'll know exactly what I can extract from this method and what I should keep.

PR #889, opened 2020-03-01
Change zts schema to enable notifications

OferLevi85:


servers/zts/schema/zts_server.sql
```diff
@@ -31,6 +31,10 @@ CREATE TABLE IF NOT EXISTS `zts_store`.`certificates` (
   `prevTime` DATETIME(3) NOT NULL,
   `prevIP` VARCHAR(64) NOT NULL,
   `clientCert` TINYINT(1) NOT NULL DEFAULT 0,
+  `lastNotifiedTime` DATETIME(3) NULL,
```
OferLevi85: - lastNotifiedTime and lastNotifiedServer are nullable (notifications might have never been sent)
- expiryTime is nullable instead of non-nullable with a default value (as a default value might be accidentally presented to users)
- hostName is nullable to support rows without this value (created by requests that didn't send the hostName). We can consider replacing it with some default value (and making the column non-nullable).

havetisyan:
@OferLevi85 as part of the schema change, we must also include the change required for any property that is already running a live server. Please include an update SQL commands as part of this PR. Take a look at the other updates in the following directory: https://github.com/yahoo/athenz/tree/master/servers/zts/schema/updates

PR #892, opened 2020-03-04
Adding validation for SSH Principals

psasidhar:


servers/zts/src/main/java/com/yahoo/athenz/zts/cert/InstanceCertManager.java
```diff
@@ -726,6 +734,67 @@ public boolean verifyInstanceCertIPAddress(final String provider, final String i
...
+
+
+    /**
+     * validates hostname against the resolver, and verifies that the ssh principals map to hostname
+     * @param hostname of the instance
+     * @param csr ssh host csr from the sia
+     * @return boolean true or false
+     */
+    public boolean validPrincipals(final String hostname, final String csr) {
+        LOGGER.error("input: {}", csr);
```
havetisyan: we should remove this or make it debug

PR #898, opened 2020-03-06
fixed timeout config for zms java client

smrutilal2:

### DIFF
--- a/clients/java/zms/pom.xml
+++ b/clients/java/zms/pom.xml
@@ -128,22 +128,11 @@
     </dependency>
     <dependency>
       <groupId>org.jboss.resteasy</groupId>
-      <artifactId>resteasy-jaxrs-all</artifactId>
-      <version>4.5.2.Final</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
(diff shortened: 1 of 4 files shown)


smrutilal2:
Fixes: https://github.com/yahoo/athenz/issues/897.
This might fix https://github.com/yahoo/athenz/issues/894 also. Need to be verified.

clients/java/zms/pom.xml
```diff
@@ -113,6 +113,37 @@
...
+    <dependency>
+      <groupId>io.undertow</groupId>
+      <artifactId>undertow-servlet</artifactId>
+      <version>2.0.29.Final</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-jaxrs-all</artifactId>
+      <version>4.5.2.Final</version>
```
havetisyan: @smrutilal2  need to add   <type>pom</type> otherwise build looks for jar type and can't find it

smrutilal2: Ohh I see. Thank you!

PR #904, opened 2020-03-09
recording last_review_date in DB when role is reviewed

abvaidya:


havetisyan:
@abvaidya not sure if I like the fact that we're creating yet another api in the storage interface to set the both timestamps in one call. we have 3 api to set 2 values. In theory we can just have 1 that takes something like EnumSet but not to make backward incompatible changes, we can just stick with calling the two separate apis. The number of review role operations is not going that many that we're concerned about performance and combining the commands into 1.

PR #905, opened 2020-03-09
Fix zts timeout

smrutilal2:


clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -380,9 +380,9 @@ private void initClient(String url, SSLContext sslContext) {
 
         Client rsClient =
             builder
+                .withConfig(clientConfig)
```
smrutilal2: `clientConfig` needs to be added before the timeouts. Otherwise `JerseyClientBuilder`.withConfig()` would replace all the existing settings with `clientConfig`.

Issue #894, opened 2020-03-05
Timeout properties don't work at the latest version of athenz-zms-java-client 


TakuyaMatsu:
Hello.
`ZMS_CLIENT_PROP_READ_TIMEOUT` and `ZMS_CLIENT_PROP_CONNECT_TIMEOUT` don't seem to work at latest athenz-zms-java-client.

# Problem?
I think timeout properties are overwritten by `withConfig(clientConfig)`.
https://github.com/yahoo/athenz/blob/9b2f1202663455322037390be2feb02d7ca0b854/clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java#L381-L384

so old version is working 
https://github.com/yahoo/athenz/blob/d767dc28dcd07d1bfdd38126fb578a10fd9fead3/clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java#L348-L350

## Fix
Change to setting properties for ClientConfig.
```
         ClientConfig clientConfig = new ClientConfig(jacksonJsonProvider)
                 .property(ClientProperties.CONNECT_TIMEOUT, connectTimeout)
                 .property(ClientProperties.READ_TIMEOUT, readTimeout);

         Client rsClient = builder
                 .withConfig(clientConfig)
                 .build();
         client = new ZMSRDLGeneratedClient(zmsUrl, rsClient);
```
OR
delete `withConfig(clientConfig)`, like old version zms-client
```
        Client rsClient = builder.property(ClientProperties.CONNECT_TIMEOUT, connectTimeout)
            .property(ClientProperties.READ_TIMEOUT, readTimeout)
            .build();
```

# Test
Test code

```
package com.hoge.test;

import com.yahoo.athenz.zms.ZMSClient;

public class ZMSClientSample {
    public static final String ZMS_CLIENT_PROP_READ_TIMEOUT    = "athenz.zms.client.read_timeout";
    public static final String ZMS_CLIENT_PROP_CONNECT_TIMEOUT = "athenz.zms.client.connect_timeout";

    public static void main(String[] args) {
        String baseUrl = args[0];
        String domainName = args[1];
        String roleName = args[2];
        String readTimeout = args[3];
        String connTimeout = args[4];

        System.setProperty(ZMS_CLIENT_PROP_READ_TIMEOUT, readTimeout);
        System.setProperty(ZMS_CLIENT_PROP_CONNECT_TIMEOUT, connTimeout);

[...]

PR #906, opened 2020-03-10
fix withConfig timing at zms-client

TakuyaMatsu:
#894


havetisyan:
Thanks @TakuyaMatsu  but that fix is already included as part of the #905 PR

TakuyaMatsu:
@havetisyan 
Thank you for your reply.
I missed it !! I close this.

PR #907, opened 2020-03-10
Implement ZTS Notifications DB logic

OferLevi85:
- Created CertFailedRefreshNotificationTask but not yet implemented. Will be implemented after DB logic is approved.
- CertRecordStoreConnection - Added 2 methods to fetch unrefreshed certificates that haven't been notified by other servers.
- DomainRoleMembersFetcherCommon - Added common implementation for ZMS entities (and added a dependency on athenz-zms-core in common)
- JDBCCertRecordStoreConnection - Modified Insert / Update methods for new notification columns.

### DIFF
--- a/libs/java/server_common/pom.xml
+++ b/libs/java/server_common/pom.xml
@@ -163,6 +163,11 @@
       <artifactId>jakarta.mail</artifactId>
       <version>1.6.4</version>
     </dependency>
+    <dependency>
+      <groupId>com.yahoo.athenz</groupId>
+      <artifactId>athenz-zms-core</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
   </dependencies>
 
   <distributionManagement>
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/ServerCommonConsts.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/ServerCommonConsts.java
@@ -21,6 +21,8 @@ public final class ServerCommonConsts {
     public static final String OBJECT_ROLE      = "role";
     public static final String USER_DOMAIN      = "user";
 
+    public static final String USER_DOMAIN_PREFIX = "user.";
+
     private ServerCommonConsts() {
     }
 }
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
@@ -251,20 +251,21 @@ public class EmailNotificationService implements NotificationService {
 
     boolean sendEmail(Set<String> recipients, String subject, String body) {
         final AtomicInteger counter = new AtomicInteger();
-        boolean status = true;
         // SES imposes a limit of 50 recipients. So we convert the recipients into batches
         if (recipients.size() > SES_RECIPIENTS_LIMIT_PER_MESSAGE) {
             final Collection<List<String>> recipientsBatch = recipients.stream()
                     .collect(Collectors.groupingBy(it -> counter.getAndIncrement() / SES_RECIPIENTS_LIMIT_PER_MESSAGE))
                     .values();
+            boolean status = true;
             for (List<String> recipientsSegment : recipientsBatch) {
-                status = sendEmailMIME(subject, body, status, recipientsSegment);
+                if (!sendEmailMIME(subject, body, recipientsSegment)) {
+                    status = false;
+                }
             }
+            return status;
         } else {
-            status = sendEmailMIME(subject, body, status, new ArrayList<>(recipients));
+            return sendEmailMIME(subject, body, new ArrayList<>(recipients));
         }
-
-        return status;
     }
 
     String readContentFromFile(String fileName) {
@@ -337,7 +338,7 @@ public class EmailNotificationService implements NotificationService {
         return message;
     }
 
-    private boolean sendEmailMIME(String subject, String body, boolean status, Collection<String> recipients) {
+    private boolean sendEmailMIME(String subject, String body, Collection<String> recipients) {
         MimeMessage mimeMessage;
         try {
             mimeMessage = getMimeMessage(subject, body, recipients, from, logoImage);
@@ -346,6 +347,6 @@ public class EmailNotificationService implements NotificationService {
             return false;
         }
 
-        return emailProvider.sendEmail(status, recipients, from + AT + emailDomainFrom, mimeMessage);
+        return emailProvider.sendEmail(recipients, from + AT + emailDomainFrom, mimeMessage);
     }
 }
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -34,8 +34,10 @@ import com.yahoo.athenz.auth.token.AccessToken;
 import com.yahoo.athenz.auth.token.IdToken;
 import com.yahoo.athenz.common.server.dns.HostnameResolver;
 import com.yahoo.athenz.common.server.dns.HostnameResolverFactory;
+import com.yahoo.athenz.common.server.notification.NotificationManager;
 import com.yahoo.athenz.zms.RoleMeta;
 import com.yahoo.athenz.zts.cert.*;
+import com.yahoo.athenz.zts.notification.ZTSNotificationTaskFactory;
 import org.bouncycastle.pkcs.PKCS10CertificationRequest;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -177,6 +179,7 @@ public class ZTSImpl implements KeyStore, ZTSHandler {
     protected Http.AuthorityList authorities = null;
     protected ZTSAuthorizer authorizer;
     protected static Validator validator;
+    protected NotificationManager notificationManager = null;
     
     enum AthenzObject {
         DOMAIN_METRICS {
@@ -305,8 +308,15 @@ public class ZTSImpl implements KeyStore, ZTSHandler {
         // make sure to set the keystore for any instance that requires it
         
         setAuthorityKeyStore();
+
+        setNotificationManager();
     }
-    
+
+    private void setNotificationManager() {
+        ZTSNotificationTaskFactory ztsNotificationTaskFactory = new ZTSNotificationTaskFactory(instanceCertManager, dataStore, userDomainPrefix, serverHostName);
+        notificationManager = new NotificationManager(ztsNotificationTaskFactory.getNotificationTasks());
+    }
+
     void loadSystemProperties() {
         String propFile = System.getProperty(ZTSConsts.ZTS_PROP_FILE_NAME,
                 getRootDir() + "/conf/zts_server/zts.properties");
@@ -2475,8 +2485,14 @@ public class ZTSImpl implements KeyStore, ZTSHandler {
         return false;
     }
 
-    X509CertRecord insertX509CertRecord(ResourceContext ctx, final String cn, final String provider,
-            final String instanceId, final String serial, Boolean certUsage) {
+    X509CertRecord insertX509CertRecord(ResourceContext ctx,
+                                        final String    cn,
+                                        final String    provider,
+                                        final String    instanceId,
+                                        final String    serial,
+                                        final Boolean   certUsage,
+                                        final Date      expirationDate,
+                                        final String    hostName) {
 
         X509CertRecord x509CertRecord = new X509CertRecord();
         x509CertRecord.setService(cn);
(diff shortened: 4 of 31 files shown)


libs/java/server_common/pom.xml
```diff
@@ -163,6 +163,12 @@
       <artifactId>jakarta.mail</artifactId>
       <version>1.6.4</version>
     </dependency>
+      <dependency>
```
OferLevi85: Added dependency so athenz-server-common can refer to entities.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/ServerCommonConsts.java
```diff
@@ -21,6 +21,8 @@
     public static final String OBJECT_ROLE      = "role";
     public static final String USER_DOMAIN      = "user";
 
+    public static final String USER_DOMAIN_PREFIX = "user.";
+
```
OferLevi85: Moved USER_DOMAIN_PREFIX to common so it can be used by ZTS.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -305,8 +308,15 @@ public ZTSImpl(CloudStore implCloudStore, DataStore implDataStore) {
         // make sure to set the keystore for any instance that requires it
         
         setAuthorityKeyStore();
+
+        setNotificationManager();
     }
-    
+
+    private void setNotificationManager() {
```
OferLevi85: Initiated NotificationManager in ZTS. The daily thread will run but at this point won't send any notifications as the current NotificationTask is not implemented (returns empty collection).

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2475,8 +2485,14 @@ boolean verifyAWSAssumeRole(String domainName, String roleResource, String princ
         return false;
     }
 
-    X509CertRecord insertX509CertRecord(ResourceContext ctx, final String cn, final String provider,
-            final String instanceId, final String serial, Boolean certUsage) {
+    X509CertRecord insertX509CertRecord(ResourceContext ctx,
```
OferLevi85: Added four new columns but only two of them, expiryTime and hostName can be inserted. The other two, lastNotifiedTime and lastNotifiedServer are updated as part of the Notification logic.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2678,7 +2696,7 @@ public Response postInstanceRegisterInformation(ResourceContext ctx, InstanceReg
             // able to validate the certificate during refresh operations
 
             if (insertX509CertRecord(ctx, cn, provider, certReqInstanceId, certSerial,
-                    InstanceProvider.ZTS_CERT_USAGE_CLIENT.equalsIgnoreCase(certUsage)) == null) {
+                    InstanceProvider.ZTS_CERT_USAGE_CLIENT.equalsIgnoreCase(certUsage), null, null) == null) {
```
OferLevi85: Currently, always put null (the columns are nullable). Will insert the real values once hca starts sending them as part of the request.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/CertRecordStoreConnection.java
```diff
@@ -70,4 +71,20 @@
      * @return number of records deleted
      */
     int deleteExpiredX509CertRecords(int expiryTimeMins);
+
+    /**
```
OferLevi85: Similar to the logic of updateRoleMemberExpirationNotificationTimestamp and getNotifyTemporaryRoleMembers in ZMS.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
```diff
@@ -46,6 +47,10 @@
     private static final String KEY_PREV_TIME = "prevTime";
     private static final String KEY_PREV_IP = "prevIP";
     private static final String KEY_CLIENT_CERT = "clientCert";
+    private static final String KEY_LAST_NOTIFIED_TIME = "lastNotifiedTime";
```
OferLevi85: Added 4 new nullable columns. Added tests that verify get / insert / update works correctly with null values for these columns.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
```diff
@@ -179,6 +212,18 @@ public int deleteExpiredX509CertRecords(int expiryTimeMins) {
         return 0;
     }
 
+    @Override
```
OferLevi85: Our new notification is only for ZTS so at this point these methods are not required.

servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
```diff
@@ -0,0 +1,74 @@
...
+        this.serverName = serverName;
+        this.instanceCertManager = instanceCertManager;
+        this.dataStore = dataStore;
+        ZTSDomainRoleMembersFetcher ztsDomainRoleMembersFetcher = new ZTSDomainRoleMembersFetcher(dataStore, USER_DOMAIN_PREFIX);
+        this.notificationCommon = new NotificationCommon(ztsDomainRoleMembersFetcher, userDomainPrefix);
+    }
+
+    @Override
+    public List<Notification> getNotifications() {
+        return new ArrayList<>();
```
OferLevi85: At this point this task doesn't return any notifications so no notifications will be sent in ZTS. It will be implemented after verifying the daily thread runs correctly in ZTS and the new columns are being populated.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/JDBCCertRecordStoreConnection.java
```diff
@@ -37,24 +34,35 @@
...
-    public static final String DB_COLUMN_CURRENT_SERIAL = "currentSerial";
-    public static final String DB_COLUMN_CURRENT_TIME   = "currentTime";
-    public static final String DB_COLUMN_PREV_IP        = "prevIP";
-    public static final String DB_COLUMN_PREV_SERIAL    = "prevSerial";
-    public static final String DB_COLUMN_PREV_TIME      = "prevTime";
-    public static final String DB_COLUMN_CLIENT_CERT    = "clientCert";
+    private static final String SQL_UPDATE_UNREFRESHED_X509_RECORDS_NOTIFICATION_TIMESTAMP = "UPDATE certificates SET lastNotifiedTime=?, lastNotifiedServer=? " +
+            "WHERE currentTime < (CURRENT_DATE - INTERVAL 1 DAY) AND DATEDIFF(CURRENT_TIME, currentTime) IN (1,7,14,21,28) AND (lastNotifiedTime IS NULL || lastNotifiedTime < (CURRENT_DATE - INTERVAL 1 DAY))";
+    private static final String SQL_LIST_NOTIFY_UNREFRESHED_X509_RECORDS = "SELECT * FROM certificates WHERE lastNotifiedTime=? AND lastNotifiedServer=?;";
+
```
OferLevi85: Currently implemented logic is to send notifications for certificates that didn't refresh for more than a day. Notification will be sent once a week for a month.
We can change this logic in a future pull request to consider expiration for example.

libs/java/server_common/pom.xml
```diff
@@ -163,6 +163,12 @@
...
+          <groupId>com.yahoo.athenz</groupId>
+          <artifactId>athenz-zms-core</artifactId>
+          <version>1.8.51-SNAPSHOT</version>
```
havetisyan: this must be    `<version>${project.parent.version}</version>`

libs/java/server_common/pom.xml
```diff
@@ -163,6 +163,12 @@
...
+          <scope>compile</scope>
```
havetisyan: why the scope only compile?

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
havetisyan: expiryTime is already available - it must come from the certificate that ZTS generates. only hostname is supposed to come from the service identity agent. We already get those when the host is requesting the hostname to be included in the car, so we there is no need to wait for any changes in the agent.

OferLevi85: Fixed.

libs/java/server_common/pom.xml
OferLevi85: Fixed.

libs/java/server_common/pom.xml
OferLevi85: I let Intellij add the dependency automatically. I will make sure to double-check the generated code next time. Fixed it.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
```diff
@@ -251,20 +251,21 @@ void processExpiryEntry(StringBuilder body, final String entryNames, final Strin
...
         final AtomicInteger counter = new AtomicInteger();
-        boolean status = true;
         // SES imposes a limit of 50 recipients. So we convert the recipients into batches
         if (recipients.size() > SES_RECIPIENTS_LIMIT_PER_MESSAGE) {
             final Collection<List<String>> recipientsBatch = recipients.stream()
                     .collect(Collectors.groupingBy(it -> counter.getAndIncrement() / SES_RECIPIENTS_LIMIT_PER_MESSAGE))
                     .values();
+            boolean status = true;
             for (List<String> recipientsSegment : recipientsBatch) {
-                status = sendEmailMIME(subject, body, status, recipientsSegment);
```
OferLevi85: As we talked about on a different pull request, the status will not be part of the interface

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
```diff
@@ -337,7 +338,7 @@ private MimeMessage getMimeMessage(String subject, String body, Collection<Strin
         return message;
     }
 
-    private boolean sendEmailMIME(String subject, String body, boolean status, Collection<String> recipients) {
+    private boolean sendEmailMIME(String subject, String body, Collection<String> recipients) {
```
OferLevi85: As we talked about on a different pull request, the status will not be part of the interface

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/JDBCCertRecordStoreConnection.java
OferLevi85: Note this logic was fixed in the up-coming pull request according to our last session. You may ignore it for now.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2678,7 +2700,7 @@ public Response postInstanceRegisterInformation(ResourceContext ctx, InstanceReg
             // able to validate the certificate during refresh operations
 
             if (insertX509CertRecord(ctx, cn, provider, certReqInstanceId, certSerial,
-                    InstanceProvider.ZTS_CERT_USAGE_CLIENT.equalsIgnoreCase(certUsage)) == null) {
+                    InstanceProvider.ZTS_CERT_USAGE_CLIENT.equalsIgnoreCase(certUsage), certExpiryTime, null) == null) {
```
havetisyan: we cannot use the certExpiryTime here since the cert signer may not honor what the server is asking  and instead use another expiry time. So the correct value here is to extract the expiry from the newCert that we're about send back to the client.

also, for the last hostname argument, why are we passing null instead of info.getHostname()?

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3117,7 +3129,7 @@ X509CertRecord getValidatedX509CertRecord(ResourceContext ctx, final String prov
 
             if (cert.getNotBefore().getTime() < x509CertRefreshResetTime) {
                 x509CertRecord = insertX509CertRecord(ctx, principalName, provider, instanceId,
-                        cert.getSerialNumber().toString(), false);
+                        cert.getSerialNumber().toString(), false, expiryTime, null);
```
havetisyan: this one is the recovery mode certificate record. so the correct value should be the expiry of the cert  certificate and not the expiryTime. So we can extract the expiry from `cert` object passed in and use that.

PR #910, opened 2020-03-17
Implement Unrefreshed Certificates Notification

OferLevi85:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
```diff
@@ -161,6 +167,9 @@ String getBody(String type, Map<String, String> details) {
             case NOTIFICATION_TYPE_PRINCIPAL_EXPIRY_REMINDER:
                 body = getPrincipalExpiryBody(details);
                 break;
+            case NOTIFICATION_TYPE_UNREFRESHED_CERTS:
+                body = getUnrefreshedCertsBody(details);
```
OferLevi85: As you can see, I have added the new notification in a similar manner to the other notifications. I have noticed that the EmailNotificationService requires similar refactoring to the NotificationManager (separating the notification specific logic from this class) but it will increase this pull request complexity substantially.
I have decided it is best to leave it as is for now. I can refactor this class after this pull request is merged or on a later date that might be more suitable.

PR #916, opened 2020-03-26
Added missing tests for Notifications in ZTS

OferLevi85:


servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
```diff
@@ -102,10 +102,6 @@ public CertFailedRefreshNotificationTask(InstanceCertManager instanceCertManager
         // certificateRecords := <certificate-entry>[|<certificate-entry]*
         // certificate-entry := <Service Name>;<Provider>;<InstanceID>;<Last refresh time>;<Expiration time>;<Hostname>;
 
-        if (certRecords == null || certRecords.isEmpty()) {
```
OferLevi85: Going over the tests I realized that it is not possible for certRecords to be null at this point.

PR #924, opened 2020-03-30
new zms api to delete pending members

havetisyan:


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1230,7 +1230,54 @@ void executeDeleteMembership(ResourceContext ctx, String domainName, String role
...
+    void executeDeletePendingMembership(ResourceContext ctx, String domainName, String roleName,
+                String normalizedMember, String auditRef, String caller) {
+
+        // our exception handling code does the check for retry count
+        // and throws the exception it had received when the retry
+        // count reaches 0
+
+        for (int retryCount = defaultRetryCount; ; retryCount--) {
+
+            try (ObjectStoreConnection con = store.getConnection(true, true)) {
```
abvaidya: should this be autoCommit=false?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3282,6 +3282,56 @@ void sendMembershipApprovalNotification(final String domain, final String org, f
...
+        // policy, service, etc name)
+
+        domainName = domainName.toLowerCase();
+        roleName = roleName.toLowerCase();
+        memberName = normalizeDomainAliasUser(memberName.toLowerCase());
+
+        final String principalDomain = getPrincipalDomain(ctx);
+        metric.increment(ZMSConsts.HTTP_REQUEST, domainName, principalDomain);
+        metric.increment(caller, domainName, principalDomain);
+        Object timerMetric = metric.startTiming("deletemembership_timing", domainName, principalDomain);
```
abvaidya: should this be deletependingmembership_timing?

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
havetisyan: For single operations, we have auto commit enabled.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
havetisyan: fixed.

PR #926, opened 2020-03-30
Fixed a bug where big emails caused http client to crash

OferLevi85:
httpPost.getEntity is limited to 25k and will throw an exception if it is bigger.

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/http/HttpDriver.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/http/HttpDriver.java
@@ -37,6 +37,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.net.ssl.SSLContext;
+import java.io.ByteArrayOutputStream;
 import java.io.Closeable;
 import java.io.FileNotFoundException;
 import java.io.IOException;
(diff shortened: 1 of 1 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/http/HttpDriver.java
```diff
@@ -258,7 +259,9 @@ public String doGet(final String url) throws IOException {
      */
     public String doPost(HttpPost httpPost) throws IOException {
         String url = httpPost.getURI().toString();
-        String query = EntityUtils.toString(httpPost.getEntity());
+        ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+        httpPost.getEntity().writeTo(byteArrayOutputStream);
+        String query = byteArrayOutputStream.toString();
```
havetisyan: where are we using this object - I only see two references in log debug and error points. In which case why are we doing this here at all? We can do a private method to return the query and then have something like:

if (LOGGER.isDebugEnabled()) {
LOGGER.debug(...., getPostQuery(httpPost);
}

and then further down:

} catch (IOException ex) {
     LOGGER.error(......, getPostQuery(httpPost), ...);
}

OferLevi85: Thanks you're right I should have thought about it. It will also enable me to cleanly surround it with try-catch so if it will fail in the future it won't stop the client from sending the message.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/http/HttpDriver.java
```diff
@@ -258,8 +259,7 @@ public String doGet(final String url) throws IOException {
      */
     public String doPost(HttpPost httpPost) throws IOException {
         String url = httpPost.getURI().toString();
-        String query = EntityUtils.toString(httpPost.getEntity());
-        LOGGER.debug("Requesting from {} with query {}", url, query);
+        LOGGER.debug("Requesting from {} with query {}", url, getPostQuery(httpPost));
```
havetisyan: we should put this in If (LOGGER.isDebugEnabled()) block otherwise we'll do all the work for the getPostQuery, then pass the argument to the function which will just throw it away since in production we don't have debug enabled (so the getPostQuery call is just wasted).

PR #943, opened 2020-04-22
add OAuth authority

WindzCUHK:
## TODO

1. add auth0 dependency to the release version of this PR

## changes

1. add OAuth authority: d72134011e62da612e09a471f8f35bafdb0e762a
	- related files:
		- `libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/OAuthCertBoundJwtAccessTokenAuthority.java`
		- `libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/*`
	- refactor:
		- `libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/CertificateAuthority.java`
		- `libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/AuthorityConsts.java`
		- `libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/AthenzUtils.java`
1. update jjwt.version
	- fix pom.xml: 3f05aa58128227a1d2e66191c9d29ba8462e3b5e
	- fix deprecated and test cases: d72134011e62da612e09a471f8f35bafdb0e762a
		- related files:
			- `libs/java/auth_core/*`
	- fix existing test cases: 70d39b4244ae41d5d8ba1d7173be8e3093be7a56
		- fix `io.jsonwebtoken.security.WeakKeyException` after version update
			- `clients/java/zpe/*`
			- `servers/zts/*`
1. support Auth0 as Athenz IdP: 34c8089e8639784d58ce67af6aaaff077e99d072
1. minor fixes
	1. fix dead link: c5dc3303cd0cedd325c76ed2d8d4b1bf678aad22
	1. remove bouncer: 70ae183c1b758af2517b2e36eae1234fc549916e
	1. fix whitespace: 897e503c6c7c930f57573e8b98c0dea64e101e8e

> I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/validator/DefaultOAuthJwtAccessTokenValidator.java
```diff
@@ -0,0 +1,128 @@
...
+    private void verifyScopes(OAuthJwtAccessToken jwt) throws OAuthJwtAccessTokenException {
+        List<String> scopes = jwt.getScopes();
+        if (scopes == null || !(new HashSet<>(scopes)).containsAll(this.requiredScopes)) {
+            // SOME required scopes NOT found
+            throw new OAuthJwtAccessTokenException("required scope not found: got=" + jwt.getScope());
+        }
+    }
+    private void verifyCertificateThumbprint(OAuthJwtAccessToken jwt, String certificateThumbprint) throws OAuthJwtAccessTokenException {
+        String certThumbprint = jwt.getCertificateThumbprint();
+        if (certificateThumbprint == certThumbprint) {
```
ssunorz, line 88: String values are compared using `==`, not `equals()`

WindzCUHK: It is to skip checking when both is null.
The actual equal checking it after that.
https://github.com/yahoo/athenz/blob/70d39b4244ae41d5d8ba1d7173be8e3093be7a56/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/validator/DefaultOAuthJwtAccessTokenValidator.java#L88-L94

contributions/authority/auth0/pom.xml
```diff
@@ -0,0 +1,195 @@
...
+
+  <groupId>com.yahoo.athenz</groupId>
+  <artifactId>athenz_auth_auth0</artifactId>
+  <packaging>jar</packaging>
+  <version>1.0.1</version>
+  <name>athenz_auth_auth0</name>
+  <description>Athenz Yahoo Server OAuth2 access token implementation for Auth0</description>
+
+  <properties>
+    <athenz.version>1.8.45</athenz.version>
```
havetisyan, line 31: this is quite an old version of athenz. let's use the latest release.

contributions/authority/auth0/src/main/java/com/yahoo/athenz/auth/oauth/auth0/Auth0Jwt.java
```diff
@@ -0,0 +1,107 @@
...
+import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.Jws;
+import io.jsonwebtoken.RequiredTypeException;
+
+/**
+ * Custom for Auth0 JWT access token format
+ */
+public class Auth0Jwt extends DefaultOAuthJwtAccessToken {
+
+    private static String claimClientId = "https://myapp.example.com/client_id";
```
havetisyan, line 32: where/how are we using this example.com/client_id? do we need a such value as default or is the caller expected to always specify a value?

contributions/authority/auth0/src/main/java/com/yahoo/athenz/auth/oauth/auth0/Auth0Jwt.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    @Override
+    public String getSubject() {
+        String subject = this.body.getSubject();
+        if (subject == null) {
+            return null;
+        }
+        // normalize as Athenz principal
+        return (AthenzUtils.getPrincipalName(Auth0Jwt.userDomain, subject.replace('|', '-'))).toLowerCase();
```
havetisyan, line 51: I would suggest if we have a AthenzUtils.getPrincipal method then that method should already return values in lowercase since that's what Athenz uses everywhere. Then we don't have to remember to call toLowerCase() everywhere.

libs/java/auth_core/pom.xml
```diff
@@ -48,10 +48,32 @@
...
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>io.jsonwebtoken</groupId>
+      <artifactId>jjwt-jackson</artifactId>
+      <version>${jjwt.version}</version>
+      <scope>runtime</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.auth0</groupId>
```
havetisyan, line 74: we should keep the number of dependencies in auth_core minimal so not sure why we're using this but it doesn't look right.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/CertificateAuthority.java
```diff
@@ -99,106 +96,22 @@ public Principal authenticate(X509Certificate[] certs, StringBuilder errMsg) {
...
-        String name = principalName.substring(idx + 1);
-        
-        // all the role members in Athenz are normalized to lower case so we need to make
-        // sure our principal's name and domain are created with lower case as well
-        
-        SimplePrincipal principal = (SimplePrincipal) SimplePrincipal.create(domain.toLowerCase(),
-                name.toLowerCase(), x509Cert.toString(), this);
+        // create principal
+        X509Certificate x509Cert = certId.getX509Certificate();
+        SimplePrincipal principal = (SimplePrincipal) SimplePrincipal.create(certId.getDomain(), certId.getService().toLowerCase(), x509Cert.toString(), this);
```
havetisyan, line 110: seems inconsistent that we lowercase service but not domain?

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/CertificateIdentityParser.java
```diff
@@ -0,0 +1,166 @@
...
+        // make sure the principal is not on our excluded list
+
+        if (this.excludedPrincipalSet != null && this.excludedPrincipalSet.contains(principalName)) {
+            throw new CertificateIdentityException("Principal is excluded");
+        }
+
+        // For role cert, the principal information is in the SAN email
+
+        List<String> roles = null;
+        int idx = principalName.indexOf(":role.");
```
havetisyan, line 94: we have :role. defined in AuthortyConsts so we should use that here

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/parser/KeyStoreJwkKeyResolver.java
```diff
@@ -0,0 +1,117 @@
...
+import org.slf4j.LoggerFactory;
+import io.jsonwebtoken.Claims;
+import io.jsonwebtoken.JwsHeader;
+import io.jsonwebtoken.SigningKeyResolver;
+
+/**
+ * KeyResolver that get public key from key store or JWKS URL
+ * @see <a href="https://tools.ietf.org/html/rfc7517" target="_top">RFC7517</a>
+ */
+public class KeyStoreJwkKeyResolver implements SigningKeyResolver {
```
havetisyan, line 37: why did we create this class? we already have the SigningKeyResolver without relying on auth0 implementation and it supports both rsa and ec keys. So rather than creating yet another and adding dependency on auth0 library, take a look at the current key resolver and if necessary extend that class.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/token/DefaultOAuthJwtAccessToken.java
```diff
@@ -0,0 +1,134 @@
...
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public List<String> getAudiences() {
+        // https://tools.ietf.org/html/rfc7519#page-9
+        List<String> audiences;
+        try {
+            // returns null if not found
+            audiences = this.body.get(Claims.AUDIENCE, ArrayList.class);
+        } catch (RequiredTypeException e) {
```
havetisyan, line 74: we should probably log these errors otherwise it's very difficult to debug any issues.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/token/DefaultOAuthJwtAccessToken.java
```diff
@@ -0,0 +1,134 @@
...
+    @Override
+    public String getCertificateThumbprint() {
+        // https://github.com/jwtk/jjwt/issues/404, custom model class not supported
+        LinkedHashMap<?, ?> certConf = null;
+        try {
+            certConf = this.body.get(OAuthJwtAccessToken.CLAIM_CONFIRM, LinkedHashMap.class);
+            if (certConf == null) {
+                return null;
+            }
+        } catch (RequiredTypeException e) {
```
havetisyan, line 95: lets include a log error entry for this as well.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/util/OAuthAuthorityUtils.java
```diff
@@ -0,0 +1,84 @@
...
+     * @param  csv       CSV string
+     * @param  delimiter CSV delimiter
+     * @return           corresponding Set object of the CSV string, or null if CSV is null or empty
+     */
+    public static final Set<String> csvToSet(String csv, String delimiter) {
+        if (csv == null || csv.isEmpty()) {
+            return null;
+        }
+        if (delimiter == null || delimiter.isEmpty()) {
+            return new HashSet<>(Arrays.asList(csv));
```
havetisyan, line 51: seems to be inefficient to handle just a single string. we can just create a set and add the entry without creating additional objects.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/validator/DefaultOAuthJwtAccessTokenValidator.java
havetisyan: I'd suggesting adding a comment explaining that otherwise it does look suspicious

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/AthenzUtils.java
```diff
@@ -145,11 +144,45 @@ public static String extractPrincipalDomainName(final String principalName) {
...
             return principalName.substring(idx + 1);
         }
     }
+
+    /**
+     * Split principal to domain and service
+     * @param  name principal
+     * @return      [domain, service], null if principal in invalid format
+     */
+    public static String[] splitPrincipalName(String name) {
```
havetisyan, line 160: I think I made a comment earlier about this. it looks like this method already does the to lowercase conversion thus the caller doesn't need to call it again.

ssunorz:
It seems to me that the version of the following file should also be updated.
https://github.com/yahoo/athenz/blob/9f62921f766faaa8e2fcf84743514b59502f7b36/examples/java/instance-provider/client/pom.xml#L59
https://github.com/yahoo/athenz/blob/9dc79384d1dd9b740857d5fd9a98be87d0935ec1/examples/java/instance-provider/provider/pom.xml#L160

contributions/authority/auth0/src/main/java/com/yahoo/athenz/auth/oauth/auth0/Auth0Jwt.java
WindzCUHK: Auth0 has restriction on custom field format, so we need the "https://myapp.example.com/" part as the namespace.
https://auth0.com/docs/api-auth/tutorials/adoption/scope-custom-claims#custom-claims

Also, as we enforced the client ID and cnf checking in the new authority class, the claim name should be specified. For the default value, I just use the value can be found in auth0's docs.

contributions/authority/auth0/pom.xml
WindzCUHK: Thank you for reminding.
As it depends on the latest auth_core, I just update to `1.8.60` first.
bc180c1f

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/validator/DefaultOAuthJwtAccessTokenValidator.java
WindzCUHK: added in 9f86ef6d

contributions/authority/auth0/src/main/java/com/yahoo/athenz/auth/oauth/auth0/Auth0Jwt.java
WindzCUHK: Thank you for suggesting.
I updated both util methods in 646436c8 & f8cb9ade.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/CertificateAuthority.java
WindzCUHK: Thank you for pointing out. `CertificateIdentityParser` should handle the lowercase. `toLowerCase()` is not needed here.
Fixed in 28698703 and also removed other unnecessary `toLowerCase()`.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/CertificateIdentityParser.java
WindzCUHK: Updated in 2c5c855c, and try to refactor other hard-coded `:role.` too.

hard-coded `:role.` values are preserved in these files, as their context seem to be different.
```
./servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
./servers/zts/src/main/java/com/yahoo/athenz/zts/AccessTokenRequest.java
./servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
*Test.java
```

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/parser/KeyStoreJwkKeyResolver.java
WindzCUHK: As the auth0 class has built-in refresh logic on the JWKS URL, I included the library. However, its behavior is difficult to trace in operation so I updated to depend on the existing `JwtsSigningKeyResolver` in 591d10eb.

I also made some changes on the `JwtsSigningKeyResolver` class to support file URL.
In addition, as the 2 classes behave differently, I just use weak dependency. Please let me know if extending the class is better.

libs/java/auth_core/pom.xml
WindzCUHK: removed. details: https://github.com/yahoo/athenz/pull/943#discussion_r418572576

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/token/DefaultOAuthJwtAccessToken.java
WindzCUHK: added in 24f02031

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/token/DefaultOAuthJwtAccessToken.java
WindzCUHK: added in 24f02031

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/util/OAuthAuthorityUtils.java
WindzCUHK: Thank you for pointing out. fixed: 1e7303e7

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/AthenzUtils.java
WindzCUHK: agreed. related reply: https://github.com/yahoo/athenz/pull/943#discussion_r418565694, https://github.com/yahoo/athenz/pull/943#discussion_r418566601

havetisyan:
> > It seems to me that the version of the following file should also be updated.
> > https://github.com/yahoo/athenz/blob/9f62921f766faaa8e2fcf84743514b59502f7b36/examples/java/instance-provider/client/pom.xml#L59
> > 
> > https://github.com/yahoo/athenz/blob/9dc79384d1dd9b740857d5fd9a98be87d0935ec1/examples/java/instance-provider/provider/pom.xml#L160
> 
> @havetisyan
> 
> * Are the examples intentionally NOT using the jjwt version in the root `pom.xml`?
> * Should I update their version too? or, it is intended for testing backward compactibility?
>   https://github.com/yahoo/athenz/blob/64747d2d858e622b0c8e0b56e7caaa45a57db0f5/pom.xml#L88



> > It seems to me that the version of the following file should also be updated.
> > https://github.com/yahoo/athenz/blob/9f62921f766faaa8e2fcf84743514b59502f7b36/examples/java/instance-provider/client/pom.xml#L59
> > 
> > https://github.com/yahoo/athenz/blob/9dc79384d1dd9b740857d5fd9a98be87d0935ec1/examples/java/instance-provider/provider/pom.xml#L160
> 
> @havetisyan
> 
> * Are the examples intentionally NOT using the jjwt version in the root `pom.xml`?
> * Should I update their version too? or, it is intended for testing backward compactibility?
>   https://github.com/yahoo/athenz/blob/64747d2d858e622b0c8e0b56e7caaa45a57db0f5/pom.xml#L88

Yes, those are meant to be standalone examples and not dependent on the athenz repo or parent pom. the idea is I can just copy the directory only and run mvn install and it will build my package without requiring anything else. so once in a while I go through those examples (there are some in the zts client I think) and update to the latest release packages.

havetisyan:
Thanks @WindzCUHK. Next time please smaller PRs :)

PR #948, opened 2020-04-25
Add new NotificationTask - RoleMemberReviewNotificationTask

OferLevi85:
1. New notification implemented without required DB changes.
DB changes will be implemented in a future commit
2. reviewReminder field was added to rdl structs RoleMember, MemberRole and Membership. reviewLastNotifiedTime was added to RoleMember.


libs/java/server_common/src/main/resources/messages/ServerCommon.properties
```diff
@@ -10,3 +10,9 @@ athenz.notification.email.principal.expiry.body.entry=<tr><td class="cv">{0}</td
 
 athenz.notification.email.unrefreshed.certs.subject=Athenz Unrefreshed Certificates Notification
 athenz.notification.email.unrefreshed.certs.body.entry=<tr><td class="cv">{0}</td><td class="cv">{1}</td><td class="cv">{2}</td><td class="cv">{3}</td><td class="cv">{4}</td><td class="cv">{5}</td></tr>
```
OferLevi85, line 12: have left ServerCommon.properties in common even tough this file contains notification specific attributes.
I will move these attributes to files in zms / zts on a separate pull request.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4226,6 +4226,11 @@ DomainRoleMembership getPendingDomainRoleMembers(final String principal) {
         return null;
     }
 
+    public Map<String, DomainRoleMember> getReviewMembers() {
```
OferLevi85, line 4229: DB is implemented in PR 951

havetisyan:
Are we combining the notifications for both review and expiry cases? I don't think we should because with expiry we like to specify that you will lose access while with review, it might say you'll be referred to security team for review but will not lose access - major distinction that probably we don't want to handle with a same email template.

OferLevi85:
Hi @havetisyan,
No, we do not combine them. They have separate email templates (for Review notification tasks, the 2 new HTML templates are principal-review.html and domain-member-review.html. Also, the email converts are different from role expiry (the classes responsible for turning the notification into an email). The DB query is also different (on a different pull request).
There is a lot of common functionality between Role Expiry and Review Reminder and I merged this functionality (both sending notifications to principal and then to domain admins) but other than that they are different.

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/RoleMemberReviewNotificationTask.java
```diff
@@ -0,0 +1,159 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import static com.yahoo.athenz.common.server.notification.NotificationServiceConstants.*;
+
+public class RoleMemberReviewNotificationTask implements NotificationTask {
+    private final DBService dbService;
+    private final RoleMemberNotificationCommon roleMemberNotificationCommon;
+    private static final Logger LOGGER = LoggerFactory.getLogger(RoleMemberReviewNotificationTask.class);
+    private final static String DESCRIPTION = "Review before expiration reminder";
```
havetisyan, line 37: A better description would be: "Periodic Review Reminder"

this feature is not really about expiration. it's just about reviewing members without expiring anyone.

servers/zms/src/main/resources/messages/domain-member-review.html
```diff
@@ -0,0 +1,36 @@
...
+        </div>
+        <div class="hdr">Domain Membership Review Details</div>
+        <div class="bt">You have one or more principals in your Athenz domain <b>{0}</b> whose review date will pass soon:</div>
+        <hr>
+        <table id="t02">
+            <thead>
+            <tr>
+                <th class="ch">MEMBER</th>
+                <th class="ch">ROLE</th>
+                <th class="ch">EXPIRATION</th>
```
havetisyan, line 22: we should remove EXPIRATION from this reminder. The user will get those separately so let's not combine and possibly confuse the users.

servers/zms/src/main/resources/messages/domain-member-review.html
```diff
@@ -0,0 +1,36 @@
...
+                <th class="ch">REVIEW</th>
+            </tr>
+            </thead>
+            <tbody></tbody>
+        </table>
+        <hr>
+        <div class="bt"><br>Please review this list and, if necessary, login to <a href="{1}/domain/{0}/role">Athenz UI</a> to extend their expiration and review dates.</div>
```
havetisyan, line 29: We should remove any references to expiration.

servers/zms/src/main/resources/messages/principal-review.html
```diff
@@ -0,0 +1,36 @@
...
+    <style></style>
+</head>
+<body style="margin: 0; background: rgba(231, 232, 231, 1.0);">
+<div class="athenz-wrapper">
+    <div class="mbrapproval">
+        <div class="logo">
+            <img src="cid:logo" class="athenzlogowhite" alt="Athenz logo"/>
+        </div>
+        <div class="hdr">Principal Review Details</div>
+        <div class="bt">Review date for principal <b>{0}</b> in the following roles will expire soon:</div>
```
havetisyan, line 15: it should say - "You are required to review your <b>{0}</b> principal's membership in the listed roles:"

servers/zms/src/main/resources/messages/principal-review.html
```diff
@@ -0,0 +1,36 @@
...
+                <th class="ch">DOMAIN</th>
+                <th class="ch">ROLE</th>
+                <th class="ch">EXPIRATION</th>
+                <th class="ch">REVIEW</th>
+            </tr>
+            </thead>
+            <tbody></tbody>
+        </table>
+        <hr>
+        <div class="bt"><br>Please review this list and, if necessary, follow up with their respective domain administrators to extend those expiration and review dates.</div>
```
havetisyan, line 29: let's remove the "those expiration" part from the message.

servers/zms/src/main/resources/messages/principal-review.html
```diff
@@ -0,0 +1,36 @@
...
+        </div>
+        <div class="hdr">Principal Review Details</div>
+        <div class="bt">Review date for principal <b>{0}</b> in the following roles will expire soon:</div>
+        <hr>
+        <table id="t03">
+            <thead>
+            <tr>
+                <th class="ch">DOMAIN</th>
+                <th class="ch">ROLE</th>
+                <th class="ch">EXPIRATION</th>
```
havetisyan, line 22: let's remove this expiration line as well.

PR #950, opened 2020-04-27
modify using localtime for retrieving modified domains

TakuyaMatsu:
## Problem
ZMS uses MySQL Connector/J for DB connections. MySQL Connector/J converted to DB's time zone until 5.x, but from 8.x, it doesn't convert to DB's time zone.

- 5.x (expected behavior): MySQL Connector/J converts GMT to JST when the DB time zone is JST.
- 8.x (Unintended behavior): MySQL Connector/J will not convert GMT even if the DB time zone is JST

When retrieving modified domains from database server, ZMS generates the query with GMT, but the time zone of database is not GMT, so I cannot retrieve modified domains.

Is there any way to convert GMT to localtime when retrieving modified domains?

## Fix
This is an idea to solve this problem.

### DIFF
--- a/servers/zms/conf/zms.properties
+++ b/servers/zms/conf/zms.properties
@@ -415,3 +415,7 @@ athenz.zms.read_only_mode=false
 # needs to contact most likely a server (e.g. mysql instance)
 # running in a different region.
 #athenz.zms.master_copy_for_signed_domains=false
+
+# Boolean value specifying whether using the server's timezone or GMT
+# at retriving modified domains from DB
+#athenz.zms.use_local_timezone_in_domain_scan=false
--- a/servers/zms/conf/zms.properties
+++ b/servers/zms/conf/zms.properties
@@ -416,6 +416,6 @@ athenz.zms.read_only_mode=false
 # running in a different region.
 #athenz.zms.master_copy_for_signed_domains=false
 
-# Boolean value specifying whether using the server's timezone or GMT
-# at retriving modified domains from DB
-#athenz.zms.use_local_timezone_in_domain_scan=false
+# Set the timezone of the database
+# when retrieving the modified domain.
+#athenz.zms.athenz.zms.mysql_server_timezone=
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
@@ -651,8 +651,6 @@ public class JDBCConnection implements ObjectStoreConnection {
     PreparedStatement prepareDomainScanStatement(String prefix, long modifiedSince)
             throws SQLException {
 
-        Calendar cal = Calendar.getInstance(TimeZone.getTimeZone(MYSQL_SERVER_TIMEZONE));
-
         PreparedStatement ps;
         if (prefix != null && prefix.length() > 0) {
             int len = prefix.length();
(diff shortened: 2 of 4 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -648,6 +649,8 @@ public boolean deleteDomain(String domainName) {
     PreparedStatement prepareDomainScanStatement(String prefix, long modifiedSince)
             throws SQLException {
         
+        Calendar cal = getCalendarInstance();
```
havetisyan, line 652: this becomes a wasted operation if the command itself does not the calendar instance. please retain the original behavior where the calendar is obtained only when its needed.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -666,13 +668,27 @@ PreparedStatement prepareDomainScanStatement(String prefix, long modifiedSince)
...
             ps.setTimestamp(1, new java.sql.Timestamp(modifiedSince), cal);
         } else {
             ps = con.prepareStatement(SQL_LIST_DOMAIN);
         }
         return ps;
     }
+
+    Calendar getCalendarInstance() {
+
+        Boolean useLocalTimeZone = Boolean.parseBoolean(
```
havetisyan, line 680: we should not fetch the property and parse it every time. this should be a static member of the class  so we only do it once and never again.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -647,7 +650,9 @@ public boolean deleteDomain(String domainName) {
 
     PreparedStatement prepareDomainScanStatement(String prefix, long modifiedSince)
             throws SQLException {
-        
+
+        Calendar cal = Calendar.getInstance(TimeZone.getTimeZone(MYSQL_SERVER_TIMEZONE));
```
havetisyan, line 654: let's move this code how it was before. as I mentioned in my earlier comment we don't need a calendar object if we're just doing a list domain operation.

servers/zms/conf/zms.properties
```diff
@@ -415,3 +415,7 @@ athenz.zms.read_only_mode=false
 # needs to contact most likely a server (e.g. mysql instance)
 # running in a different region.
 #athenz.zms.master_copy_for_signed_domains=false
+
+# Boolean value specifying whether using the server's timezone or GMT
+# at retriving modified domains from DB
+#athenz.zms.use_local_timezone_in_domain_scan=false
```
havetisyan, line 421: please update this setting to reflect the new name.

ssunorz:
For example, is it necessary to use a calendar object in the following code?https://github.com/yahoo/athenz/blob/8a2db8f8407c21e706048594e827d2f2ddeaa6d4/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java#L3819

havetisyan:
> For example, is it necessary to use a calendar object in the following code?
> 
> https://github.com/yahoo/athenz/blob/8a2db8f8407c21e706048594e827d2f2ddeaa6d4/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java#L3819

the value on line must match the value set on the following one:

https://github.com/yahoo/athenz/blob/8a2db8f8407c21e706048594e827d2f2ddeaa6d4/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java#L3879

because you might have multiple servers sending notifications, you want one server to mark some as its own and then process them - this is how we use the timestamp and server attributes. so it's not necessary to use the calendar object but those two lines must match.

PR #955, opened 2020-04-29
Template

jothi-prasad:
Adding RDL changes plus unit tests to Template structure. this is to support metadata fields in Athenz template (solution-templates.json)
 


core/zms/src/main/rdl/Template.tdl
```diff
@@ -7,11 +7,22 @@ include "Role.tdl";
 include "Policy.tdl";
 include "ServiceIdentity.tdl"
 
+//MetaData for template.
+type MetaData Struct {
```
havetisyan, line 11: let's call this something more specific like TemplateMetaData. Since this ends up as a Class name, MetaData is too generic

core/zms/src/main/rdl/Template.tdl
```diff
@@ -7,11 +7,22 @@ include "Role.tdl";
...
+    String description (optional); //description of the template
+    Int32 currentVersion (optional); //Version from DB(zms_store->domain_template->version)
+    Int32 updatedVersion (optional); //Bumped up version from solutions-template.json when there is a change
```
havetisyan, line 14: instead of updatedVersion, let's call this latestVersion.

core/zms/src/main/rdl/Template.tdl
jothi-prasad: Done.

core/zms/src/main/rdl/Template.tdl
jothi-prasad: Done. changed and checked in

PR #961, opened 2020-05-04
Paranoids review endpoint

OferLevi85:

### DIFF
--- a/core/zms/src/main/rdl/Role.rdli
+++ b/core/zms/src/main/rdl/Role.rdli
@@ -103,6 +103,19 @@ resource Membership GET "/domain/{domainName}/role/{roleName}/member/{memberName
     }
 }
 
+//Get members with overdue review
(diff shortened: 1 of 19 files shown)


core/zms/src/main/rdl/Role.rdli
```diff
@@ -103,6 +103,19 @@ resource Membership GET "/domain/{domainName}/role/{roleName}/member/{memberName
     }
 }
 
+//Get members with overdue review
+resource DomainRoleMembers GET "/domain/{domainName}/overdue" (name=GetOverdueReview) {
+    DomainName domainName; //name of the domain
+    authorize ("access", "sys.auth:overdue-review");
```
OferLevi85, line 109: Not sure if this is the correct way to authorize the users. If it isn't let me know and I'll do the necessary changes.

havetisyan: we don't need authorize here as any domain admin should be able to get the list of members that need to be reviewed. so let's just make it authenticate.

PR #971, opened 2020-05-07
RDL changes to expose template details in UI or any other client.

jothi-prasad:
Added the necessary end point client should call plus additional metadata fields needed.


jothi-prasad:
> whenever we add a new api entry, the zms server must be updated accordingly and the code should be checked in as well. you can start with an empty implementation that just returns an exception and then have the implementation in another pr.
> 
> also need to ZMSClient.java to have a wrapper function (along with a test case). the auto generated code only updates the internal client which is not exposed to users.

got it. i added the implementation code plus client wrapper function and tests in this pr - https://github.com/yahoo/athenz/pull/972. thought the pr would be big if i checkin zms-core and server code at the same time. ignore this PR and i will close it. the complete implementation is in PR-972

PR #972, opened 2020-05-08
Template api implementation to expose metadata details

jothi-prasad:


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2603,6 +2603,55 @@ public Template getTemplate(ResourceContext ctx, String templateName) {
         return template;
     }
 
+    @Override
+    public DomainTemplateDetailsList getDomainTemplateDetailsList(ResourceContext ctx, String domainName) {
+        DomainTemplateDetailsList domainTemplateDetailsList = new DomainTemplateDetailsList();
+        List<TemplateMetaData> templateMetaDataList = new ArrayList<>();
+        Map<String, Integer> templateDomainMapping = new HashMap<>();
```
havetisyan, line 2610: these should be created after the validation of data otherwise we're just wasting resources if the request is invalid

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2603,6 +2603,55 @@ public Template getTemplate(ResourceContext ctx, String templateName) {
...
+                        metaData.setLatestVersion(template.getMetadata().getLatestVersion());
+                        metaData.setAutoUpdate(template.getMetadata().getAutoUpdate());
+                        metaData.setDescription(template.getMetadata().getDescription());
+                        metaData.setKeywordsToReplace(template.metadata.getKeywordsToReplace());
+                        metaData.setTimestamp(template.metadata.getTimestamp());
+                        templateMetaDataList.add(metaData);
+                    }
+                }
+                domainTemplateDetailsList.setMetaData(templateMetaDataList);
+            } catch (Exception ex) {
```
havetisyan, line 2646: what are the possible cases for exception and why are hiding them?

otherwise it looks like if there is some type of exception we're going to return an empty list to the client which is not correct.

so unless we're absolutely sure we want to swallow the exception, we shouldn't catch it and let an error be returned to the client

servers/zms/src/main/java/com/yahoo/athenz/zms/store/ObjectStoreConnection.java
```diff
@@ -146,4 +146,6 @@
     boolean updateRoleMemberReviewNotificationTimestamp(String server, long timestamp);
 
     DomainRoleMembers listOverdueReviewRoleMembers(String domainName);
+
+    Map<String, Integer> getDomainTemplates(String domainName);
```
havetisyan, line 150: Not sure why we're returning Map<String, Integer> here. if we add additional attributes, then this is not going to work thus is going to be changed again.

Looking at the server side, it looks like we're taking this Map<String, Integer> and then we generate List<TemplateMetaData> by filling the rest of the fields in the template meta objects. So why not just return here. the api should be:

List<TemplateMetaData> getDomainTemplates(String domainName);

then you just need fill in the other fields before returning to the client. in the future, if we need to add additional attributes, it would just require updating the TemplateMetaData object and not the api itself

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
jothi-prasad: there is absolutely no situation where this code will throw an exception that needs to be handled because it is mostly local object sets and list/map functions. even if it is throwing an error it should be thrown back to the client YES!. i added it while simulating a test scenario in dev box and did not remove. i will take care of this

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
jothi-prasad: Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
jothi-prasad: Checked in updated changes

servers/zms/src/main/java/com/yahoo/athenz/zms/store/ObjectStoreConnection.java
jothi-prasad: replaced Map with List<TemplateMetaData> object

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2607,6 +2607,53 @@ public Template getTemplate(ResourceContext ctx, String templateName) {
...
+        // all domain into lower case
+
+        domainName = domainName.toLowerCase();
+        final String principalDomain = getPrincipalDomain(ctx);
+        Object timerMetric = metric.startTiming("getdomainrolemembers_timing", domainName, principalDomain);
+
+        List<TemplateMetaData> templateDomainMapping = dbService.getDomainTemplates(domainName);
+        if (templateDomainMapping != null) {
+            domainTemplateDetailsList = new DomainTemplateDetailsList();
+            templateMetaDataList = new ArrayList<>();
```
havetisyan, line 2632: there is no need to create this list again. we already have a list returned from dbservice. we should just add attributes and set that and return as part of our data. This way we're not just creating another set of objects which contains the same data as the original list. So just go through the list that we got on line 2629, update the objects and then on line 2650 we use that to set the meta data.

if we come across an entry that has not value then we should send it back since that shows something not right in the db - where we have a template that no longer exists in the server.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2607,6 +2607,53 @@ public Template getTemplate(ResourceContext ctx, String templateName) {
         return template;
     }
 
+    @Override
+    public DomainTemplateDetailsList getDomainTemplateDetailsList(ResourceContext ctx, String domainName) {
+        DomainTemplateDetailsList domainTemplateDetailsList = null;
+        List<TemplateMetaData> templateMetaDataList;
```
havetisyan, line 2613: this one is not needed, please see my comment below.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2607,6 +2607,53 @@ public Template getTemplate(ResourceContext ctx, String templateName) {
         return template;
     }
 
+    @Override
+    public DomainTemplateDetailsList getDomainTemplateDetailsList(ResourceContext ctx, String domainName) {
+        DomainTemplateDetailsList domainTemplateDetailsList = null;
```
havetisyan, line 2612: we should move this down before line 2630 to be consistent with all other methods in the class.

jothi-prasad: Make sense Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
jothi-prasad: good catch. Done. reused the same object that comes back from DBservice.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
jothi-prasad: done.

PR #975, opened 2020-05-13
zms-cli expose overdue review members functionality

OferLevi85:

### DIFF
--- a/libs/go/zmscli/cli.go
+++ b/libs/go/zmscli/cli.go
@@ -110,6 +110,15 @@ func (cli *Zms) EvalCommand(params []string) (*string, error) {
 				return cli.LookupDomainById("", &productID)
 			}
 			return cli.helpCommand(params)
+		case "overdue-review":
+			if argc == 1 {
+				//override the default domain, this command can show any of them
+				dn = args[0]
+			}
+			if dn != "" {
+				return cli.ShowOverdueReview(dn)
+			}
+			return nil, fmt.Errorf("No domain specified")
 		case "use-domain":
 			var s string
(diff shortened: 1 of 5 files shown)


libs/go/zmscli/cli.go
```diff
@@ -1861,7 +1913,7 @@ func (cli Zms) HelpSpecificCommand(interactive bool, cmd string) string {
 		buf.WriteString("   list-pending-members user.john\n")
 	case "put-membership-decision":
 		buf.WriteString(" syntax:\n")
-		buf.WriteString("   " + domain_param + " put-membership-decision role member [expiration] approval\n")
+		buf.WriteString("   " + domain_param + " put-membership-decision role member [expiration] approval [review]\n")
```
havetisyan, line 1916: I don't think we're adding review here for the put-membership-decision, right?

libs/go/zmscli/cli.go
```diff
@@ -1956,7 +2013,7 @@ func (cli Zms) HelpListCommand() string {
 	buf.WriteString("   set-role-cert-expiry-mins group_role cert-expiry-mins\n")
 	buf.WriteString("   set-role-token-sign-algorithm group_role algorithm\n")
 	buf.WriteString("   set-role-notify-roles group_role rolename[,rolename...]\n")
-	buf.WriteString("   put-membership-decision group_role user_or_service [expiration] decision\n")
+	buf.WriteString("   put-membership-decision group_role user_or_service [expiration] decision [review]\n")
```
havetisyan, line 2016: same comment as above

PR #980, opened 2020-05-16
remove a problematic file

LCamel:


havetisyan:
Thanks. Not sure how that got checked in - looks like we missed during one of the reviews.

PR #984, opened 2020-05-20
Added caching to ZTSClient

gilad-bendor:

### DIFF
--- a/clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClientCache.java
+++ b/clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClientCache.java
@@ -1,0 +1,285 @@
+/*
+ * Copyright 2016 Yahoo Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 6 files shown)


clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClientCache.java
```diff
@@ -0,0 +1,285 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+
+/**
+ * Provides optional caching for {@link ZTSClient}. <br>
+ * This class is designed to support multiple caches - but is currently only support a
+ *  cache from {domain+principal} to {@link RoleAccess} - see
+ *  {@link com.yahoo.athenz.zts.ZTSClient#getRoleAccess(java.lang.String, java.lang.String)}.  <br><br>
+ *
+ * TODO: Henry - what about:
```
havetisyan, line 50: For now, we don't need to. The getAccess is used for centralized authorization checks and a proper usage is only for small number of requests and in that case you want to get the changes to any role/policy asap and caching would just delay the response. Though it should be very simple and easy to add additional caches when use cases materialize.

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClientCache.java
```diff
@@ -0,0 +1,285 @@
...
+ *      {@link ZTSClient#getAccess(java.lang.String, java.lang.String, java.lang.String)}
+ *      {@link ZTSClient#getResourceAccess(java.lang.String, java.lang.String, java.lang.String, java.lang.String)} ?
+ *
+ * TODO: Henry - no unit-tests - need to discuss
```
havetisyan, line 54: It should be very simple to add the unit tests for the caching case. We already provide a setZTSRDLGeneratedClient method where you can just pass a mock for the zts client which can just return the data first time it's called but throws an exception when the second time it's called. with caching we should only get it called once and we should never get an exception.

PR #986, opened 2020-05-20
Enforce user authority restrictions daily

havetisyan:


servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
```diff
@@ -15,6 +15,7 @@
  */
 package com.yahoo.athenz.zms;
 
+import com.amazonaws.annotation.SdkTestInternalApi;
```
abvaidya, line 18: unused import?

havetisyan: yup, not sure why that got auto-added??

PR #991, opened 2020-05-24
Fix UI issues in Role Member Review Reminder

OferLevi85:
1. Moved the Role Member Reminder Icon to the left so it can be hovered without being covered by the 'X'
2. Disabled the Reminder icon title which caused it to appear on top of the tooltip

### DIFF
--- a/ui/src/components/denali/icons/Icon.js
+++ b/ui/src/components/denali/icons/Icon.js
@@ -36,7 +36,7 @@ const Icon = (props) => {
             ref={props.innerRef}
             data-testid='icon'
         >
-            <title>{props.icon}</title>
+            {props.enableTitle && <title>{props.icon}</title>}
             {icon.map((path, index) => (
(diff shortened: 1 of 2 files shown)


ui/src/components/denali/icons/Icon.js
```diff
@@ -59,12 +59,15 @@ Icon.propTypes = {
...
     innerRef: P.oneOfType([P.func, P.object]),
+    /*By default icon will display name on hover. Setting this to false will stop it*/
+    enableTitle: PropTypes.bool
 };
 
 Icon.defaultProps = {
     icon: 'x',
     size: '1em',
     verticalAlign: 'text-bottom',
+    enableTitle: true
```
OferLevi85, line 70: I have set the default behaviour to true to prevent changes in other users of Icon.js. However, I can see that there are locations that might benefit from setting it to false (such as the icons that appear in the list of role members)

PR #995, opened 2020-05-28
update DB version when a client applies template

jothi-prasad:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
@@ -406,6 +406,29 @@ public class FileConnection implements ObjectStoreConnection {
     }
 
     @Override
+    public boolean updateDomainTemplate(String domainName, String templateName, TemplateMetaData templateMetaData) {
+        DomainStruct domainStruct = getDomainStruct(domainName);
+        ArrayList<TemplateMetaData> templateMetalist = new ArrayList<>();
+        if (domainStruct == null) {
+            throw ZMSUtils.error(ResourceException.NOT_FOUND, "domain not found", "updateDomainTemplate");
+        }
+        if (domainStruct.getTemplates() == null) {
+            domainStruct.setTemplates(new ArrayList<>());
+        }
+        ArrayList<String> templates = domainStruct.getTemplates();
+        if (!templates.contains(templateName)) {
+            templates.add(templateName);
+        }
+        TemplateMetaData templateMeta = new TemplateMetaData();
+        templateMeta.setCurrentVersion(templateMetaData.getLatestVersion());
+        templateMetalist.add(templateMeta);
+        domainStruct.setTemplateMeta(templateMetalist);
+
+        putDomainStruct(domainName, domainStruct);
+        return true;
+    }
+
+    @Override
     public boolean deleteDomainTemplate(String domainName, String templateName, String params) {
         DomainStruct domainStruct = getDomainStruct(domainName);
         if (domainStruct == null) {
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
@@ -428,18 +428,12 @@ public class FileConnection implements ObjectStoreConnection {
             for (TemplateMetaData meta : templateMetaList) {
                 if (meta.getTemplateName().equals(templateName)) {
                     meta.setCurrentVersion(templateMetaData.getLatestVersion());
+                } else {
+                    domainStruct.setTemplateMeta(setTemplateNameAndVersion(templateName, templateMetaData));
                 }
             }
         } else {
-            TemplateMetaData templateMeta = new TemplateMetaData();
-            templateMeta.setTemplateName(templateName);
-            templateMeta.setKeywordsToReplace(templateMetaData.getKeywordsToReplace());
-            templateMeta.setAutoUpdate(templateMetaData.getAutoUpdate());
-            templateMeta.setDescription(templateMetaData.getDescription());
-            templateMeta.setTimestamp(templateMetaData.getTimestamp());
-            templateMeta.setCurrentVersion(templateMetaData.getLatestVersion());
-            templateMetalist.add(templateMeta);
-            domainStruct.setTemplateMeta(templateMetalist);
+            domainStruct.setTemplateMeta(setTemplateNameAndVersion(templateName, templateMetaData));
         }
 
         putDomainStruct(domainName, domainStruct);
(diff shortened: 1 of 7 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -405,6 +405,29 @@ public boolean insertDomainTemplate(String domainName, String templateName, Stri
         return true;
     }
 
+    @Override
+    public boolean updateDomainTemplate(String domainName, String templateName, TemplateMetaData templateMetaData) {
```
havetisyan, line 409: this implementation doesn't look correct to me. we're setting a templateMetaList object which contains only a single templateMeta which only has a version number. where are the rest of the templates or where are the template names we're storing?

jothi-prasad: done.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -405,6 +405,34 @@ public boolean insertDomainTemplate(String domainName, String templateName, Stri
...
+        }
+        TemplateMetaData templateMeta = new TemplateMetaData();
+        templateMeta.setTemplateName(templateName);
+        templateMeta.setKeywordsToReplace(templateMetaData.getKeywordsToReplace());
+        templateMeta.setAutoUpdate(templateMetaData.getAutoUpdate());
+        templateMeta.setDescription(templateMetaData.getDescription());
+        templateMeta.setTimestamp(templateMetaData.getTimestamp());
+        templateMeta.setCurrentVersion(templateMetaData.getLatestVersion());
+        templateMetalist.add(templateMeta);
+        domainStruct.setTemplateMeta(templateMetalist);
```
havetisyan, line 430: we're creating the template meta list every time this api is called and replacing the struct in domain struct. which means it will only have a single template and will be overwritten every time. The implementation needs to be:

if (domainStruct.getTemplateMeta() == null) {
             domainStruct.setTemplateMeta(new ArrayList<>());
 }
ArrayList<TemplateMetaData> templateMetalist = domainStruct.getTemplateMeta();

// then iterate through the list and see if we have an object
// with that template. If yes, update the version only,
// if not then create a new object and add it to the list

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -405,6 +405,47 @@ public boolean insertDomainTemplate(String domainName, String templateName, Stri
...
+        if (!templates.contains(templateName)) {
+            templates.add(templateName);
+        }
+
+        if (domainStruct.getTemplateMeta() == null) {
+            domainStruct.setTemplateMeta(new ArrayList<>());
+        }
+        ArrayList<TemplateMetaData> templateMetaList = domainStruct.getTemplateMeta();
+        if (!templateMetaList.isEmpty()) {
+            for (TemplateMetaData meta : templateMetaList) {
```
havetisyan, line 428: we're missing the handling of the when when templateMetaList is not empty but it doesn't contain the templateName we're updating. In that case, we also need to add a new object.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -405,6 +405,47 @@ public boolean insertDomainTemplate(String domainName, String templateName, Stri
...
+                    meta.setCurrentVersion(templateMetaData.getLatestVersion());
+                }
+            }
+        } else {
+            TemplateMetaData templateMeta = new TemplateMetaData();
+            templateMeta.setTemplateName(templateName);
+            templateMeta.setKeywordsToReplace(templateMetaData.getKeywordsToReplace());
+            templateMeta.setAutoUpdate(templateMetaData.getAutoUpdate());
+            templateMeta.setDescription(templateMetaData.getDescription());
+            templateMeta.setTimestamp(templateMetaData.getTimestamp());
```
havetisyan, line 439: these lines 436-439  are not necessary since the domain object only needs to contain the name of the template and the latest version and no other detail.

jothi-prasad: Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
jothi-prasad: Done.

havetisyan:
`    @Override
    public boolean updateDomainTemplate(String domainName, String templateName, TemplateMetaData templateMetaData) {
        DomainStruct domainStruct = getDomainStruct(domainName);
        if (domainStruct == null) {
            throw ZMSUtils.error(ResourceException.NOT_FOUND, "domain not found", "updateDomainTemplate");
        }
        if (domainStruct.getTemplates() == null) {
            domainStruct.setTemplates(new ArrayList<>());
        }
        ArrayList<String> templates = domainStruct.getTemplates();
        if (!templates.contains(templateName)) {
            templates.add(templateName);
        }
        if (domainStruct.getTemplateMeta() == null) {
            domainStruct.setTemplateMeta(new ArrayList<>());
        }
        TemplateMetaData domTemplate = null;
        ArrayList<TemplateMetaData> templateMetaList = domainStruct.getTemplateMeta();
        for (TemplateMetaData meta : templateMetaList) {
            if (meta.getTemplateName().equals(templateName)) {
                domTemplate = meta;
                break;
            }
        }
        if (domTemplate == null) {
            domTemplate = new TemplateMetaData();
            domTemplate.setTemplateName(templateName);
            domTemplate.setCurrentVersion(templateMetaData.getLatestVersion());
            templateMetaList.add(domTemplate);
        } else {
            domTemplate.setCurrentVersion(templateMetaData.getLatestVersion());
        }

        putDomainStruct(domainName, domainStruct);
        return true;
    }`

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -405,6 +405,41 @@ public boolean insertDomainTemplate(String domainName, String templateName, Stri
...
+
+        if (domainStruct.getTemplateMeta() == null) {
+            domainStruct.setTemplateMeta(new ArrayList<>());
+        }
+        ArrayList<TemplateMetaData> templateMetaList = domainStruct.getTemplateMeta();
+        if (!templateMetaList.isEmpty()) {
+            for (TemplateMetaData meta : templateMetaList) {
+                if (meta.getTemplateName().equals(templateName)) {
+                    meta.setCurrentVersion(templateMetaData.getLatestVersion());
+                } else {
```
havetisyan, line 431: this is not still correct since if you have 2 template in your list, when comparing the first element, you're just overriding the full list.

https://github.com/yahoo/athenz/pull/1000

I created a new PR based on your PR. Remove the setTemplateNameAndVersion method from your PR and use the implementation of updateDomainTemplate method from that PR.

jothi-prasad: Done.

PR #1001, opened 2020-06-05
Auto Apply template.

jothi-prasad:
This Feature lets you set the template metadata field called Autoupdate to "true" and when there is a template version bump, the changes will be automatically applied to the template.


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4695,6 +4695,25 @@ void processUserAuthorityRestrictions() {
         }
     }
 
+    Map<String, List<String>> getDomainNamesFromTemplate(Map<String, Integer> templateDetails) {
```
havetisyan, line 4698: the name of the method does not reflect what the method is doing. looks like we're actually doing the auto update there and not just getting domain names from the template

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -473,6 +473,44 @@ public boolean deleteDomainTemplate(String domainName, String templateName, Stri
...
+            ArrayList<TemplateMetaData> templateMetaList = domainStruct.getTemplateMeta();
+
+            if (templateMetaList != null) {
+                for (TemplateMetaData meta : templateMetaList) {
+                    for (String templateName : templateNameAndLatestVersion.keySet()) {
+                        if (meta.getTemplateName().equals(templateName) && meta.getCurrentVersion() < templateNameAndLatestVersion.get(templateName)) {
+                            if (domainNameTemplateListMap.get(domainName) != null) {
+                                List<String> tempTemplateList = domainNameTemplateListMap.get(domainName);
+                                tempTemplateList.add(meta.getTemplateName());
+                                domainNameTemplateListMap.put(domainName, tempTemplateList);
```
havetisyan, line 499: no need for this put since we're modifying the existing list

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -870,7 +870,33 @@ public boolean deleteDomainTemplate(String domainName, String templateName, Stri
...
+
+    @Override
+    public Map<String, List<String>> getDomainFromTemplateName(Map<String, Integer> templateNameAndLatestVersion) {
+        final String caller = "getDomainsFromTemplate";
+        Map<String, List<String>> domainNameTemplateListMap = new HashMap<>();
+
+        try (PreparedStatement ps = con.prepareStatement(ZMSUtils.generateDomainTemplateVersionQuery(templateNameAndLatestVersion))) {
+            try (ResultSet rs = executeQuery(ps, caller)) {
+                while (rs.next()) {
+                    if (domainNameTemplateListMap.get(rs.getString(ZMSConsts.DB_COLUMN_NAME)) != null) {
```
havetisyan, line 882: instead of calling rs.getString multiple times in the if block we should just get those two values and assign to local variables and then do the checks with those values which makes the code more readable.

jothi-prasad: Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
jothi-prasad: Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
jothi-prasad: Done changed it.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -496,8 +497,13 @@ public ZMSImpl() {
         // Initialize Notification Manager
 
         setNotificationManager();
+
+        //autoupdate templates
+
+        autoApplyTemplates();
```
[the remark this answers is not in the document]
jothi-prasad: Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -890,6 +896,31 @@ void loadSolutionTemplates() {
             LOG.error("Generating empty solution template list...");
             serverSolutionTemplates = new SolutionTemplates();
             serverSolutionTemplates.setTemplates(new HashMap<>());
+        } else {
+            eligibleTemplatesForAutoUpdate = new HashMap<>();
+            for (String templateName : serverSolutionTemplates.getTemplates().keySet()) {
+                Template template = serverSolutionTemplates.get(templateName);
+                if (template.getMetadata().getAutoUpdate()
+                        && template.getMetadata().getKeywordsToReplace() == "") {
```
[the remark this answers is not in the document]
jothi-prasad: Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4695,6 +4695,25 @@ void processUserAuthorityRestrictions() {
...
+
+        final String caller = "getDomainNamesFromTemplate";
+        final String auditRef = "";
```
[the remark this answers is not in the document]
jothi-prasad: Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/utils/ZMSUtils.java
```diff
@@ -411,4 +412,17 @@ public static boolean isUserAuthorityFilterValid(Authority userAuthority, final
             return true;
         }
     }
+
+    // To avoid firing multiple queries against DB, this function will generate 1 consolidated query for all domains->templates combination
```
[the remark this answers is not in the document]
jothi-prasad: Moved it to JDBCConnection.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -144,6 +144,7 @@
     protected int domainNameMaxLen;
     protected AuthorizedServices serverAuthorizedServices = null;
     protected SolutionTemplates serverSolutionTemplates = null;
+    protected Map<String, Integer> eligibleTemplatesForAutoUpdate = null;
```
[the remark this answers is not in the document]
jothi-prasad: this is needed because we are populating this map via load solution templates during early stage of app startup and using it in autoApplyTemplate() in the later stages. also we need to load object store in between which is required to called applytemplate function from dbservice. so there is no other way to pass this data down to autoApplyTemplate() function.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -4130,6 +4157,19 @@ private boolean updateRoleMemberNotificationTimestamp(String server, long timest
         return roles;
     }
 
+    // To avoid firing multiple queries against DB, this function will generate 1 consolidated query for all domains->templates combination
+    public String generateDomainTemplateVersionQuery(Map<String, Integer> templateNameAndLatestVersion) {
```
havetisyan, line 4161: this doesn't need to be public and doesn't look we have any unit tests for this method.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -4130,6 +4157,19 @@ private boolean updateRoleMemberNotificationTimestamp(String server, long timest
...
+        StringBuilder query = new StringBuilder("SELECT domain.name, domain_template.template FROM domain_template " +
+                "JOIN domain ON domain_template.domain_id=domain.domain_id WHERE ");
+
+        for (String templateName : templateNameAndLatestVersion.keySet()) {
+            query.append("(domain_template.template = '" + templateName + "' and current_version < " + templateNameAndLatestVersion.get(templateName) + ") OR ");
```
havetisyan, line 4166: why not just call multiple appends instead of adding the strings first and then calling append?

servers/zms/src/test/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnectionTest.java
```diff
@@ -9358,4 +9358,37 @@ public void testListRolesWithUserAuthorityRestrictionsException() throws Excepti
...
+                .thenReturn(domainName); // domain name
+        Mockito.when(mockResultSet.getString("template"))
+                .thenReturn(templateName1); // template name
+        Mockito.when(mockResultSet.getString("name"))
+                .thenReturn(domainName);
+        Mockito.when(mockResultSet.getString("template"))
+                .thenReturn(templateName2);
+
+        Mockito.when(mockResultSet.next())
+                .thenReturn(true).thenReturn(true).thenReturn(false); // this one is for domain id
```
havetisyan, line 9383: not sure what the comment at the end about domain id is referring to

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4695,6 +4695,24 @@ void processUserAuthorityRestrictions() {
...
+        DomainTemplate domainTemplate = new DomainTemplate();
+        try (ObjectStoreConnection con = store.getConnection(true, false)) {
+             domainTemplateListMap = con.getDomainFromTemplateName(templateDetails);
+        }
+
+        for (String domainName : domainTemplateListMap.keySet()) {
+            domainTemplate.setTemplateNames(domainTemplateListMap.get(domainName));
+            //Passing null context since it is an internal call during app start up
+            //executePutDomainTemplate can bulk apply templates given a domain hence sending domainName and templatelist
+            this.executePutDomainTemplate(null, domainName, domainTemplate, auditRef, caller);
```
havetisyan, line 4711: this call should be protected such that if any exception is thrown, we don't stop the process but rather just swallow it and let it continue with the rest of domains.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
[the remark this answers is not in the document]
jothi-prasad: will move the code block and remove the field reference from class.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
jothi-prasad: this code is covered by unit tests in (JDBCConnectionTest->getDomainFromTemplateName) already. i will mark the method as private.

havetisyan: it's not covered by that test. that's just a mock that doesn't look at the query string at all so we never verify that we actually created the correct query string.

jothi-prasad: yes you are right it is just a mock. the code coverage part is taken care of while we run the JDBCconnectiontest but i will add a test to validate the query pattern.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
jothi-prasad: Done.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
jothi-prasad: Done. i added a test for query validation.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -890,6 +894,24 @@ void loadSolutionTemplates() {
...
             serverSolutionTemplates.setTemplates(new HashMap<>());
+        } else {
+            eligibleTemplatesForAutoUpdate = new HashMap<>();
+            for (String templateName : serverSolutionTemplates.getTemplates().keySet()) {
+                Template template = serverSolutionTemplates.get(templateName);
+                if (template.getMetadata().getAutoUpdate()
+                        && template.getMetadata().getKeywordsToReplace().isEmpty()) {
+                    eligibleTemplatesForAutoUpdate.put(templateName, template.getMetadata().getLatestVersion());
+                }
+            }
```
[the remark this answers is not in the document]
jothi-prasad: you are right. this is me being stupid. i forgot the fact that solutionServerTemplate is global. made this change and removed the class level declaration

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
jothi-prasad: yep handled exception case to skip failures and continue.. thank you.

servers/zms/src/test/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnectionTest.java
jothi-prasad: sorry my bad. removed it

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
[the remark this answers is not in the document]
jothi-prasad: > Please also make sure to run openclover at the end and make sure all new lines of code are covered with test cases:
> 
> mvn -P coverage clean org.openclover:clover-maven-plugin:instrument org.openclover:clover-maven-plugin:clover

yes did this and added more tests to cover few missing lines.

PR #1003, opened 2020-06-09
Enabled Email Notifications for ZTS with DynamoDB

OferLevi85:

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/cert/CertRecordStoreConnection.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/cert/CertRecordStoreConnection.java
@@ -73,21 +73,13 @@ public interface CertRecordStoreConnection extends Closeable {
     int deleteExpiredX509CertRecords(int expiryTimeMins);
 
     /**
-     * Update lastNotifiedServer and lastNotifiedTime for certificate that failed to refresh for more than one day.
+     * List all certificates that failed to refresh for more than three days.
      * @param lastNotifiedServer
      * @param lastNotifiedTime
      * @param provider
-     * @return True if at least one certificate record was updated (needs notification to be sent)
+     * @return List of certificates that failed to refresh
      */
-    boolean updateUnrefreshedCertificatesNotificationTimestamp(String lastNotifiedServer,
-                                                               long lastNotifiedTime,
-                                                               String provider);
-
-    /**
-     * List all certificates that failed to refresh and require notifications to be sent
-     * @param lastNotifiedServer
-     * @param lastNotifiedTime
-     * @return List of unrefreshed certificate records that need to be modified
-     */
-    List<X509CertRecord> getNotifyUnrefreshedCertificates(String lastNotifiedServer, long lastNotifiedTime);
+    List<X509CertRecord> updateUnrefreshedCertificatesNotificationTimestamp(String lastNotifiedServer,
+                                                                            long lastNotifiedTime,
+                                                                            String provider);
 }
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStore.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStore.java
@@ -32,18 +32,20 @@ public class DynamoDBCertRecordStore implements CertRecordStore {
     private static final Logger LOGGER = LoggerFactory.getLogger(DynamoDBCertRecordStore.class);
     private static final Logger CERTLOGGER = LoggerFactory.getLogger("X509CertLogger");
 
-    private DynamoDB dynamoDB;
     private String tableName;
+    private String indexName;
+    private DynamoDB dynamoDB;
 
-    public DynamoDBCertRecordStore(AmazonDynamoDB client, final String tableName) {
-        dynamoDB = new DynamoDB(client);
+    public DynamoDBCertRecordStore(AmazonDynamoDB client, final String tableName, String indexName) {
+        this.dynamoDB = new DynamoDB(client);
         this.tableName = tableName;
+        this.indexName = indexName;
     }
 
     @Override
     public CertRecordStoreConnection getConnection() {
         try {
-            return new DynamoDBCertRecordStoreConnection(dynamoDB, tableName);
+            return new DynamoDBCertRecordStoreConnection(dynamoDB, tableName, indexName);
         } catch (Exception ex) {
             LOGGER.error("getConnection: {}", ex.getMessage());
             throw new ResourceException(ResourceException.SERVICE_UNAVAILABLE, ex.getMessage());
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
@@ -15,22 +15,22 @@
  */
 package com.yahoo.athenz.zts.cert.impl;
 
-import java.util.ArrayList;
-import java.util.Date;
-import java.util.List;
+import java.util.*;
+import java.util.concurrent.TimeUnit;
 
+import com.amazonaws.services.dynamodbv2.document.*;
+import com.amazonaws.services.dynamodbv2.document.spec.QuerySpec;
+import com.amazonaws.services.dynamodbv2.document.utils.ValueMap;
+import com.amazonaws.services.dynamodbv2.model.ReturnValue;
 import com.yahoo.athenz.common.server.cert.CertRecordStoreConnection;
 import com.yahoo.athenz.common.server.cert.X509CertRecord;
 import com.yahoo.athenz.zts.ZTSConsts;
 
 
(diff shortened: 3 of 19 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
```diff
@@ -132,7 +132,7 @@ public CertFailedRefreshNotificationTask(InstanceCertManager instanceCertManager
 
     private List<X509CertRecord> getRecordsWithValidHosts(List<X509CertRecord> unrefreshedCerts) {
         return unrefreshedCerts.stream()
-                    .filter(record -> hostnameResolver.isValidHostname(record.getHostName()))
+                    .filter(record -> hostnameResolver == null || hostnameResolver.isValidHostname(record.getHostName()))
```
OferLevi85, line 135: Note that I've changed this check to only verify hostname only If hostnameResolver is used.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
```diff
@@ -54,17 +59,25 @@
     private static final String KEY_EXPIRY_TIME = "expiryTime";
     private static final String KEY_HOSTNAME = "hostName";
     private static final String KEY_TTL = "ttl";
+    public static final String INDEX_CURRENT_DATE = "currentDate-index";
```
OferLevi85, line 62: Note that we need to create a new index in prod and any other environment before pushing these changes

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
```diff
@@ -213,17 +231,92 @@ public int deleteExpiredX509CertRecords(int expiryTimeMins) {
...
     public List<X509CertRecord> getNotifyUnrefreshedCertificates(String lastNotifiedServer, long lastNotifiedTime) {
-        // Currently unimplemented for AWS
-        return new ArrayList<>();
+
+        Map<String, AttributeValue> expressionAttributeValues = new HashMap<>();
+        expressionAttributeValues.put(":lastNotifiedServerVal", new AttributeValue().withS(lastNotifiedServer));
+        expressionAttributeValues.put(":lastNotifiedTimeVal", new AttributeValue().withN(Long.toString(lastNotifiedTime)));
+
+        // Get all records that that should receive notifications.
+        ScanRequest scanRequest = new ScanRequest()
```
OferLevi85, line 253: Regarding the scan, please note the following:
1. Couldn't avoid the scan when getting the updated records as I must use consistent reading. Indexes only support eventual consistency so there's a good chance we'll miss records if we use it.
2. We will need to enable Scan DynamoDB for the IAM role in prod and any other environments in order for scanning to work.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
```diff
@@ -139,10 +156,10 @@ public boolean updateX509CertRecord(X509CertRecord certRecord) {
                             new AttributeUpdate(KEY_CURRENT_SERIAL).put(certRecord.getCurrentSerial()),
                             new AttributeUpdate(KEY_CURRENT_IP).put(certRecord.getCurrentIP()),
                             new AttributeUpdate(KEY_CURRENT_TIME).put(getLongFromDate(certRecord.getCurrentTime())),
+                            new AttributeUpdate(KEY_CURRENT_DATE).put(dynamoDBUtils.getIso8601FromDate(certRecord.getCurrentTime())),
```
OferLevi85, line 159: Everytime we update the currentTime column (hold epoch in long format), we should also update the new currentDate column (ISO date in the format YYYY-MM-DD).

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
OferLevi85: We actually make an update just moments when writing the lastNotifiedTime and server.
In InstanceCertManager.getUnrefreshedCertsNotifications we call 2 functions right after the other:
1. updateUnrefreshedCertificatesNotificationTimestamp - we quey the index and update the records with the lastNotifiedTime and server
2. Make the call to this function, getNotifyUnrefreshedCertificates to get all the records that were just updated by us.
If we update the records and don't read them with consistency then we'll miss these records and skip their notifications.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
```diff
@@ -139,10 +156,10 @@ public boolean updateX509CertRecord(X509CertRecord certRecord) {
...
                             new AttributeUpdate(KEY_PREV_SERIAL).put(certRecord.getPrevSerial()),
                             new AttributeUpdate(KEY_PREV_IP).put(certRecord.getPrevIP()),
                             new AttributeUpdate(KEY_PREV_TIME).put(getLongFromDate(certRecord.getPrevTime())),
-                            new AttributeUpdate(KEY_CLIENT_CERT).put(certRecord.getClientCert()),
```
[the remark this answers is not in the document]
OferLevi85: Yeah sorry I put the explanation in the closed PR and forgot to copy it here.
In any case, I noticed that in JDBC connection we do not update this column during an update.
After checking the purpose of this attribute (differentiating between server and client certificates), I assumed that it is a column that should never be changed after the record is first inserted.
If my assumption is incorrect let me know (in which case we should also add it to the JDBC update call).

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
OferLevi85: Edit: There is actually no good reason for InstanceCertManager.getUnrefreshedCertsNotifications to make separate calls.
We can make a single call to the store connection and have it responsible for getting all records that should receive notifications. 
For JDBC we can leave the implementation as is (it will just do the calls one after the other) but for DynamoDB we can just return all the updated records from the query. I'll fix it and push

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/cert/CertRecordStoreConnection.java
```diff
@@ -73,21 +73,13 @@
     int deleteExpiredX509CertRecords(int expiryTimeMins);
 
     /**
-     * Update lastNotifiedServer and lastNotifiedTime for certificate that failed to refresh for more than one day.
+     * List all certificates that failed to refresh for more than one day.
      * @param lastNotifiedServer
      * @param lastNotifiedTime
      * @param provider
      * @return True if at least one certificate record was updated (needs notification to be sent)
```
havetisyan, line 80: 1) by default we were list certs that were not refreshed for more than 3 days (the default value in jdbc) 
2) let's update the return description since we're returning List now.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStore.java
```diff
@@ -32,18 +33,18 @@
...
-        dynamoDB = new DynamoDB(client);
+        this.dynamoDB = new DynamoDB(client);
         this.tableName = tableName;
     }
 
     @Override
     public CertRecordStoreConnection getConnection() {
         try {
-            return new DynamoDBCertRecordStoreConnection(dynamoDB, tableName);
+            return new DynamoDBCertRecordStoreConnection(dynamoDB, tableName, new DynamoDBUtils());
```
havetisyan, line 47: what are we creating a new instance of DynamoDBUtils and passing here? Why can't DynamoDBCertRecordStoreConnection just maintain it's own instance of the class since it only has a couple of public methods? Any reason why those methods can't be static and then no need to create any instances?

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
havetisyan: In the conf/zts.properties for the athenz.zts.cert_dynamodb_table_name setting I have small section what type of table we need to create. can you also update that section to include the requirement for this index as well.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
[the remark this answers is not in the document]
havetisyan: Yes, we should update it and fix jdbc as well. The logic is that once you get a client cert, you can't request a service cert. However, the provider may decide that it wants ZTS to issue a client cert only for a request in which case we honor the downgrade option.

PR #1005, opened 2020-06-12
adding additional role meta attributes via templates

jothi-prasad:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
@@ -2820,7 +2820,20 @@ public class DBService {
         }
         Role templateRole = new Role()
                 .setName(templateRoleName)
-                .setTrust(role.getTrust());
+                .setTrust(role.getTrust())
+                //adding additional role meta attributes if present in template->roles
+                .setCertExpiryMins(role.getCertExpiryMins())
+                .setSelfServe(role.getSelfServe())
+                .setMemberExpiryDays(role.getMemberExpiryDays())
+                .setTokenExpiryMins(role.getTokenExpiryMins())
+                .setSignAlgorithm(role.getSignAlgorithm())
+                .setServiceExpiryDays(role.getServiceExpiryDays())
+                .setMemberReviewDays(role.getMemberReviewDays())
+                .setServiceReviewDays(role.getServiceReviewDays())
+                .setReviewEnabled(role.getReviewEnabled())
+                .setNotifyRoles(role.getNotifyRoles())
+                .setUserAuthorityFilter(role.getUserAuthorityFilter())
+                .setUserAuthorityExpiration(role.getUserAuthorityExpiration());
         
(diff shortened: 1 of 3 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2820,7 +2820,20 @@ Role updateTemplateRole(Role role, String domainName, List<TemplateParam> params
         }
         Role templateRole = new Role()
```
havetisyan, line 2821: The change itself is not sufficient since it's possible that the user has made some changes already and thus we don't want to lose any of those changes. So I want to make our logic a little more smarter.

When applying the template and updating the role, we should read the original values and only override those that have been specified in the template. (let's have a unit test for this case as well).

servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
```diff
@@ -7012,4 +7012,72 @@ public void testAutoApplySolutionTemplate() {
...
+        // verify that our role collection includes the expected roles
+
+        List<String> names = zms.dbService.listRoles(domainName);
+        assertEquals(2, names.size());
+        assertTrue(names.contains("admin"));
+        assertTrue(names.contains("vip_admin"));
+
+        Role role = zms.dbService.getRole(domainName, "vip_admin", false, false, false);
+        assertEquals(domainName + ":role.vip_admin", role.getName());
+        assertNull(role.getTrust());
```
havetisyan, line 7044: in our test, we should verify that all attributes have been set that are defined in the template to make sure all attributes are correctly processed.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
jothi-prasad: Done. Added logic to merge existing rolemeta from DB with new rolemeta values form template.
added a unit test to test this scenario too.

servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
jothi-prasad: Done

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -509,6 +509,7 @@ boolean processRole(ObjectStoreConnection con, Role originalRole, String domainN
         } else {
             // carrying over auditEnabled from original role
             role.setAuditEnabled(originalRole.getAuditEnabled());
+            mergeOriginalRoleAndMetaRoleAttributes(originalRole, role);
```
havetisyan, line 512: we cannot change the behavior of the processRole - that method is used in lots of places.
this call must be done only in the method where templates are being handled.

jothi-prasad: Done.

PR #1012, opened 2020-06-18
Zts health check

OferLevi85:


servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBStatusChecker.java
```diff
@@ -0,0 +1,77 @@
...
+        } catch (Throwable ex) {
+            throw new StatusCheckException(ex);
+        } finally {
+            // Close resources
+            if (clientAndCreds != null) {
+                try {
+                    if (clientAndCreds.getAmazonDynamoDB() != null) {
+                        clientAndCreds.getAmazonDynamoDB().shutdown();
+                    }
+                    if (clientAndCreds.getAwsCredentialsProvider() != null) {
```
OferLevi85, line 64: The reason I had to return the credentialProvider from dynamoDBClientFetcher is to be able to close it here. I didn't want to wait for garbage collector to remove everything after the dynamoDB client shuts down (and there is no way to close it from dynamoDB client as AWSCredentialsProvider interface isn't closeable).

PR #1014, opened 2020-06-19
ZTS provider: validate hostname attribute based on incoming source IP

psasidhar:

### DIFF
--- a/libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceZTSProvider.java
+++ b/libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceZTSProvider.java
@@ -18,6 +18,7 @@ package com.yahoo.athenz.instance.provider.impl;
 import com.yahoo.athenz.auth.KeyStore;
 import com.yahoo.athenz.auth.token.PrincipalToken;
 import com.yahoo.athenz.auth.token.Token;
+import com.yahoo.athenz.common.server.dns.HostnameResolver;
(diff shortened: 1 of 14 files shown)


libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceZTSProvider.java
```diff
@@ -127,24 +157,113 @@ public InstanceConfirmation confirmInstance(InstanceConfirmation confirmation) {
...
+
+        if (sanUri == null || sanUri.isEmpty()) {
+            LOGGER.info("Request contains no sanURI to verify");
+            return true;
+        }
+
+        for (String uri: sanUri.split(",")) {
+            int idx = uri.indexOf(URI_HOSTNAME_PREFIX);
+            if (idx != -1) {
+                if (!uri.substring(idx + 18).equals(hostname)) {
```
abvaidya, line 256: We should use URI_HOSTNAME_PREFIX.length() instead of hardcoded 18

psasidhar: and that gets calculated on each api call, and for each uri. we are hardcoding known lengths in other places.

psasidhar: ZTSProvider itself expected to be less frequently used. So, changed it to length().

PR #1019, opened 2020-06-24
go cli changes to include meta data fields.

jothi-prasad:
go cli changes to include meta data fields for show-server-template command

### DIFF
--- a/libs/go/zmscli/template.go
+++ b/libs/go/zmscli/template.go
@@ -46,7 +46,7 @@ func (cli Zms) ShowServerTemplate(templateName string) (*string, error) {
 	}
 	var buf bytes.Buffer
 	buf.WriteString("template:\n")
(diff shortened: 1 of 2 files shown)


libs/go/zmscli/template.go
```diff
@@ -46,6 +46,7 @@ func (cli Zms) ShowServerTemplate(templateName string) (*string, error) {
 	}
 	var buf bytes.Buffer
 	buf.WriteString("template:\n")
+	cli.dumpMetadata(&buf, template.Metadata ,indent_level2_dash, templateName)
```
havetisyan, line 49: fix the comma before and after the space for template.Metadata

gofmt -d template.go

will show all the formatting errors. not sure why this is not being flagged during the build since we do call gofmt.

jothi-prasad: Thank you. Done.

PR #1028, opened 2020-07-06
ZTS Client to read access tokens from file system

MartinTrojans:


clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
```diff
@@ -1168,31 +1192,41 @@ public AccessTokenResponse getAccessToken(String domainName, List<String> roleNa
             }
         }
 
+        // if no hit then we need to look up in disk
+        try {
+            accessTokenResponse = ztsFileUtil.lookupAccessTokenFromDisk(domainName, roleNames);
+        } catch (IOException e) {
+            LOG.warn("GetAccessToken: failed to load access token from disk ", e.getMessage());
```
havetisyan, line 1199: this should be an error if file based access tokens is enabled and we're not able to read them from disk.

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+    // function load the access token from file
+    public AccessTokenResponse lookupAccessTokenFromDisk(String domain, List<String> rolesName) throws IOException {
+        String path = System.getProperty(ACCESS_TOKEN_PATH_PROPERTY, DEFAULT_ACCESS_TOKEN_DIR_PATH);
+        LOG.debug("Trying to fetch access token from disk for domain: {}, roleNames: {}", domain, rolesName);
+        String rolesStr = getRolesStr(domain, rolesName);
+        String fileName = roleNameMap.get(rolesStr);
+        if (fileName == null) {
+            return null;
+        }
+        File tokenFile = new File(path + "/" + domain + "/" + fileName);
```
havetisyan, line 76: we should use pathSeparator instead of "/"

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                }
+            }
+        }
+
+    }
+
+    // function load the access token from file
+    public AccessTokenResponse lookupAccessTokenFromDisk(String domain, List<String> rolesName) throws IOException {
+        String path = System.getProperty(ACCESS_TOKEN_PATH_PROPERTY, DEFAULT_ACCESS_TOKEN_DIR_PATH);
```
havetisyan, line 69: make this and other String variables as final as necessary.

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+        LOG.debug("Trying to fetch access token from disk for domain: {}, roleNames: {}", domain, rolesName);
+        String rolesStr = getRolesStr(domain, rolesName);
+        String fileName = roleNameMap.get(rolesStr);
+        if (fileName == null) {
+            return null;
+        }
+        File tokenFile = new File(path + "/" + domain + "/" + fileName);
+
+        AccessTokenResponse accessTokenResponse = objectMapper.readValue(tokenFile, AccessTokenResponse.class);
```
havetisyan, line 78: no need to create this object and and instead we can just have:

return objectMapper.readValue(tokenFile, AccessTokenResponse.class);

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+            //if no role name specific, should return all roles
+            return "*";
+        }
+        Collections.sort(roleNames);
+        return domain + ROLE_NAME_CONNECTOR + String.join(ROLE_NAME_CONNECTOR, roleNames);
+    }
+
+    private void addToRoleMap(String domain, String fileName, AccessTokenResponse accessTokenResponse) throws IOException {
+        // parse roles from access token
+        String token = accessTokenResponse.getAccess_token();
```
havetisyan, line 94: this should be final

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+
+        try {
+            AccessToken accessToken = new AccessToken(token, accessSignKeyResolver);
+            List<String> roleNames = accessToken.getScope();
+            roleNameMap.put(getRolesStr(domain, roleNames), fileName);
+        } catch (Exception e) {
+            LOG.warn("Got error to parse access token file {}, error: {}", fileName, e.getMessage());
```
havetisyan, line 101: any exceptions should be logger as error

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+    public ZTSFileUtil(JwtsSigningKeyResolver resolver) {
+        roleNameMap = new HashMap<>();
+        accessSignKeyResolver = resolver;
+    }
+
+    public void preload() {
+        String path = System.getProperty(ACCESS_TOKEN_PATH_PROPERTY, DEFAULT_ACCESS_TOKEN_DIR_PATH);
+        File dir = new File(path);
+
+        // preload the map from the domain, rolesname to file path
```
havetisyan, line 33: we should add more comments here as far what is the expected directory structure otherwise it's not easy to follow why we're keep traversing, etc

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+        if (dir.exists() && dir.isDirectory()) {
+            for (File domainDir: dir.listFiles()) {
+                if (domainDir.isDirectory()) {
+                    for (File tokenFile: domainDir.listFiles()) {
+                        if (!tokenFile.isDirectory()) {
+                            AccessTokenResponse accessTokenResponse = null;
+                            try {
+                                accessTokenResponse = objectMapper.readValue(tokenFile, AccessTokenResponse.class);
+                            } catch (IOException e) {
+                                LOG.warn("Failed to load or parse token file: {}", tokenFile);
```
havetisyan, line 43: this should also be an error

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+                                continue;
+                            }
+
+                            AccessTokenResponseCacheEntry cacheEntry = new AccessTokenResponseCacheEntry(accessTokenResponse);
+
+                            if (!cacheEntry.isExpired(-1)) {
+                                try {
+                                    addToRoleMap(domainDir.getName(), tokenFile.getName(), accessTokenResponse);
+                                } catch (IOException e) {
+                                    LOG.warn("Failed to put role map for token: {}", tokenFile.getName());
```
havetisyan, line 56: here as well let's make this an error

clients/java/zts/core/src/test/java/com/yahoo/athenz/zts/ZTSClientTest.java
```diff
@@ -3014,6 +3019,31 @@ public void testGetAccessTokenCacheKeySSLContext() {
         client.close();
     }
 
+    @Test
+    public void testGetAccessTokenFromFile() {
```
havetisyan, line 3023: we should add test cases where we have invalid access tokens on disk and make sure those are correctly handled and skipped by the client

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
```diff
@@ -327,6 +334,23 @@ public static void lookupZTSUrl() {
             }
         }
     }
+
+    public static void initZtsFileUtil() {
+        ztsFileUtil = new ZTSFileUtil(getAccessTokenSignKeyResolver());
+        ztsFileUtil.preload();
+    }
+
+    public static void setAccessTokenSignKeyResolver(JwtsSigningKeyResolver jwtsSigningKeyResolver) {
```
havetisyan, line 343: I'm not sure why we're only setting the resolver here and not doing anything else since it means after I do this I also need to call initZtsFileUtil to set the resolver and call preload. So why just within this call we reload the file util object ourselves instead of asking the user to do it.

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
```diff
@@ -327,6 +334,23 @@ public static void lookupZTSUrl() {
...
+        resolver = jwtsSigningKeyResolver;
+    }
+
+    public static JwtsSigningKeyResolver getAccessTokenSignKeyResolver() {
```
havetisyan, line 347: based on the previous comment for the setAccessTokenSignKeyResolver I don't think we need this function at all.

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
```diff
@@ -172,6 +174,7 @@
     private static ServiceLoader<ZTSClientService> ztsTokenProviders;
     private static AtomicReference<Set<String>> svcLoaderCacheKeys;
     private static PrivateKeyStore PRIVATE_KEY_STORE = loadServicePrivateKey();
+    private static ZTSFileUtil ztsFileUtil;
```
havetisyan, line 177: Let's call it more specific ZTSAccessTokenFileLoader instead of ZTSFileUtil

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
```diff
@@ -0,0 +1,105 @@
...
+import java.io.File;
+import java.io.IOException;
+import java.util.*;
+
+public class ZTSFileUtil {
+
+    private static final Logger LOG = LoggerFactory.getLogger(ZTSFileUtil.class);
+
+    static private final String DEFAULT_ACCESS_TOKEN_DIR_PATH = "/var/lib/sia/tokens/";
+    static private final String ACCESS_TOKEN_PATH_PROPERTY = "zts.client.accesstoken.path";
```
havetisyan, line 18: all of our property names start with athenz.

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
MartinTrojans: Having setAccessTokenSignKeyResolver and getAccessTokenSignKeyResolver is for setting mock resolver in unit test

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
MartinTrojans: Having setAccessTokenSignKeyResolver and getAccessTokenSignKeyResolver is for setting mock resolver in unit test

clients/java/zts/core/src/test/java/com/yahoo/athenz/zts/ZTSClientTest.java
MartinTrojans: Added invalid access token

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSFileUtil.java
MartinTrojans: updated

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
havetisyan: but we still need to make them usable since they're declared as public. so rather than just having a resolver kept as a member variable, we can call the init method again to reinitialize the file utility and you don't need the getAccessTokenSignKeyResolver() at all.

This also gives the client the flexibility for someone who doesn't want to use the conf file but define their own resolver, so that they can do so and pass it to the client.

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
havetisyan: the unit tests can only be done with the set method . no need for this get method.

PR #1031, opened 2020-07-08
Members of roles with filters that don't match their attributes will …

OferLevi85:
…appear disabled

### DIFF
--- a/ui/src/__tests__/components/role/__snapshots__/RoleMember.test.js.snap
+++ b/ui/src/__tests__/components/role/__snapshots__/RoleMember.test.js.snap
@@ -83,7 +83,6 @@ exports[`RoleMember should render 1`] = `
 }
 
 .emotion-0 {
-  color: #3570f4;
   -webkit-text-decoration: none;
   text-decoration: none;
   cursor: pointer;
@@ -95,6 +94,7 @@ exports[`RoleMember should render 1`] = `
 >
   <a
     class="emotion-0"
+    style="color: rgb(53, 112, 244);"
   >
      
     user.test2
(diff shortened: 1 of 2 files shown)


ui/src/components/role/RoleMember.js
```diff
@@ -109,8 +122,13 @@ export default class RoleMember extends React.Component {
                     >
                         {' '}
                         {review && (
-                            <StyledTagFullNameExpiry>
-                                {' | '}
```
OferLevi85, line 113: Removed this ' | ' on purpose. Not related to my fix but there was a separator accidentally placed at the beginning of the name before the reminder icon (for pending members only). 
Removed it so it will match the way role members look in the active members' list.

ui/src/__tests__/components/role/__snapshots__/RoleMember.test.js.snap
```diff
@@ -83,7 +83,6 @@ exports[`RoleMember should render 1`] = `
 }
 
 .emotion-0 {
-  color: #3570f4;
```
OferLevi85, line 86: Note that rgb(53, 112, 244) that was added is equal to #3570f4 that was removed. The snapshot style didn't actually change...

ui/src/components/role/RoleMember.js
```diff
@@ -46,11 +45,13 @@ const MenuDiv = styled.div`
 `;
 
 const StyledAnchor = styled.a`
-    color: #3570f4;
     text-decoration: none;
     cursor: pointer;
 `;
 
+const StyledAnchorActiveInline = { color: '#3570f4' };
```
pratikgote, line 52: can we move these to constants file ?

PR #1034, opened 2020-07-09
Added query to get all roles for principal

OferLevi85:


core/zms/src/main/rdl/Role.rdli
```diff
@@ -129,6 +129,19 @@ resource DomainRoleMembers GET "/domain/{domainName}/member" {
     }
 }
 
+// Fetch all the roles across domains by either calling or specified principal
+resource DomainRoleMember GET "/all_roles?principal={principal}" (name=GetAllRoles) {
```
havetisyan, line 133: can we also extend this to include an optional domainName attribute where I would only return the role names from the given domain name if specified. We don't have an API that just returns the role names that a given principal is member  in a domain so we don't want them to get the full list and then parse and filter out the domain they're interested in.

OferLevi85: Sure thing.

OferLevi85: Done

clients/java/zms/pom.xml
```diff
@@ -28,6 +28,7 @@
 
   <properties>
     <code.coverage.min>0.42</code.coverage.min>
+    <code.coverage.min>0.95</code.coverage.min>
```
havetisyan, line 31: we should remove the previous line

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
```diff
@@ -1639,6 +1640,51 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+            List<RoleMember> roleMembers = role.getRoleMembers();
+            if (roleMembers == null) {
+                continue;
+            }
+            for (RoleMember roleMember : roleMembers) {
+                if (!roleMember.getMemberName().equals(principal)) {
+                    continue;
+                }
+
+                MemberRole memberRole = new MemberRole();
```
havetisyan, line 1677: we should also return the system disabled flag if we're return all the other attributes

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -3751,6 +3765,65 @@ public DomainRoleMembers listDomainRoleMembers(String domainName) {
...
+        roleMember.setMemberName(principal);
+        if (StringUtil.isEmpty(domainName)) {
+            try (PreparedStatement ps = con.prepareStatement(SQL_GET_ALL_ROLES)) {
+                ps.setInt(1, principalId);
+                return getRolesForPrincipal(caller, roleMember, ps);
+            } catch (SQLException ex) {
+                throw sqlError(ex, caller);
+            }
+        } else {
+            int domainId = getDomainId(domainName);
```
havetisyan, line 3788: we should. handle the case where we have an invalid domain:

        if (domainId == 0) {
            throw notFoundError(caller, ZMSConsts.OBJECT_DOMAIN, domainName);
        }

and we should have a unit test case for this as well.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2956,6 +2957,34 @@ public DomainRoleMembers getDomainRoleMembers(ResourceContext ctx, String domain
...
+        final String caller = "getallroles";
+        metric.increment(ZMSConsts.HTTP_GET);
+        logPrincipal(context);
+
+        if (StringUtil.isEmpty(principal)) {
+            // If principal not specified, get roles for current user
+            principal = ((RsrcCtxWrapper) context).principal().getFullName();
+        }
+        validateRequest(context.request(), caller);
+        validate(principal, TYPE_ENTITY_NAME, caller);
```
havetisyan, line 2971: if domainName is specified, we need to validate its value as well

core/zms/src/main/rdl/Role.rdli
```diff
@@ -129,6 +129,20 @@ resource DomainRoleMembers GET "/domain/{domainName}/member" {
     }
 }
 
+// Fetch all the roles across domains by either calling or specified principal
+resource DomainRoleMember GET "/all_roles?principal={principal}&domain={domainName}" (name=GetAllRoles) {
```
havetisyan, line 133: for the uri should be consistent with others and use the first component as the object type, e.g. /domain, /resource,e tc.

we should just call it /role instead of /all_roles and instead of getAllRoles, let's call the API as getPrincipalRoles

clients/java/zms/pom.xml
OferLevi85: Done

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
OferLevi85: Done

servers/zms/src/main/java/com/yahoo/athenz/zms/store/file/FileConnection.java
OferLevi85: Done

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
OferLevi85: Done

core/zms/src/main/rdl/Role.rdli
OferLevi85: Done

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2956,6 +2957,38 @@ public DomainRoleMembers getDomainRoleMembers(ResourceContext ctx, String domain
...
+
+        if (!StringUtil.isEmpty(domainName)) {
+            validate(domainName, TYPE_DOMAIN_NAME, caller);
```
havetisyan, line 2974: we should also lowercase the domainname here

OferLevi85: Fixed

PR #1042, opened 2020-07-14

OferLevi85:


libs/java/server_common/pom.xml
```diff
@@ -172,6 +184,20 @@
       <artifactId>athenz-zms-core</artifactId>
       <version>${project.parent.version}</version>
     </dependency>
+      <dependency>
+          <groupId>org.eclipse.jetty</groupId>
```
OferLevi85, line 188: Added missing dependencies from zts

libs/java/server_common/src/main/java/com/yahoo/athenz/common/ServerCommonConsts.java
```diff
@@ -23,6 +23,28 @@
...
+    public static final String ZTS_PROP_AWS_CREDS_INVALID_CACHE_TIMEOUT     = "athenz.zts.aws_creds_invalid_cache_timeout";
+    public static final String ZTS_PROP_AWS_ENABLED                         = "athenz.zts.aws_enabled";
+    public static final String ZTS_PROP_AWS_CREDS_UPDATE_TIMEOUT            = "athenz.zts.aws_creds_update_timeout";
+    public static final String ZTS_SSH_TYPE                                 = "certtype";
+
+    // For S3ChangeLogStore
+    public static final String ZTS_PROP_AWS_BUCKET_NAME = "athenz.zts.aws_bucket_name";
+
+    // For ZMSFileChangeLogStore
+    public static final String ZTS_PROP_ZMS_URL_OVERRIDE    = "athenz.zts.zms_url";
```
OferLevi85, line 38: Moved ZTS consts here and fixed the references (which is the reason many files have changed)

libs/java/server_common/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
OferLevi85, line 1: I had to add this file (it is also located in the same directory in zts). It is used to tell mockito that it can mock final methods (without it some tests failed).

libs/java/server_common/src/test/java/com/yahoo/athenz/common/server/store/MockCloudStore.java
```diff
@@ -0,0 +1,104 @@
+/*
+ *  Copyright 2020 Verizon Media
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
```
OferLevi85, line 4: I had to copy this file and the other "MockChangeLog" classes from ZTS as they are shared between tests in ZTS and in Common (and test classes are not included in the common jar). 
There are 2 ways to overcome this code duplication:
1. Move these classes outside of test directory
2. Add a "packaging" step for tests in common that will create a jar that includes tests. We can then make ZTS and ZMS dependent on that jar (see https://dzone.com/articles/sharing-test-classes-between-multiple-modules-in-a).
For now, I thought it's best to leave it as is. We can fix it in a later PR if needed.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSConsts.java
```diff
@@ -198,7 +183,7 @@
     public static final String ZTS_PROP_SSH_RECORD_STORE_FACTORY_CLASS   = "athenz.zts.ssh_record_store_factory_class";
 
     public static final String ZTS_METRIC_FACTORY_CLASS            = "com.yahoo.athenz.common.metrics.impl.NoOpMetricFactory";
-    public static final String ZTS_CHANGE_LOG_STORE_FACTORY_CLASS  = "com.yahoo.athenz.zts.store.impl.ZMSFileChangeLogStoreFactory";
+    public static final String ZTS_CHANGE_LOG_STORE_FACTORY_CLASS  = "com.yahoo.athenz.common.server.store.impl.ZMSFileChangeLogStoreFactory";
```
OferLevi85, line 186: Changed the default class location

havetisyan:
@OferLevi85 looks like we're moving more stuff into common that I was expecting - specifically that Cloud Store stuff. I don't think that should be part of the server common. The only place we're using that is to get an S3 client. So why not just remove the CloudStore from the interface and create an S3 client object in the S3ChangeLogFactory class and pass it to the S3ChangeLog implementation. Let's talk about this tomorrow.

PR #1045, opened 2020-07-15
Verifying URI fields for service/role cert

psasidhar:


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2191,6 +2191,18 @@ boolean validateRoleCertificateRequest(final String csr, final String domainName
...
+        // validate uriHostname in request matches the uriHostname in cert
+        if (!validateUriHostname(certReq.getUriHostname(), cert)) {
+            return false;
+        }
+
+        // validate InstanceID in request matches the instanceid in the cert
+        if (!validateInstanceId(certReq.getInstanceId(), cert)) {
+            // Todo: minitor the logs, and make this a hard failure in later iteration
+            LOGGER.error("unable to match request id: {} with cert id: {}", certReq.getInstanceId(), X509CertUtils.extractRequestInstanceId(cert));
+            return true;
```
havetisyan, line 2203: we should not return true here. otherwise we're skipping the last check.

psasidhar: done

PR #1046, opened 2020-07-16

OferLevi85:


libs/java/server_common/pom.xml
```diff
@@ -172,6 +184,20 @@
       <artifactId>athenz-zms-core</artifactId>
       <version>${project.parent.version}</version>
     </dependency>
+      <dependency>
+          <groupId>org.eclipse.jetty</groupId>
+          <artifactId>jetty-client</artifactId>
```
OferLevi85, line 189: Added missing dependencies required for ChangeLogStore

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/AWSCredentialsRefresher.java
```diff
@@ -0,0 +1,111 @@
...
+
+import java.io.Closeable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import static com.yahoo.athenz.common.ServerCommonConsts.ZTS_PROP_AWS_CREDS_UPDATE_TIMEOUT;
+import static com.yahoo.athenz.common.ServerCommonConsts.ZTS_PROP_AWS_REGION_NAME;
+
+public class AWSCredentialsRefresher implements Closeable {
```
OferLevi85, line 34: Extracted this class from CloudStore. It gets credentials and region from AWSInstanceMetadataFetcher (also a new class) and has a periodic task that refreshes the credentials.
Note that in the CloudStore, this task was responsible for 2 things:
1. Refresh the credentials
2. Clear the CloudStore Credentials Cache.

As they are now separated, there are 2 unrelated periodic tasks - one here to refresh the credentials and the other remained in the CloudStore

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/AWSInstanceMetadataFetcher.java
```diff
@@ -0,0 +1,263 @@
...
+public class AWSInstanceMetadataFetcher implements Closeable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AWSInstanceMetadataFetcher.class);
+
+    String awsRole = null;
+
+    String awsRegion;
+    private HttpClient httpClient;
+
+    public AWSInstanceMetadataFetcher() {
+        httpClient = new HttpClient();
```
OferLevi85, line 42: New class extracted from CloudStore. Gets the AWS Instance metadata for AWSCredentialsRefresher that provide credentials for CloudStore and for the S3ClientFactory (that creates S3Clients for the S3ChangeLogStore)

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/S3ClientFactory.java
```diff
@@ -0,0 +1,56 @@
...
+    public S3ClientFactory() {
+        // Instantiate credentials fetcher
+        awsCredentialsRefresher = new AWSCredentialsRefresher();
+    }
+
+    public S3ClientFactory(AWSCredentialsRefresher awsCredentialsRefresher) {
+        this.awsCredentialsRefresher = awsCredentialsRefresher;
+    }
+
+    public AmazonS3 create() {
```
OferLevi85, line 38: This factory replaces the CloudStore dependency in the S3ChangeLog. It provides clients with credentials taken from AWSCredentialsRefresher

libs/java/server_common/src/test/java/com/yahoo/athenz/common/server/store/AWSCredentialsRefresherTest.java
```diff
@@ -0,0 +1,119 @@
...
+
+import static com.yahoo.athenz.common.server.store.AWSInstanceMetadataFetcherTest.AWS_IAM_ROLE_INFO;
+import static com.yahoo.athenz.common.server.store.AWSInstanceMetadataFetcherTest.AWS_INSTANCE_DOCUMENT;
+import static org.testng.Assert.assertEquals;
+import static org.testng.Assert.fail;
+
+public class AWSCredentialsRefresherTest {
+    @Test
+    public void testAWSCredentialsUpdaterExceptions () {
+
```
OferLevi85, line 37: Extracted tests from CloudStoreTest covering the new classes (AWSCredentialsRefresherTes, AWSInstanceMetadataFetcherTest and S3ClientFactoryTest)

servers/zts/src/main/java/com/yahoo/athenz/zts/store/CloudStore.java
```diff
@@ -147,231 +108,23 @@ void initializeAwsSupport() {
...
-                    path, response.getStatus());
-            return null;
-        }
-
-        return data;
-    }
-
-    AssumeRoleRequest getAssumeRoleRequest(String account, String roleName, String principal,
-            Integer durationSeconds, String externalId) {
+    AssumeRoleRequest getAssumeRoleRequest(String account, String roleName,
```
OferLevi85, line 126: Removed the principal argument - wasn't used

libs/java/server_common/src/main/java/com/yahoo/athenz/common/ServerCommonConsts.java
```diff
@@ -23,6 +23,24 @@
...
+    // For ZMSFileChangeLogStore
+    public static final String ZTS_PROP_ZMS_URL_OVERRIDE    = "athenz.zts.zms_url";
+    public static final String ATHENZ_SYS_DOMAIN            = "sys.auth";
+    public static final String ZTS_SERVICE                  = "zts";
+    public static final String PROP_USER_DOMAIN             = "athenz.user_domain";
+    public static final String PROP_ATHENZ_CONF             = "athenz.athenz_conf";
+    public static final String ZTS_PROP_FILE_NAME           = "athenz.zts.prop_file";
+
+    // for tests
+    public static final String ZTS_PROP_AWS_PUBLIC_CERT = "athenz.zts.aws_public_cert";
```
havetisyan, line 42: I don't think this should be here. The public cert is only used for validating instance documents so it belongs to ZTS server only.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/AWSCredentialsRefresher.java
```diff
@@ -0,0 +1,110 @@
...
+import org.slf4j.LoggerFactory;
+
+import java.io.Closeable;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.TimeUnit;
+
+import static com.yahoo.athenz.common.ServerCommonConsts.*;
+
+public class AWSCredentialsRefresher implements Closeable {
```
havetisyan, line 33: Do we really need this class nowadays? Doesn't AWS provide us with a credentials provider that is asynchronously refreshed in the background since we're running in EC2 thus we don't need to write this class at all?

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/AWSInstanceMetadataFetcher.java
```diff
@@ -0,0 +1,259 @@
...
+
+public class AWSInstanceMetadataFetcher implements Closeable {
+    private static final Logger LOGGER = LoggerFactory.getLogger(AWSInstanceMetadataFetcher.class);
+
+    String awsRole = null;
+
+    String awsRegion;
+    private HttpClient httpClient;
+
+    public AWSInstanceMetadataFetcher() {
```
havetisyan, line 41: Based on my comment on the S3 client code, does this need to be in server common or it belongs in its original location back in ZTS?

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/AWSCredentialsRefresher.java
OferLevi85: Yes, we usually don't need to specify credentials when working in AWS but I assumed that there is a reason for us getting the credentials through the metadata in CloudStore (that I'm not familiar with).
If it is just for historical reasons (wasn't possible in older AWS SDK), I'll just leave the DataStore as is and just build the s3client without credentials

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/AWSInstanceMetadataFetcher.java
OferLevi85: If providing the credentials isn't necessary then both classes are not required

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/AWSCredentialsRefresher.java
havetisyan: Yes, that's the right approach. I think we had some issues (this is several years back now) with the credentials provider so we opted to fetch the creds ourselves and it's been on my list all this time to revisit this code. But let's just remove it and rely on providers from aws directly. At some point we also need to look into what it would take to move to 2.x but for now let's stick with 1.x of aws java sdk.

PR #1048, opened 2020-07-19

OferLevi85:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/S3ChangeLogStore.java
```diff
@@ -72,6 +80,13 @@ public S3ChangeLogStore(CloudStore cloudStore) {
         jsonMapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
     }
 
+    void initAwsRegion() {
+        awsRegion = System.getProperty(ZTS_PROP_AWS_REGION_NAME);
+        if (StringUtil.isEmpty(awsRegion)) {
+            awsRegion = EC2MetadataUtils.getEC2InstanceRegion();
```
OferLevi85, line 86: If region is specified, I'm taking it from the instance metadata as in AWSObjectStoreFactory

PR #1049, opened 2020-07-20
Add instructions on deleting cert records in Cert Refresh Failure mail

OferLevi85:

### DIFF
--- a/servers/zts/src/main/resources/messages/unrefreshed-certs.html
+++ b/servers/zts/src/main/resources/messages/unrefreshed-certs.html
@@ -45,7 +45,14 @@
         </table>
         <hr>
         <div class="bt unrefreshedcerts">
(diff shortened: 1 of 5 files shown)


servers/zts/src/main/resources/messages/unrefreshed-certs.html
```diff
@@ -45,7 +45,14 @@
         </table>
         <hr>
         <div class="bt unrefreshedcerts">
-            <br>Please review this list and, if necessary, login to the hosts and verify sia runs on schedule.
+            <br>Please review this list and take one of the following actions:
+            <br>
+            <br> 1. For entries belonging to hosts that were re-bootstrapped or changed identity, please delete the
```
havetisyan, line 50: let's change the order. some engineers will just do the first step and delete the record and make things worst. So we should say that you should login first and verify that it's working correctly first.

PR #1055, opened 2020-07-25
Added missing ZTS tests

OferLevi85:


servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBClientFetcherImpl.java
```diff
@@ -48,7 +58,10 @@ public DynamoDBClientAndCredentials getDynamoDBClient(ZTSClientNotificationSende
             LOGGER.info("DynamoDB client will use existing AWS authentication");
             if (dynamoDBClientSettings.getRegion().isEmpty()) {
                 // Use default region
-                AmazonDynamoDB client = AmazonDynamoDBClientBuilder.standard().build();
+                AmazonDynamoDB client = AmazonDynamoDBClientBuilder
```
OferLevi85, line 61: I couldn't get the test to pass in Travis CI without explicitly specifying the region (locally it works fine). 
I'm getting the following message: "com.amazonaws.SdkClientException: Unable to find a region via the region provider chain. Must provide an explicit region in the builder or setup environment to supply a region."

I've decided to provide the region explicitly (like we do in the S3ChangeLogStore and AwsPrivateKeyStor
e) just to be on the safe side.

PR #1056, opened 2020-07-25
Disable OPTIONS and TRACE

OferLevi85:
For ZMS only - enable OPTIONS for /zms/v1


servers/zts/src/main/webapp/WEB-INF/web.xml
```diff
@@ -17,4 +17,23 @@
         <servlet-name>zts-server</servlet-name>
         <url-pattern>/*</url-pattern>
     </servlet-mapping>
+    <!-- ==================================================================== -->
```
ShayDavidson, line 20: typo

PR #1062, opened 2020-07-28
fix updating ssh records from on-prem to dynamodb

havetisyan:


servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBClientFetcherImpl.java
```diff
@@ -33,10 +34,9 @@
 
 public class DynamoDBClientFetcherImpl implements DynamoDBClientFetcher {
     private static final Logger LOGGER = LoggerFactory.getLogger(DynamoDBClientFetcherImpl.class);
-    private final String defaultAwsRegion;
+    private String defaultAwsRegion;
 
     public DynamoDBClientFetcherImpl() {
-        this.defaultAwsRegion = EC2MetadataUtils.getEC2InstanceRegion();
```
havetisyan, line 39: moved the code within the function since during startup we don't have aws access on prep so we see the exceptions

PR #1071, opened 2020-07-30
[DO NOT MERGE] Sample generate metrics rdl

OferLevi85:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSHandler.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSHandler.java
@@ -96,4 +96,6 @@ public interface ZMSHandler {
     DomainRoleMembership getPendingDomainRoleMembersList(ResourceContext context, String principal);
     Schema getRdlSchema(ResourceContext context);
     ResourceContext newResourceContext(HttpServletRequest request, HttpServletResponse response);
+    void recordMetrics(String httpMethod, int httpStatus, String apiName, ResourceContext ctx);
+    Object startMetricTiming();
 }
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -1131,7 +1131,6 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
     public Domain getDomain(ResourceContext ctx, String domainName) {
 
         final String caller = "getdomain";
-        metric.increment(ZMSConsts.HTTP_GET);
         logPrincipal(ctx);
 
         validateRequest(ctx.request(), caller);
@@ -1142,18 +1141,14 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         // policy, service, etc name)
 
         domainName = domainName.toLowerCase();
+        setRequestDomain(ctx, domainName);
         final String principalDomain = getPrincipalDomain(ctx);
-        metric.increment(ZMSConsts.HTTP_REQUEST, domainName, principalDomain);
-        metric.increment(caller, domainName, principalDomain);
-
-        Object timerMetric = metric.startTiming("getdomain_timing", domainName, principalDomain);
 
         Domain domain = dbService.getDomain(domainName, false);
         if (domain == null) {
             throw ZMSUtils.notFoundError("getDomain: Domain not found: " + domainName, caller);
         }
 
-        metric.stopTiming(timerMetric, domainName, principalDomain);
         return domain;
     }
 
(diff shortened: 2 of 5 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -8279,4 +8282,19 @@ public void run() {
...
+        String domainName = getRequestDomainName(ctx);
+        metric.increment("zms_api", domainName, principalDomainName, httpMethod, httpStatus, apiName);
+    }
+
+    public Object startMetricTiming() {
+        return metric.startTiming("zms_api_latency", null);
+    }
+
+    public void stopTiming(Object timerMetric, String httpMethod, int httpStatus, String apiName, ResourceContext ctx) {
+        String principalDomainName = getPrincipalDomain(ctx);
```
havetisyan, line 8296: should make these 2 final as well.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
```diff
@@ -17,12 +17,15 @@
     @Path("/domain/{domain}")
     @Produces(MediaType.APPLICATION_JSON)
     public Domain getDomain(@PathParam("domain") String domain) {
+        int code = 200;
```
havetisyan, line 20: we should use ResourceException.OK

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
```diff
@@ -17,12 +17,15 @@
...
+        ResourceContext context = this.delegate.newResourceContext(this.request, this.response);
+        Object timerMetric = this.delegate.startMetricTiming();
         try {
-            ResourceContext context = this.delegate.newResourceContext(this.request, this.response);
             context.authenticate();
-            return this.delegate.getDomain(context, domain);
+            Domain returnedValue = this.delegate.getDomain(context, domain);
```
havetisyan, line 25: not sure why this change was made?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
```diff
@@ -38,6 +41,9 @@ public Domain getDomain(@PathParam("domain") String domain) {
                 System.err.println("*** Warning: undeclared exception (" + code + ") for resource getDomain");
                 throw typedException(code, e, ResourceError.class);
             }
+        } finally {
+            this.delegate.stopTiming(timerMetric, "GET", code, "getDomain", context);
```
havetisyan, line 45: since the code is auto generated I'm guessing the generator needs to include the "getDomain" here and in the next line automatically. otherwise we can just set the method name in the resource context as well.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
```diff
@@ -17,12 +17,15 @@
     @Path("/domain/{domain}")
     @Produces(MediaType.APPLICATION_JSON)
     public Domain getDomain(@PathParam("domain") String domain) {
+        int code = 200;
+        ResourceContext context = this.delegate.newResourceContext(this.request, this.response);
+        Object timerMetric = this.delegate.startMetricTiming();
```
havetisyan, line 22: we should probably move this one line up since we should include the time required for context resource creation as well. Another option would be create the object within ResourceContext itself and then just get that on line 45 below.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
```diff
@@ -38,6 +41,9 @@ public Domain getDomain(@PathParam("domain") String domain) {
                 System.err.println("*** Warning: undeclared exception (" + code + ") for resource getDomain");
                 throw typedException(code, e, ResourceError.class);
             }
+        } finally {
+            this.delegate.stopTiming(timerMetric, "GET", code, "getDomain", context);
+            this.delegate.incrementMetric("GET", code, "getDomain", context);
```
havetisyan, line 46: Also see if you can combine these two calls into one generic update metric call and then internally we can make two separate calls - one for increment and one for the timer. at least the code generation might look simpler

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
OferLevi85: Yes I assumed that while way we generate the function name in ZMSResources we can put it here

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
OferLevi85: Mistake. I originally planned on stopping the timer just before returning the value (before figuring I can stop it in finally and have the error code if needed)

OferLevi85:
Fixed it according to your comments so you can see the final result. No need to merge this of course. I've started working on the code generation.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSHandler.java
```diff
@@ -96,4 +96,6 @@
     DomainRoleMembership getPendingDomainRoleMembersList(ResourceContext context, String principal);
     Schema getRdlSchema(ResourceContext context);
     ResourceContext newResourceContext(HttpServletRequest request, HttpServletResponse response);
+    void recordMetrics(String httpMethod, int httpStatus, String apiName, ResourceContext ctx);
```
havetisyan, line 99: let's use the ResourceContext as the first argument to be consistent with other APIs

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSHandler.java
```diff
@@ -96,4 +96,6 @@
...
+    Object startMetricTiming();
```
havetisyan, line 100: not sure why we need this? we should be able to hide the constructions of the metric within the newResourceContext api call so we don't need to expose this

PR #1073, opened 2020-08-02
[Merge only after UltraBrew changes are merged]Generate Metrics code in API Requests

OferLevi85:

### DIFF
--- a/rdl/rdl-gen-athenz-server/main.go
+++ b/rdl/rdl-gen-athenz-server/main.go
@@ -257,6 +257,7 @@ import javax.servlet.http.HttpServletResponse;
 public interface {{cName}}Handler {{openBrace}} {{range .Resources}}
     {{methodSig .}};{{end}}
     ResourceContext newResourceContext(HttpServletRequest request, HttpServletResponse response);
+    void recordMetrics(ResourceContext ctx, String httpMethod, int httpStatus, String apiName);
 }
 `
 
@@ -407,8 +408,10 @@ func (gen *javaServerGenerator) resourcePath(r *rdl.Resource) string {
 
 func (gen *javaServerGenerator) handlerBody(r *rdl.Resource) string {
 	noContent := r.Expected == "NO_CONTENT" && r.Alternatives == nil
-	s := "        try {\n"
-	s += "            ResourceContext context = this.delegate.newResourceContext(this.request, this.response);\n"
+	s := "        int code = ResourceException.OK;\n"
+	s += "        ResourceContext context = null;\n"
+	s += "        try {\n"
+	s += "            context = this.delegate.newResourceContext(this.request, this.response);\n"
 	var fargs []string
 	bodyName := ""
 	if r.Auth != nil {
@@ -457,7 +460,7 @@ func (gen *javaServerGenerator) handlerBody(r *rdl.Resource) string {
 		s += "            return this.delegate." + methName + "(context" + sargs + ");\n"
 	}
 	s += "        } catch (ResourceException e) {\n"
-	s += "            int code = e.getCode();\n"
+	s += "            code = e.getCode();\n"
 	s += "            switch (code) {\n"
 	if r.Exceptions != nil && len(r.Exceptions) > 0 {
 		keys := sortedExceptionKeys(r.Exceptions)
@@ -471,6 +474,8 @@ func (gen *javaServerGenerator) handlerBody(r *rdl.Resource) string {
 	s += "                System.err.println(\"*** Warning: undeclared exception (\" + code + \") for resource " + methName + "\");\n"
 	s += "                throw typedException(code, e, ResourceError.class);\n" //? really
 	s += "            }\n"
+	s += "        } finally {\n"
+	s += "            this.delegate.recordMetrics(context, \"" + r.Method + "\", code, \"" + methName + "\");\n"
 	s += "        }\n"
 	return s
 }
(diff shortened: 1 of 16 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2536,15 +2433,14 @@ Access getAccessCheck(Principal principal, String action, String resource,
         // retrieve the domain based on our resource and action/trustDomain pair
 
         String domainName = retrieveResourceDomain(resource, action, trustDomain);
+        setRequestDomain(ctx, domainName);
         if (domainName == null) {
-            metric.increment(ZMSConsts.HTTP_REQUEST, ZMSConsts.ZMS_INVALID_DOMAIN, principal.getDomain());
-            metric.increment(caller, ZMSConsts.ZMS_INVALID_DOMAIN, principal.getDomain());
+            setRequestDomain(ctx, ZMSConsts.ZMS_INVALID_DOMAIN);
```
OferLevi85, line 2438: I'm setting the domain name here to "invalid_domain"

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2536,15 +2433,14 @@ Access getAccessCheck(Principal principal, String action, String resource,
...
             throw ZMSUtils.notFoundError("getAccessCheck: Unable to extract resource domain", caller);
         }
         AthenzDomain domain = retrieveAccessDomain(domainName, principal);
         if (domain == null) {
-            metric.increment(ZMSConsts.HTTP_REQUEST, ZMSConsts.ZMS_UNKNOWN_DOMAIN, principal.getDomain());
-            metric.increment(caller, ZMSConsts.ZMS_UNKNOWN_DOMAIN, principal.getDomain());
+            setRequestDomain(ctx, ZMSConsts.ZMS_UNKNOWN_DOMAIN);
```
OferLevi85, line 2443: I'm setting the domain name here to "unknown_domain"

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -1054,12 +1054,7 @@ public DomainList getDomainList(ResourceContext ctx, Integer limit, String skip,
             String modifiedSince) {
 
         final String caller = "getdomainlist";
```
OferLevi85, line 1056: I originally had the caller (method name in lowercase) placed in the context in ZMSResources and then I was able to have the same line for every method: "final String caller = ctx.getApiName();".
The reason I didn't change it eventually as I noticed that it is sometimes used in authority checks (as the operation in verifyAuthorizedServiceOperation calls). So maybe it is best to keep it explicit.
Let me know what you think.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5385,6 +5026,7 @@ public Response getSignedDomains(ResourceContext ctx, String domainName, String
         if (domainName != null) {
             domainName = domainName.toLowerCase();
             validate(domainName, TYPE_DOMAIN_NAME, caller);
+            setRequestDomain(ctx, domainName);
```
OferLevi85, line 5029: Originally this method doesn't send the domain but I think it is correct to put it (if it exists)

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5530,6 +5166,7 @@ public JWSDomain getJWSDomain(ResourceContext ctx, String domainName) {
         // policy, service, etc name)
 
         domainName = domainName.toLowerCase();
+        setRequestDomain(ctx, domainName);
```
OferLevi85, line 5169: Originally this method doesn't send the domain but I think it is correct to put it (if it exists)

rdl/rdl-gen-athenz-server/main.go
```diff
@@ -407,8 +408,9 @@ func (gen *javaServerGenerator) resourcePath(r *rdl.Resource) string {
 
 func (gen *javaServerGenerator) handlerBody(r *rdl.Resource) string {
 	noContent := r.Expected == "NO_CONTENT" && r.Alternatives == nil
-	s := "        try {\n"
-	s += "            ResourceContext context = this.delegate.newResourceContext(this.request, this.response);\n"
+	s := "        int code = ResourceException.OK;\n"
+	s += "        ResourceContext context = this.delegate.newResourceContext(this.request, this.response);\n"
```
havetisyan, line 412: Not sure we can move this call out of try block. If it throws an exception, it might expose our trace and some other internal data to the caller which we can't have. We should be able to define the ResourceContext context = null outside of the try block and keep the delegate call within the block and within the recordMetrics calls make sure to handle the case where context is null.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2938,9 +2784,6 @@ public InstanceIdentity postInstanceRefreshInformation(ResourceContext ctx, Stri
             String domain, String service, String instanceId, InstanceRefreshInformation info) {
 
         final String caller = "postinstancerefreshinformation";
-        final String callerTiming = "postinstancerefreshinformation_timing";
-
-        metric.increment(HTTP_POST);
```
OferLevi85, line 2943: postInstanceRefreshInformation has many metrics that are not related to the HTTP request (such as "providerconfirm"). I didn't change them.

PR #1075, opened 2020-08-04
Update Centrelized Authentication example (using X509 Cert instead of…

OferLevi85:
… NToken)


upgle:
Hi @OferLevi85, I am trying to apply authentication using Ntoken in go language.
Is this Ntoken deprecated? Or is it possible to use both x509 and Ntoken? I can't find any documentation about it, so I'm leaving a comment here. Thanks.

havetisyan:
Hi @upgle,

NToken is not deprecated through we prefer the use of x.509 certificates which are much more secure. Support for NTokens is controlled whether or not your ZMS server is configured with the PrincipalAuthority (NTokens) or not. I would suggest you join our slack channel if you have any further questions and we'll be happy to help. (Click on the Slack icon in the top right corner on our website: https://www.athenz.io)

PR #1076, opened 2020-08-05
use embedded mysql server for unit tests

havetisyan:


servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
```diff
@@ -5170,15 +5190,6 @@ public void testGetExpiredPendingDomainRoleMembers() {
         Role role = zms.dbService.getRole(domainName, roleName, false, false, true);
         assertEquals(role.getRoleMembers().size(), 4);
 
-        // now let's expire our entries
-
```
OferLevi85, line 5174: I assumed that this was removed because it is already covered in testProcessExpiredPendingMembers.
But in testProcessExpiredPendingMembers there's no assertions. The test just runs processExpiredPendingMembers and doesn't verify that the members were expired

havetisyan: It was removed since it can't be tested with a real database. The test used the file connection where we forced the request time to be whatever we wanted (in this case we predated back 31 days). with the jdbc, the request is set when you insert the object and defaults to current time. So we can't expire anything. So while the actual lines are covered by other mock tests, the real test will not be possible with MySQL DB.

servers/zms/src/test/resources/schema/zms_server.sql
```diff
@@ -0,0 +1,357 @@
+-- MySQL Script generated by MySQL Workbench
+-- Tue Jun 30 12:33:10 2020
+-- Model: New Model    Version: 1.0
+-- MySQL Workbench Forward Engineering
```
OferLevi85, line 4: Is this file different from the original zms_server.sql? If it isn't that means we will need to remember updating this file whenever we update the original.
Can we reference the original file instead of copying it to the test directory?

havetisyan: It's the same file for now. remembering part is not a concern since if you write unit tests and forget to update the sql file, your tests are going to fail anyway. but the library is based around resources and I have to review their actual code to see how I can specify a relative path (which might be different depending what. class we're calling this from??).

PR #1077, opened 2020-08-06
Changes in Metrics API

OferLevi85:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -1054,7 +1054,7 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
             Integer depth, String account, Integer productId, String roleMember, String roleName,
             String modifiedSince) {
 
-        final String caller = "getdomainlist";
+        final String caller = ctx.getApiName();
 
         logPrincipal(ctx);
 
@@ -1125,7 +1125,7 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
 
     public Domain getDomain(ResourceContext ctx, String domainName) {
 
-        final String caller = "getdomain";
+        final String caller = ctx.getApiName();
         logPrincipal(ctx);
 
         validateRequest(ctx.request(), caller);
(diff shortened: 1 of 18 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -7231,8 +7231,10 @@ void logPrincipal(ResourceContext ctx) {
     }
 
     public ResourceContext newResourceContext(HttpServletRequest request,
-                                              HttpServletResponse response) {
-        Object timerMetric = metric.startTiming("zms_api_latency", null);
+                                              HttpServletResponse response,
+                                              String httpMethod,
+                                              String apiName) {
+        Object timerMetric = metric.startTiming("zms_api_latency", null, null, httpMethod, apiName);
```
OferLevi85, line 7237: Getting the principal domain will require me to start the timer after the principal is authenticated (so not when creating the context). 
For now, I've decided to leave this out along with the request domain.
This will only affect implementers that get the tags at the start of the timer (we send all arguments when stopping the timer)

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3643,8 +3643,8 @@ Principal createPrincipalForName(String principalName) {
     @Override
     public ResourceAccess getResourceAccessExt(ResourceContext ctx, String action, String resource,
             String trustDomain, String checkPrincipal) {
-        
-        final String caller = "getaccessext";
+
+        final String caller = ctx.getApiName();
```
OferLevi85, line 3647: Please note that I have changed the caller from "getaccessext" to "getresourceaccessext". Unlike the call in putServiceIdentitySystemMeta, there is no call to verifyAuthorizedServiceOperation so this should affect anything.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4552,7 +4552,7 @@ public void putServiceIdentitySystemMeta(ResourceContext ctx, String domainName,
         verifyAuthorizedServiceOperation(principal.getAuthorizedService(), caller);
```
OferLevi85, line 4553: Unlike the other methods, I didn't change the caller as the name of the method "putServiceIdentitySystemMeta" is different than the caller "putservicesystemmeta". Changing this will break the call to verifyAuthorizedServiceOperation as it appears differently in authorized_services.json file

PR #1086, opened 2020-08-16
Enable case-sensitive assertion action/resource value storage

OferLevi85:


core/zms/src/main/java/com/yahoo/athenz/zms/Assertion.java
```diff
@@ -20,6 +20,9 @@
     @RdlOptional
     @JsonInclude(JsonInclude.Include.NON_EMPTY)
     public Long id;
+    @RdlOptional
+    @JsonInclude(JsonInclude.Include.NON_EMPTY)
+    public Boolean caseSensitive;
```
OferLevi85, line 25: I didn't want to change the API. Users that would like to save assertion as case-sensitive will need to set the assertion as caseSensitive as part of the structure passed to Athenz in putAssertion and putPolicy.
Please note the following:
1. This value will not be saved to DB. It will only be used to tell ZMS to keep the resource and action as case sensitive
2. The policy structure will also have this field. If it is set, all assertions will be lowercased

servers/zms/src/main/java/com/yahoo/athenz/zms/utils/ZMSUtils.java
```diff
@@ -406,4 +406,18 @@ public static String combineUserAuthorityFilters(final String roleUserAuthorityF
 
         return authorityFilter;
     }
+
+    public static String lowerDomainInResource(String resource) {
```
OferLevi85, line 410: Even if we want to keep in original case - domain will be lower-cased.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4320,6 +4333,10 @@ AthenzDomain virtualHomeDomain(Principal principal, String domainName) {
     boolean assertionMatch(Assertion assertion, String identity, String action, String resource,
             String domain, List<Role> roles, List<String> authenticatedRoles, String trustDomain) {
 
+        // Lowercase action and resource in assertion as it is possible to store them case-sensitive
+        assertion.setResource(assertion.getResource().toLowerCase());
```
OferLevi85, line 4337: We should now lower-case assertions from the DB before checking them. This change was also made in ZTSAuthorizer and ZPE (ZpeUpdPolLoader)

havetisyan:
this change allows if there is an application that it does the authorization based on case sensitive values, the user can store them as such in Athen ZMS without automatically being converted to lowercase.

however, for authz checks within ZMS and ZTS, it will still only support case-insensitive matches.

PR #1097, opened 2020-08-27
initial changes for group implementation

havetisyan:


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -125,14 +126,17 @@ void setAuditRefObjectBits() {
         // the value is a comma separated list of supported objects:
```
OferLevi85, line 126: Please change the comment from  "by default we're only going to handle audit enabled roles" to "by default we're only going to handle audit enabled roles and groups"

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -934,6 +1026,63 @@ void executePutRole(ResourceContext ctx, String domainName, String roleName, Rol
...
+
+                // first verify that auditing requirements are met
+
+                checkDomainAuditEnabled(con, domainName, auditRef, ctx.getApiName(), principal, AUDIT_TYPE_GROUP);
+
+                // check that quota is not exceeded
+
+                quotaCheck.checkGroupQuota(con, domainName, group, ctx.getApiName());
+
+                // retrieve our original role
```
OferLevi85, line 1049: // retrieve our original group

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1348,6 +1560,98 @@ void executeDeletePendingMembership(ResourceContext ctx, String domainName, Stri
...
+
+                // process our delete pending group member operation
+
+                if (!con.deletePendingGroupMember(domainName, groupName, normalizedMember, principal, auditRef)) {
+                    con.rollbackChanges();
+                    throw ZMSUtils.notFoundError("unable to delete pending group member: " +
+                            normalizedMember + " from group: " + groupName, ctx.getApiName());
+                }
+
+                // update our role and domain time-stamps, and invalidate local cache entry
```
OferLevi85, line 1634: // update our group and domain time-stamps, and invalidate local cache entry

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2624,36 +3017,46 @@ void updateSystemMetaFields(Domain domain, final String attribute, boolean delet
...
+        if (ZMSConsts.SYSTEM_META_AUDIT_ENABLED.equals(attribute)) {
+            role.setAuditEnabled(meta.getAuditEnabled());
+        } else {
+            throw ZMSUtils.requestError("unknown role system meta attribute: " + attribute, caller);
+        }
+    }
+
+    void updateGroupSystemMetaFields(Group group, final String attribute, GroupSystemMeta meta) {
+
+        final String caller = "putgroupsystemmeta";
```
OferLevi85, line 3036: I think that it will be preferable to place the content of this method inline in executePutGroupSystemMeta or make the caller an input argument.

This method is called only from executePutGroupSystemMeta so the "caller" is correct but if someone will reuse this method then he can accidentally leave this value as-is.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4206,6 +4722,96 @@ public void executePutRoleMeta(ResourceContext ctx, String domainName, String ro
...
+                if (originalGroup == null) {
+                    con.rollbackChanges();
+                    throw ZMSUtils.notFoundError("Unknown group: " + groupName, ctx.getApiName());
+                }
+
+                checkObjectAuditEnabled(con, originalGroup.getAuditEnabled(), originalGroup.getName(),
+                        auditRef, ctx.getApiName(), getPrincipalName(ctx));
+
+                // now process the request. first we're going to make a
+                // copy of our role
```
OferLevi85, line 4765: // copy of our group

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4206,6 +4722,96 @@ public void executePutRoleMeta(ResourceContext ctx, String domainName, String ro
...
+
+                // audit log the request
+
+                StringBuilder auditDetails = new StringBuilder(ZMSConsts.STRING_BLDR_SIZE_DEFAULT);
+                auditLogGroupMeta(auditDetails, updatedGroup, groupName);
+
+                auditLogRequest(ctx, domainName, auditRef, ctx.getApiName(), ZMSConsts.HTTP_PUT,
+                        domainName, auditDetails.toString());
+
+                // if the role member expiry date or review date has changed then we're going
```
OferLevi85, line 4792: if the role member => if the group member

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -570,6 +570,61 @@ boolean processRole(ObjectStoreConnection con, Role originalRole, String domainN
         return true;
     }
 
+    boolean processGroup(ObjectStoreConnection con, Group originalGroup, final String domainName,
+                        final String groupName, Group group, final String admin, final String auditRef,
+                        StringBuilder auditDetails) {
+
+        // check to see if we need to insert the role or update it
```
OferLevi85, line 577: insert the role => insert the group

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -3779,6 +4182,38 @@ boolean auditLogRoleMember(StringBuilder auditDetails, RoleMember roleMember, bo
...
+        }
+        auditDetails.append(", \"approved\": ");
+        auditDetails.append(auditLogBooleanDefault(groupMember.getApproved(), Boolean.FALSE));
+        auditDetails.append(", \"system-disabled\": ");
+        auditDetails.append(groupMember.getSystemDisabled() == null ? 0 : groupMember.getSystemDisabled());
+        auditDetails.append("}");
+        return firstEntry;
+    }
+
+    String auditLogBooleanDefault(Boolean value, Boolean defaultValue) {
```
OferLevi85, line 4209: I might be missing something but shouldn't this be reversed?
     if (defaultValue == Boolean.TRUE) {
            return value == Boolean.FALSE ? "false" : "true"; // in any other case that it isn't false - set to true
        } else {
            return value == Boolean.TRUE ? "true" : "false"; // in any other case that it isn't true - set to false
        }

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4531,6 +5314,41 @@ void updateRoleMembersSystemDisabledState(ResourceContext ctx, ObjectStoreConnec
...
+        // check if the authority filter has changed otherwise we have
+        // nothing to do
+
+        if (!isUserAuthorityFilterChanged(originalGroup.getUserAuthorityFilter(), updatedGroup.getUserAuthorityFilter())) {
+            return;
+        }
+
+        final String principal = getPrincipalName(ctx);
+
+        // process our role members and if there were any changes processed then update
```
OferLevi85, line 5336: our role members => our group members

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4531,6 +5314,41 @@ void updateRoleMembersSystemDisabledState(ResourceContext ctx, ObjectStoreConnec
...
+        // our role and domain time-stamps, and invalidate local cache entry
+
+        List<GroupMember> groupMembersWithUpdatedDisabledState = getGroupMembersWithUpdatedDisabledState(groupMembers,
+                updatedGroup.getUserAuthorityFilter(), getDomainUserAuthorityFilter(con, domainName));
+        if (updateGroupMemberDisabledState(ctx, con, groupMembersWithUpdatedDisabledState, domainName,
+                groupName, principal, auditRef, ctx.getApiName())) {
+
+            // update our role and domain time-stamps, and invalidate local cache entry
```
OferLevi85, line 5344: update our group and domain

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -5001,6 +6047,42 @@ void processUserAuthorityRestrictions() {
...
+        List<PrincipalGroup> groups;
+        try (ObjectStoreConnection con = store.getConnection(true, false)) {
+            groups = con.listGroupsWithUserAuthorityRestrictions();
+        }
+
+        if (groups == null) {
+            return;
+        }
+
+        // for each role catch any exception and ignore since we
```
OferLevi85, line 6071: role => group

OferLevi85 (changes requested):
Small issues. Mostly fixing comments

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
havetisyan: the logic is correct but I renamed the defaultValue to checkValue to be more clear the usage. the unit test has more explanation

OferLevi85 (approved):

PR #1099, opened 2020-09-01
Implemented restricted mTLS certs

OferLevi85:


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/GlobStringsMatcher.java
```diff
@@ -30,7 +29,7 @@
     private final List<String> patterns;
 
     public GlobStringsMatcher(String systemProperty) {
-        List<String> globList = ZTSUtils.splitCommaSeperatedSystemProperty(systemProperty);
+        List<String> globList = AthenzUtils.splitCommaSeperatedSystemProperty(systemProperty);
```
havetisyan, line 32: we should update the debug statement on line 35 below. it's referring to Ignored services - we should just include the property name in the statement

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3482,6 +3482,8 @@ public void putMembership(ResourceContext ctx, String domainName, String roleNam
 
         verifyAuthorizedServiceRoleOperation(principal.getAuthorizedService(), caller, roleName);
 
+        validateNotMtlsRestricted(ctx, caller);
```
havetisyan, line 3485: For ZMS - why not just put the check in the authorize call. We should not allow any write operations with the restricted cert.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -1372,6 +1372,8 @@ public RoleToken getRoleToken(ResourceContext ctx, String domainName, String rol
                     ZTSConsts.ZTS_UNKNOWN_DOMAIN, principalDomain);
         }
 
+        validateNotMtlsRestricted(ctx, caller, domainName);
```
havetisyan, line 1375: For ZTS, we should be probably be more strict and add it to the authenticate method. We should not allow anyone with the restricted certs to connect to ZTS. I can't think of any reason, can you? This would simplify the code for any future commands as well.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
```diff
@@ -1115,6 +1117,12 @@ public static String extractX509CertSubjectOField(X509Certificate x509Cert) {
         return extractX509CertSubjectField(x509Cert, BCStyle.O);
     }
 
+    public static boolean isRestrictedCertificate(X509Certificate x509Cert) {
```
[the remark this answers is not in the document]
OferLevi85: Yeah sounds good. At first I had a single static instance of GlobStringsMatcher in Crypto but it made it hard to unit-test as it required each test to re-initialize the static instance (so it can re-read the property).
I eventually placed it in the method to write the unit-tests and I planned on fixing it by doing the reload in each unit-test before the push but it slipped my mind.
But yeah making the Authority pass the reference is actually a much better solution.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
OferLevi85: There are some methods to the ones in ZMS where user without authorization can retrieve data.
For example:
- enable users the option to check access with the restricted certs (get /access)
- Get services in a domain (GET /domain/{domainName}/service/)
- Get a service public key (GET /domain/{domainName}/service/{serviceName}/publickey/{keyId}) 
- Get a list of services to run on a specific host (GET /host/{host}/services)
- Get Status (/status)

We can block them but then why block them and not all the methods in ZMS?

havetisyan: those operations can be carried out against ZMS. There is no point of going to ZTS to get a service or public keys. and if you have a restricted cert, you should not really carry out any access checks. It's basically a limited client certificate. So I think I'm ok restricting from zts server completely.

havetisyan (changes requested):

havetisyan (approved):

PR #1102, opened 2020-09-02
adding group-expiry-days attributes to role/domain meta objects

havetisyan:
The PR is mostly auto generated code and setting/getting the attribute on role and domain objects. 


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -1801,6 +1804,7 @@ void validateDomainValues(Domain domain) {
         validateIntegerValue(domain.getMemberExpiryDays(), "memberExpiryDays");
         validateIntegerValue(domain.getRoleCertExpiryMins(), "roleCertExpiryMins");
         validateIntegerValue(domain.getServiceExpiryDays(), "serviceExpiryDays");
+        validateIntegerValue(domain.getGroupExpiryDays(), "serviceExpiryDays");
```
abvaidya, line 1807: second argument should be "groupExpiryDays"

abvaidya (changes requested):

abvaidya (approved):

PR #1105, opened 2020-09-04
support adding groups as role members

havetisyan:
since role members can include *'s and group members and group members must only include users and services, updated MemberName definition and switched Groups to use new GroupMemberName definition so the validation takes place correctly.

introduced a concept of principal type in member objects so we don't keep parsing to find out if the principal is user, service or group. 


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3566,7 +3617,7 @@ String enforcedUserAuthorityFilter(final String roleUserAuthorityFilter, final S
     }
 
     boolean shouldValidateRoleMembers(final String userAuthorityFilter) {
-        return validateUserRoleMembers || validateServiceRoleMembers || userAuthorityFilter != null;
+        return validateGroupRoleMembers || validateUserRoleMembers || validateServiceRoleMembers || userAuthorityFilter != null;
```
abvaidya, line 3620: This method is always going to return true, so does it make sense to just remove it?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -8043,6 +8095,13 @@ void normalizeGroupMembers(Group group) {
                 addNormalizedGroupMember(normalizedMembers, member);
             }
         }
+
+        // go through each member and set the principal type
+
+        for (GroupMember member : normalizedMembers.values()) {
```
abvaidya, line 8101: instead of looping over again, can we set the principal type which adding them to "normalizedGroupMembers" above?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
abvaidya: I see all references are removed?

abvaidya (changes requested):

core/zms/src/main/rdl/Group.tdl
```diff
@@ -25,11 +25,12 @@ type GroupMember Struct {
     ResourceName requestPrincipal (optional); //pending members only - name of the principal requesting the change
     Timestamp reviewLastNotifiedTime (optional); //for pending membership requests, time when last notification was sent (for file store)
     Int32 systemDisabled (optional); //user disabled by system based on configured group setting
+    Int32 principalType (optional); //server use only - principal type: user or service
 }
 
 //The representation for a group membership.
 type GroupMembership Struct {
```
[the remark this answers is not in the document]
havetisyan: in the next PR when we add support for membership we'll probably add those. this pr only handles putRole operations

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
havetisyan: eventually yes, it will be removed. I've removed in a couple of place already. once other functionality is added, it will slowly be removed

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
havetisyan: sure, we'll just need to add some logic to skip duplicates and don't waste lookup operations.

core/zms/src/main/rdl/Role.tdl
```diff
@@ -30,6 +30,7 @@ type RoleMember Struct {
     ResourceName requestPrincipal (optional); //pending members only - name of the principal requesting the change
     Timestamp reviewLastNotifiedTime (optional); //for pending membership requests, time when last notification was sent (for file store)
     Int32 systemDisabled (optional); //user disabled by system based on configured role setting
+    Int32 principalType (optional); //server use only - principal type: user, group, or service
```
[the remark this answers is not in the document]
havetisyan: sure

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
havetisyan: ok, looks like it got cleaned up in all place, so I'll remove it in this PR as well.

abvaidya (approved):

PR #1106, opened 2020-09-09
validation of role/group members

havetisyan:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -3192,7 +3192,67 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         }
     }
 
-    void validateRoleMemberPrincipal(final String memberName, int principalType, final String userAuthorityFilter, final String caller) {
+    void validateUserPrincipal(final String memberName, boolean validateUserMember, final String userAuthorityFilter,
+                               final String caller) {
+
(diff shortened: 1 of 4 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3192,7 +3192,67 @@ void updateRoleMemberUserAuthorityExpiry(final Role role, final String caller) {
...
+            if (dbService.getServiceIdentity(domainName, serviceName, true) == null) {
+                throw ZMSUtils.requestError("Principal " + memberName + " is not a valid service", caller);
+            }
+        }
+    }
+
+    void validateGroupPrincipal(final String memberName, final String caller) {
+
+        int idx = memberName.indexOf(AuthorityConsts.GROUP_SEP);
+        final String domainName = memberName.substring(0, idx);
```
abvaidya, line 3246: we can have a check for idx == -1 here as well

abvaidya (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
havetisyan: We don't need it. ZMSUtils.principalType sets the principal type to GROUP only if it contains the AuthorityConsts.GROUP_SEP value thus idx will never be -1. We do need it for service use case since the function treats non user/group types as service thus we need additional validation there.

abvaidya (approved):

PR #1134, opened 2020-10-06
Implemented notification to metrics converters

OferLevi85:

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/Notification.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/Notification.java
@@ -16,6 +16,8 @@
 
 package com.yahoo.athenz.common.server.notification;
 
+import com.yahoo.rdl.Timestamp;
+
 import java.util.*;
 
 public class Notification {
@@ -26,11 +28,11 @@ public class Notification {
     // key value pair describing additional details about notification
     private Map<String, String> details;
 
-    // Type of notification
-    private String type;
-
     // Utility class to convert the notification into an email
     private NotificationToEmailConverter notificationToEmailConverter;
+
+    // Utility class to convert the notification into metric attributes
+    private NotificationToMetricConverter notificationToMetricConverter;
 
     public Notification () {
     }
@@ -72,15 +74,6 @@ public class Notification {
         return this;
     }
 
-    public String getType() {
-        return type;
-    }
-
-    public Notification setType(String type) {
-        this.type = type;
-        return this;
-    }
-
     public Notification setNotificationToEmailConverter(NotificationToEmailConverter notificationToEmailConverter) {
         this.notificationToEmailConverter = notificationToEmailConverter;
         return this;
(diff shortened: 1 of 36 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/Notification.java
```diff
@@ -26,12 +26,12 @@
...
     private Map<String, String> details;
 
-    // Type of notification
-    private String type;
-
     // Utility class to convert the notification into an email
     private NotificationToEmailConverter notificationToEmailConverter;
 
+    // Utility class to convert the notification into metric attributes
+    private NotificationToMetricConverter notificationToMetricConverter;
```
OferLevi85, line 33: NotificationToMetricConverter purpose is similar to NotificationToEmailConverter (instead of mail convert to metrics)

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationMetric.java
```diff
@@ -0,0 +1,77 @@
...
+            return true;
+        }
+        if (o == null || getClass() != o.getClass()) {
+            return false;
+        }
+
+        NotificationMetric that = (NotificationMetric) o;
+
+        if (attributes.size() != ((NotificationMetric) o).attributes.size()) {
+            return false;
```
OferLevi85, line 48: NotificationMetric holds attributes in the data structure List<String[]> (a list of flat array metric records).
So I had to implement custom equals and hashCode (for tests to check equality correctly and to enable holding this structure in a hashmap if we would like to do so in the future).

servers/zms/src/main/java/com/yahoo/athenz/zms/notification/PendingGroupMembershipApprovalNotificationTask.java
```diff
@@ -88,4 +91,20 @@ public NotificationEmail getNotificationAsEmail(Notification notification) {
...
+
+        @Override
+        public NotificationMetric getNotificationAsMetrics(Notification notification) {
+            String[] record = new String[] {
+                    METRIC_NOTIFICATION_TYPE_KEY, NOTIFICATION_TYPE
+            };
+
+            List<String[]> attributes = new ArrayList<>();
+            attributes.add(record);
+            // This notification doesn't contain any details. We should consider adding the recipients in their own tags.
```
OferLevi85, line 106: There are two notification tasks that don't contain details at all (PendingGroupMembershipApprovalNotificationToMetricConverter and PendingRoleMembershipApprovalNotificationToMetricConverter).
Currently, I only record that notification of that type happened. Let me know if I should record the recipients.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/Notification.java
```diff
@@ -93,6 +86,18 @@ public NotificationEmail getNotificationAsEmail() {
         return null;
     }
 
+    public Notification setNotificationToMetricConverter(NotificationToMetricConverter notificationToMetricConverter) {
+        this.notificationToMetricConverter = notificationToMetricConverter;
+        return this;
+    }
+
+    public NotificationMetric getNotificationAsMetrics(Timestamp currentTime) {
```
OferLevi85, line 94: I get the currentTime as an interface argument so equality checks will be successful in tests (calculating the currentTime inside each notification makes a difference for attributes such as "expiry_days" which can sometimes result in test failures)

servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
```diff
@@ -25,17 +25,18 @@
 import com.yahoo.athenz.zts.ZTSConsts;
 import com.yahoo.athenz.zts.cert.InstanceCertManager;
 import com.yahoo.athenz.zts.store.DataStore;
+import com.yahoo.rdl.Timestamp;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.sql.Timestamp;
```
OferLevi85, line 31: I also took advantage of the changes to change the Timestamps used in this notification from "java.sql.Timestamp" to "com.yahoo.rdl.Timestamp".

havetisyan (approved):
👍

PR #1149, opened 2020-10-14
Notification page throws an error when the user/service mentioned con…

pratikgote:
…tains * pattern


abvaidya (approved):

abvaidya (changes requested):
should also change at https://github.com/yahoo/athenz/blob/master/ui/src/components/role/AddMemberForm.js#L200

abvaidya (approved):

PR #1157, opened 2020-10-21
system disabled state for principals 

abvaidya:


servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -5603,6 +5605,43 @@ public DomainGroupMember getPrincipalGroups(String principal, String domainName)
...
+
+    @Override
+    public List<String> getPrincipals(int queriedState) {
+        final String caller = "getPrincipals";
+        List<String> principals = new ArrayList<>();
+        try (PreparedStatement ps = con.prepareStatement(SQL_GET_PRINCIPAL)) {
+            ps.setInt(1, queriedState);
+            try (ResultSet rs = executeQuery(ps, caller)) {
+                while (rs.next()) {
+                    principals.add(rs.getString(1));
```
havetisyan, line 5636: prefer using the column names since it's easier to add more fields later and not worry about the order

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -6402,6 +6403,147 @@ void enforceGroupUserAuthorityRestrictions(final String domainName, final String
...
+
+    /**
+     * This method toggles state for supplied Principals based on the flag in DB
+     * as well as modifies memberships of all roles and groups of current principal(s)
+     * @param changedPrincipals List of Principals from User Authority
+     * @param suspended boolean indicating principal's state
+     */
+    void updatePrincipalByStateFromAuthority(List<Principal> changedPrincipals, boolean suspended) {
+        final String caller = "updatePrincipalByStateFromAuthority";
+        List<Principal> updatedUsers = new ArrayList<>();
```
havetisyan, line 6434: this should be within the if block so we're not wasting create the listing if there is nothing to do. or better yet in order not to have the whole function as we have now within an if block, it's simpler just do the check up front and return and that makes it easier to follow:

if (chargedPrincipals.isEmpty()) {
  return;
}

// continue with the rest of the method

final String caller = "updatePrincipalByStateFromAuthority";
List<Principal> updatedUsers = new ArrayList<>();
...

servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -479,6 +479,8 @@
             + "JOIN principal ON principal.principal_id=principal_group_member.principal_id "
             + "JOIN domain ON domain.domain_id=principal_group.domain_id "
             + "WHERE principal_group_member.last_notified_time=? AND principal_group_member.server=?;";
+    private static final String SQL_UPDATE_PRINCIPAL = "UPDATE principal SET system_suspended=? WHERE principal_id=?;";
```
havetisyan, line 482: need to update this command to use the name field instead of principal_id

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -6402,6 +6403,147 @@ void enforceGroupUserAuthorityRestrictions(final String domainName, final String
...
+        if (!domainRoleMember.getMemberRoles().isEmpty()) {
+            for (MemberRole memberRole : domainRoleMember.getMemberRoles()) {
+                roleMember = new RoleMember();
+                roleMember.setMemberName(updatedUser.getFullName());
+                if (suspended) {
+                    roleMember.setSystemDisabled(memberRole.getSystemDisabled() & ~Principal.State.AUTHORITY_SYSTEM_SUSPENDED.getValue());
+                } else {
+                    roleMember.setSystemDisabled(memberRole.getSystemDisabled() | Principal.State.AUTHORITY_SYSTEM_SUSPENDED.getValue());
+                }
+                roleMembersWithUpdatedState = new ArrayList<>();
```
havetisyan, line 6512: no need to create an array list with one member all the time. Instead use Collections.singletonList()

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -6402,6 +6403,147 @@ void enforceGroupUserAuthorityRestrictions(final String domainName, final String
...
+                roleMembersWithUpdatedState.add(roleMember);
+
+                // Following method does Audit entry as well
+                if (updateRoleMemberDisabledState(null, con, roleMembersWithUpdatedState, memberRole.getDomainName(),
+                        memberRole.getRoleName(), ZMSConsts.SYS_AUTH_MONITOR, AUDIT_REF, caller)) {
+                    updatedRoles.add(memberRole);
```
havetisyan, line 6518: the use of the this does not look correct since we're still processing the domain multiple times. if the principal is part of 10 roles in a domain, then we're updating the role mod timestamp for 10 roles, but then we're doing the same thing for the same domain 10 times.

instead of painting a updatedRoles struct, we should maintain:

Set<String> updatedDomains;

and then in the function we just call

con.updateRoleModTimestamp(memberRole.getDomainName(), memberRole.getRoleName());
updatedDomains.put(memberRole.getDomainName()); // this will ignore all duplicates

then outside of the for loop you can just iterate through the domain names and call

con.updateDomainModTimestamp(updatedMR.getDomainName());
cacheStore.invalidate(updatedMR.getDomainName());

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -6402,6 +6403,147 @@ void enforceGroupUserAuthorityRestrictions(final String domainName, final String
...
+                con.updateRoleModTimestamp(updatedMR.getDomainName(), updatedMR.getRoleName());
+                con.updateDomainModTimestamp(updatedMR.getDomainName());
+                cacheStore.invalidate(updatedMR.getDomainName());
+            }
+        }
+    }
+
+    private void updatePrincipalState(boolean suspended, List<Principal> updatedUsers, ObjectStoreConnection con, Principal changedPrincipal) {
+        try {
+            if (suspended) {
```
havetisyan, line 6533: we should simplify this by calling the updatePrincipal only once. 

newState = suspended ? Principal.State.AUTHORITY_SYSTEM_SUSPENDED.getValue() : Principal.State.ACTIVE.getValue();
if (con.updatePrincipal(changedPrincipal.getFullName(), newState)) {
    updatedUsers.add(changedPrincipal);
}

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -6402,6 +6403,147 @@ void enforceGroupUserAuthorityRestrictions(final String domainName, final String
...
+                } else {
+                    groupMember.setSystemDisabled(currentGroup.getSystemDisabled() | Principal.State.AUTHORITY_SYSTEM_SUSPENDED.getValue());
+                }
+                groupMembersWithUpdatedState = new ArrayList<>();
+                groupMembersWithUpdatedState.add(groupMember);
+
+                // Following method does Audit entry as well
+                if (updateGroupMemberDisabledState(null, con, groupMembersWithUpdatedState, currentGroup.getDomainName(),
+                        currentGroup.getGroupName(), ZMSConsts.SYS_AUTH_MONITOR, AUDIT_REF, caller)) {
+                    updatedGroups.add(currentGroup);
```
havetisyan, line 6483: see the comment in the role section which applies here as well.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -6402,6 +6403,147 @@ void enforceGroupUserAuthorityRestrictions(final String domainName, final String
...
+        GroupMember groupMember;
+        DomainGroupMember domainGroupMember;
+        Set<GroupMember> updatedGroups = new HashSet<>();
+
+        domainGroupMember = con.getPrincipalGroups(updatedUser.getFullName(), null);
+        if (!domainGroupMember.getMemberGroups().isEmpty()) {
+            for (GroupMember currentGroup : domainGroupMember.getMemberGroups()) {
+                groupMember = new GroupMember();
+                groupMember.setMemberName(updatedUser.getFullName());
+                if (suspended) {
```
havetisyan, line 6472: let's call groupMember.setSystemDisabled only once.
we should determine the state and then just pass that to the method e.g. 

state = suspended ? a : b;
groupMember.setSystemDisabled(state);

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -6402,6 +6403,147 @@ void enforceGroupUserAuthorityRestrictions(final String domainName, final String
...
+                    groupMember.setSystemDisabled(currentGroup.getSystemDisabled() & ~Principal.State.AUTHORITY_SYSTEM_SUSPENDED.getValue());
+                } else {
+                    groupMember.setSystemDisabled(currentGroup.getSystemDisabled() | Principal.State.AUTHORITY_SYSTEM_SUSPENDED.getValue());
+                }
+                groupMembersWithUpdatedState = new ArrayList<>();
```
havetisyan, line 6477: no need to create an array list with one member all the time. Instead use Collections.singletonList()

havetisyan (changes requested):

havetisyan (approved):

PR #1159, opened 2020-10-25
Ignore ConditionalCheckFailedException when updating

OferLevi85:
notification server and time in records. This error will appear after we try to update a record
that was already upated by a different zts server.


havetisyan (changes requested):
was the AWSCLIV2.pkg checked in by error?

OferLevi85:
> was the AWSCLIV2.pkg checked in by error?

Yes it was :)

havetisyan (approved):
👍

PR #1162, opened 2020-10-26
docker changes for local run using dockerhub images

abvaidya:
new make targets for local docker from dockerhub images
ui port changes inside the container
addition of linux-pam package and default user for UserAuthority to work
Inclusion of UserAuthority and PrincipalAuthority in zms docker conf


havetisyan (approved):
👍

PR #1163, opened 2020-10-27
Prevent Cert Fail Notifications to Rebootstrapped Hosts

OferLevi85:


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSConsts.java
```diff
@@ -102,6 +102,8 @@
     public static final String ZTS_PROP_CERT_DYNAMODB_TABLE_NAME                = "athenz.zts.cert_dynamodb_table_name";
     public static final String ZTS_PROP_CERT_DYNAMODB_ITEM_TTL_HOURS            = "athenz.zts.cert_dynamodb_item_ttl_hours";
     public static final String ZTS_PROP_CERT_DYNAMODB_INDEX_CURRENT_TIME_NAME   = "athenz.zts.cert_dynamodb_index_current_time_name";
+    public static final String ZTS_PROP_CERT_DYNAMODB_INDEX_HOST_NAME           = "athenz.zts.cert_dynamodb_index_host_name";
```
OferLevi85, line 105: Need to create the index in DynamoDB before pushing these changes

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
```diff
@@ -310,23 +315,50 @@ private boolean isRecordUpdatedWithNotificationTimeAndServer(String lastNotified
...
+                .collect(Collectors.toList());
+
         return items;
     }
 
-    private List<Item> getUnrefreshedCertRecordsByDate(String provider, Index index, long yesterday, String unrefreshedCertDate) {
+    private boolean mostUpdatedHostRecord(Item recordToCheck) {
+        try {
+            // Query all records with the same hostName as recordToCheck
+            QuerySpec spec = new QuerySpec()
```
OferLevi85, line 332: For every record:
1. Get all records with the same hostName
2. Check which one is the most up to date (biggest currentTime).
3. Check if it is the same record we evaluate now. If it is we will send a notification if not we'll skip it.

havetisyan:
@OferLevi85 this implementation only addresses the dynamo db cert record store. how about the jdbc one? Rather than implementing the changes in the cert record store itself, can we move the logic out of it so we only have a single implementation that would apply to regardless what record store we're using and keep the record store logic simple?
Also, this might require indexing on the hostname field in the jdbc table as well. so we need to update the schema files accordingly.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/cert/CertRecordStoreConnection.java
```diff
@@ -73,8 +73,16 @@
     int deleteExpiredX509CertRecords(int expiryTimeMins);
 
     /**
-     * List all certificates that failed to refresh for more than three days.
-     * @param lastNotifiedServer
+     * Return all certificate records that failed to refresh after updating them with the current notification time and server.
+     * Implementers of this interface should note the following:
+     * 1. The records should be filtered according to the following criteria:
+     *      - The certificates should have a grace period in which no notifications will be sent. The default grace period is 2 weeks
```
OferLevi85, line 79: I mentioned the default grace period is 2 weeks as this will be the case after the "grace period" PR will be merged

PR #1165, opened 2020-10-28
local athenz start and stop scripts

abvaidya:

### DIFF
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -110,3 +110,11 @@ clean: remove-all
 	docker image rm athenz-conf || true
 	docker image rm athenz-setup-env || true
 	docker image rm rdl-athenz-java-model || true
+
+remove-local: remove-all
+	docker image rm abvaidya/athenz-setup-env:1.9.22 || true
+	docker image rm abvaidya/athenz-zms-db:1.9.22 || true
+	docker image rm abvaidya/athenz-zms-server:1.9.22 || true
+	docker image rm abvaidya/athenz-zts-db:1.9.22 || true
+	docker image rm abvaidya/athenz-zts-server:1.9.22 || true
+	docker image rm abvaidya/athenz-ui:1.9.22 || true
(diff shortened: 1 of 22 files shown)


docker/Makefile
```diff
@@ -110,3 +110,11 @@ clean: remove-all
 	docker image rm athenz-conf || true
 	docker image rm athenz-setup-env || true
 	docker image rm rdl-athenz-java-model || true
+
+remove-local: remove-all
+	docker image rm abvaidya/athenz-setup-env:1.9.22 || true
```
OferLevi85, line 115: Noticed your username (abvaidya) is part of the path. Is that intentional?

abvaidya: Yes. For now the docker images are uploaded to my personal docker account. Once we test it enough and agree , I can upload the same to Athenz docker account and change references to Athenz docker.

havetisyan: why not use something like $USER instead?

docker/zms/conf/zms.properties
```diff
@@ -9,7 +9,7 @@
 
 # Comma separated list of authority implementation classes to support
 # authenticating principals in ZMS
-athenz.zms.authority_classes=com.yahoo.athenz.auth.impl.PrincipalAuthority,com.yahoo.athenz.auth.impl.UserAuthority,com.yahoo.athenz.auth.oauth.OAuthCertBoundJwtAccessTokenAuthority,com.yahoo.athenz.auth.impl.CertificateAuthority
+athenz.zms.authority_classes=com.yahoo.athenz.auth.impl.PrincipalAuthority,com.yahoo.athenz.auth.impl.TestUserAuthority,com.yahoo.athenz.auth.oauth.OAuthCertBoundJwtAccessTokenAuthority,com.yahoo.athenz.auth.impl.CertificateAuthority
```
havetisyan, line 12: if this is a test image why have the other authorities? this might mislead others and use them as is and left behind the TestUserAuthority. If it's for local dev, then I expect TestUserAuthority should be the only one added.

havetisyan (approved):
👍

PR #1166, opened 2020-10-28
If both expration and review dates are set, start sending

OferLevi85:
notifications from the review date

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
@@ -331,8 +331,15 @@ public class JDBCConnection implements ObjectStoreConnection {
     private static final String SQL_UPDATE_PENDING_ROLE_MEMBERS_NOTIFICATION_TIMESTAMP = "UPDATE pending_role_member SET last_notified_time=?, server=? " +
             "WHERE DAYOFWEEK(req_time)=DAYOFWEEK(?) AND (last_notified_time IS NULL || last_notified_time < (CURRENT_TIME - INTERVAL ? DAY));";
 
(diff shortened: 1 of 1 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/store/jdbc/JDBCConnection.java
```diff
@@ -331,17 +331,27 @@
...
             "JOIN principal ON principal.principal_id=role_member.principal_id " +
             "JOIN domain ON domain.domain_id=role.domain_id " +
             "WHERE role_member.last_notified_time=? AND role_member.server=?;";
 
-    private static final String SQL_UPDATE_ROLE_MEMBERS_REVIEW_NOTIFICATION_TIMESTAMP = "UPDATE role_member SET review_last_notified_time=?, review_server=? " +
-            "WHERE review_reminder > CURRENT_TIME AND DATEDIFF(review_reminder, CURRENT_TIME) IN (0,1,7,14,21,28) AND (review_last_notified_time IS NULL || review_last_notified_time < (CURRENT_TIME - INTERVAL ? DAY));";
+    private static final String SQL_UPDATE_ROLE_MEMBERS_REVIEW_NOTIFICATION_TIMESTAMP =
+            "UPDATE role_member SET review_last_notified_time=?, review_server=? " +
+            "WHERE (" +
+                    "review_reminder > CURRENT_TIME AND (expiration is NULL) AND DATEDIFF(review_reminder, CURRENT_TIME) IN (0,1,7,14,21,28) AND " +
```
OferLevi85, line 353: We only send review notifications if expiry isn't set. If it is we'll send expiration notifications (if the review date is set it will be used to determine when to start sending the notifications but the content will be an "expiration notification" email and not a "review reminder" email)

havetisyan (approved):

PR #1169, opened 2020-11-01
Configurable Cert Failure Notification Grace Period

OferLevi85:
(default set to 2 weeks)

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/JDBCCertRecordStoreConnection.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/JDBCCertRecordStoreConnection.java
@@ -19,6 +19,7 @@ import java.sql.*;
 import java.util.*;
 import java.util.Date;
 
+import com.yahoo.athenz.zts.ZTSConsts;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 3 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/JDBCCertRecordStoreConnection.java
```diff
@@ -60,7 +65,7 @@
                 "GROUP BY hostName, provider, service" +
             ") AS m on (m.hostName=a.hostName AND m.provider=a.provider AND m.service=a.service AND a.currentTime=date_updated) " +
             "SET lastNotifiedTime=?, lastNotifiedServer=? " +
-            "WHERE a.currentTime < (CURRENT_DATE - INTERVAL 3 DAY) AND " +
+            "WHERE a.currentTime < (CURRENT_DATE - INTERVAL ? DAY) AND " +
```
havetisyan, line 68: The setting is in hours so we need to update DAY parameter.

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/JDBCCertRecordStoreConnection.java
OferLevi85: Right, this was lost when I did the merge... Fixing now

havetisyan (approved):
👍

PR #1172, opened 2020-11-03
added method to extract jwks uri from openid config + unit tests

havetisyan:

### DIFF
--- a/libs/java/auth_core/src/test/java/com/yahoo/athenz/auth/token/jwts/JwtsHelperTest.java
+++ b/libs/java/auth_core/src/test/java/com/yahoo/athenz/auth/token/jwts/JwtsHelperTest.java
@@ -31,6 +31,26 @@ public class JwtsHelperTest {
     }
 
     @Test
+    public void testExtractJwksNullUri() {
+
+        MockJwtsHelper.setResponseCode(200);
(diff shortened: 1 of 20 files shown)


libs/java/auth_core/src/test/java/com/yahoo/athenz/auth/token/jwts/JwtsHelperTest.java
```diff
@@ -0,0 +1,52 @@
...
+import org.testng.annotations.Test;
+import static org.testng.Assert.*;
+
+public class JwtsHelperTest {
+
+    @Test
+    public void testExtractJwksUri() {
+
+        MockJwtsHelper.setResponseCode(200);
+        MockJwtsHelper.setResponseBody("{\"token_endpoint\":\"https://localhost/oauth2/token\",\"jwks_uri\":\"https://localhost/oauth2/keys\"}");
```
OferLevi85, line 27: Is it possible for the attribute jwks_uri to exist but be empty? Or exist but be malformed (not in a URI structure)? If so we should probably add these test cases

havetisyan: added test case for null/empty uri. malformed - library doesn't care. it's up to the caller to verify if it's valid or not.

PR #1174, opened 2020-11-05
Added tests to get to 100% coverage in zts

OferLevi85:

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
@@ -267,13 +267,8 @@ public class DynamoDBCertRecordStoreConnection implements CertRecordStoreConnect
     public List<X509CertRecord> updateUnrefreshedCertificatesNotificationTimestamp(String lastNotifiedServer,
                                                                       long lastNotifiedTime,
                                                                       String provider) {
-        try {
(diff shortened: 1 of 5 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/DynamoDBCertRecordStoreConnection.java
```diff
@@ -267,13 +267,8 @@ public int deleteExpiredX509CertRecords(int expiryTimeMins) {
     public List<X509CertRecord> updateUnrefreshedCertificatesNotificationTimestamp(String lastNotifiedServer,
                                                                       long lastNotifiedTime,
                                                                       String provider) {
-        try {
-            List<Item> items = getUnrefreshedCertsRecords(lastNotifiedTime, provider);
```
OferLevi85, line 271: Removed the try / catch as the catch could never happen. The inner methods already catch the exceptions

servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
```diff
@@ -160,7 +160,7 @@ public CertFailedRefreshNotificationTask(InstanceCertManager instanceCertManager
             }
 
             String expiryTime =  getTimestampAsString(certRecord.getExpiryTime());
-            String hostName = (certRecord.getHostName() != null) ? certRecord.getHostName() : "";
```
OferLevi85, line 163: At this point we already check that the record is valid and so it could never be nulll

havetisyan (approved):
👍

PR #1178, opened 2020-11-08
Azure VM Service Identity Agent

havetisyan:


abvaidya:
I think we should consider the idea to create a go lib with common files like logutil, options, utils etc. between aws and azure sia.

abvaidya (approved):

PR #1179, opened 2020-11-09
role tags - generated code and schema changes

dvirguttman:


core/zms/src/main/rdl/Role.tdl
```diff
@@ -48,6 +48,7 @@ type RoleMeta Struct {
     String userAuthorityFilter (optional); //membership filtered based on user authority configured attributes
     String userAuthorityExpiration (optional); //expiration enforced by a user authority configured attribute
     Int32 groupExpiryDays (optional); //all groups in the domain roles will have specified max expiry days
+    Map<SimpleName,SimpleName> tags (optional); //key-value pair tags
```
havetisyan, line 51: After thinking about this a little bit more, we should make the value as CompoundName. This would give us to have a little bit more flexibility to have .'s in our names - e.g. prod.us-west-2, etc.

servers/zms/schema/updates/update-20201109.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS `zms_server`.`role_tags` (
+  `role_id` INT UNSIGNED NOT NULL,
+  `tag_key` VARCHAR(64) NOT NULL,
```
havetisyan, line 3: since the table itself is already called role_tags, I don't think we need to specify tag_key/tag_value as column names, just having key/value is sufficient.

core/zms/src/main/rdl/Role.tdl
havetisyan, line 51: If this is a map, how are we going to support multiple tags with the same key? I thought we made the index key as role_id+key+value so we can have multiple tags with the same key name -e.g. env=prod,env=stage, etc.

libs/go/zmscli/dump.go
```diff
@@ -182,7 +182,7 @@ func (cli Zms) dumpRoles(buf *bytes.Buffer, dn string) {
 	buf.WriteString(indentLevel1)
 	buf.WriteString("roles:\n")
 	members := true
-	roles, err := cli.Zms.GetRoles(zms.DomainName(dn), &members)
+	roles, err := cli.Zms.GetRoles(zms.DomainName(dn), &members, "", "") //TODO what should be the values here?
```
havetisyan, line 185: empty strings is the correct value. the client code will skip and treat them as nils

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2782,10 +2784,28 @@ public RoleList getRoleList(ResourceContext ctx, String domainName, Integer limi
             }
         }
 
+        // filter roles by their tags
```
havetisyan, line 2787: In this method when we ask for roles without any members, we generate a new list of roles. In that section we need to update to return the tags as well.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -5746,4 +5755,92 @@ Boolean nullIfDefaultValue(boolean flag, boolean defaultValue) {
     Integer nullIfDefaultValue(int value, int defaultValue) {
         return value == defaultValue ? null : value;
     }
+    
+    private void addTagsToRoles(Map<String, Role> roleMap, String domainName) {
```
havetisyan, line 5759: This could be a very expensive operation. If a domain has 100 roles (which is not that uncommon), we're going to make 100 sql commands - with just regular latency this will add up quick considerably.

Any reason why we're not just running a query to get all tags for a domain and then parse the results and update accordingly. The roles are already in a map so we can quickly update each role accordingly.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -557,11 +557,49 @@ boolean processRole(ObjectStoreConnection con, Role originalRole, String domainN
             processUpdateRoleMembers(con, originalRole, roleMembers, ignoreDeletes,
                     domainName, roleName, admin, auditRef, auditDetails);
         }
-        
+
+        //TODO : should we fail unsuccessful operation?
+        processRoleTags(role, roleName, domainName, originalRole, con);
```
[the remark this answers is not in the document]
dvirguttman: Noticed that too..
will fix it

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
dvirguttman: This is correct for getRoles.
But this method is still needed for getRole..

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/store/ObjectStoreConnection.java
```diff
@@ -188,4 +188,12 @@
     boolean updatePrincipal(String principal, int newState);
 
     List<String> getPrincipals(int queriedState);
+
+    boolean insertRoleTags(String roleName, String domainName, Map<String, StringList> roleTags);
+
+    boolean deleteRoleTags(String roleName, String domainName, Set<String> tagKeys);
+
+    Map<String, Map<String, StringList>> getDomainRoleTags(int domainId);
```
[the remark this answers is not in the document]
dvirguttman: It is in the interface mainly for testing and maybe future use..
What if the method will get domainName (and the query will be changed as well)?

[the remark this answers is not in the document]
havetisyan: there is no need to add functions in the interface just for testing only. once it's private, we can keep the domain id in the signature - there is no need for the domain name for a private method.

havetisyan (approved):

PR #1186, opened 2020-11-15
zms-cli role tags

dvirguttman:


servers/zms/schema/updates/update-20201109.sql
```diff
@@ -3,9 +3,7 @@ CREATE TABLE IF NOT EXISTS `zms_server`.`role_tags` (
   `key` VARCHAR(64) NOT NULL,
   `value` VARCHAR(64) NOT NULL,
   PRIMARY KEY (`role_id`, `key`, `value`),
-  INDEX `fk_role_tag_role_idx` (`role_id` ASC),
-  INDEX `tag_key_sec_idx` (`key` ASC),
-  INDEX `tag_value_sec_idx` (`value` ASC),
+  INDEX `tag_key_val_sec_idx` (`key` ASC, `value` ASC),
```
dvirguttman, line 6: I noticed that I had a mistake in previous schema..
secondary index should be `key`+`value` not index per column (btree) 
Also, role_id fk shouldn't be index but constraint only..

havetisyan (approved):

PR #1201, opened 2020-12-02
Add zts statuschecker

ssunorz:
## Motivation
We create a StatusChecker class for JDBCCertRecordStore and want to use it as a health check for the ZTS.

## Behavior this change
1. Set `athenz.zts.status_checker_factory_class=com.yahoo.athenz.zts.cert.impl.JDBCCertRecordStoreStatusCheckerFactory` in zts.properties.
2. Start the ZTS and MySQL.
3. Run `curl -k https://localhost:8443/zts/v1/status -v`, The response is below.
```
{"code":200,"message":"OK"}
```
4. Stop the MySQL.
5. Run `curl -k https://localhost:8443/zts/v1/status -v`, The response is below.
```
{"code":500,"message":"ResourceException (503): Communications link failure\n\nThe last packet sent successfully to the server was 0 milliseconds ago. The driver has not received any packets from the server."}
```

------

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/JDBCCertRecordStoreStatusChecker.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/JDBCCertRecordStoreStatusChecker.java
@@ -1,0 +1,45 @@
+/*
+ *  Copyright 2020 Verizon Media
+ *
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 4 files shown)


havetisyan:
@ssunorz just one comment about using such functionality in a production environment. typically ZTS does more than just issue x.509 certs - it also handles access tokens, access checks, etc. So if you have such status checker and your mysql server goes down, all your servers will effectively taken out of rotation since status will start returning failure. This is ok if you were only handling x.509 requests, but will also affect your authorization access token request which will work fine since they're not dependent on x.509 cert record store. At Verizon Media, we have a health check that runs every minute which verifies that we can get x.509 certs such that if mysql goes down, we're alerted right away. So something to consider before using such functionality in your production.

ssunorz:
@havetisyan Thank you for your comment.

PR #1216, opened 2020-12-18
updated crypki http signer to support different keys per provider

havetisyan:

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/crypki/ProviderSignerKey.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/crypki/ProviderSignerKey.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2018 Oath Inc.
+ *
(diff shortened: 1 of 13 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/crypki/ProviderSignerKeys.java
```diff
@@ -0,0 +1,40 @@
+/*
```
abvaidya, line 1: should be verizon media copyright

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/crypki/ProviderSignerKey.java
```diff
@@ -0,0 +1,40 @@
+/*
```
abvaidya, line 1: should be verizon media copyright

abvaidya (changes requested):

abvaidya (approved):

PR #1235, opened 2021-01-11
domain tags implementation

dvirguttman:


servers/zms/src/main/java/com/yahoo/athenz/zms/store/AthenzDomain.java
```diff
@@ -28,16 +30,18 @@
     private List<Group> groups;
     private List<Policy> policies;
     private List<ServiceIdentity> services;
+    private Map<String, StringList> tags;
```
havetisyan, line 33: why did we add tags as a field here? The domain object below already has the tags as a field since it's part of the meta data so I would expect that the domain object contains the tags when the AthenzDomain object was created thus there is no need for this extra field here.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5172,11 +5177,12 @@ SignedDomain retrieveSignedDomainData(final String domainName, long modifiedTime
             domainData.setTokenExpiryMins(athenzDomain.getDomain().getTokenExpiryMins());
         }
 
-        // set the roles, services, and groups
+        // set the roles, services, groups and tags
 
         domainData.setRoles(athenzDomain.getRoles());
         domainData.setServices(athenzDomain.getServices());
         domainData.setGroups(athenzDomain.getGroups());
+        domainData.setTags(athenzDomain.getTags());
```
havetisyan, line 5185: to follow up from the other comment I would have expected here just to have:

domainData.setTags(athenzDomain.getDomain().getTags());

servers/zms/src/main/java/com/yahoo/athenz/zms/store/AthenzDomain.java
dvirguttman: indeed, initially I didn't notice the domain object is included there.
Will fix it

havetisyan (approved):

PR #1236, opened 2021-01-11
[Review after PR #1235 is merged] Snooze cert fail

OferLevi85:

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
@@ -22,6 +22,8 @@ import com.yahoo.athenz.common.server.cert.X509CertRecord;
 import com.yahoo.athenz.common.server.dns.HostnameResolver;
 import com.yahoo.athenz.common.server.notification.*;
 import com.yahoo.athenz.common.server.util.ResourceUtils;
+import com.yahoo.athenz.zms.DomainData;
+import com.yahoo.athenz.zms.StringList;
 import com.yahoo.athenz.zts.ZTSConsts;
 import com.yahoo.athenz.zts.cert.InstanceCertManager;
 import com.yahoo.athenz.zts.store.DataStore;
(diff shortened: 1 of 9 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
```diff
@@ -115,6 +127,32 @@ public CertFailedRefreshNotificationTask(InstanceCertManager instanceCertManager
...
+        Map<String, StringList> domainTags = domainData.getTags();
+        if (domainTags == null || domainTags.get(SNOOZED_DOMAIN_TAG_KEY) == null || domainTags.get(SNOOZED_DOMAIN_TAG_KEY).getList() == null) {
+            return false;
+        }
+
+        List<String> snoozeTagValues = domainTags.get(SNOOZED_DOMAIN_TAG_KEY).getList();
+        return snoozeTagValues.stream()
+                .filter(value -> value.toLowerCase().equals(SNOOZED_DOMAIN_TAG_VALUE))
+                .findAny()
+                .isPresent();
```
dvirguttman, line 153: can be replaced by `anyMatch(value -> value.toLowerCase().equals(SNOOZED_DOMAIN_TAG_VALUE))`

servers/zts/src/test/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTaskTest.java
```diff
@@ -414,6 +416,107 @@ public void testNoValidServices() {
...
+
+        for (int i = 0; i < 8; ++i) {
+            X509CertRecord record = getMockX509CertRecord(currentDate, i);
+            records.add(record);
+            NotificationTestsCommon.mockDomainData(i, dataStore);
+            DomainData domainData = new DomainData();
+            // Make even domains snooze
+            if (i % 2 == 0) {
+                Map<String, StringList> tags = new HashMap<>();
+                tags.put("DisableCertRefreshNotification", new StringList().setList(Arrays.asList("true")));
```
dvirguttman, line 434: pls use the constant value..

servers/zts/src/main/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTask.java
OferLevi85: Nice wasn't familiar with it thanks!

servers/zts/src/test/java/com/yahoo/athenz/zts/notification/CertFailedRefreshNotificationTaskTest.java
OferLevi85: I prefer not to make the const public to make it available in test as this const is only relevant for this class.
Also, if the value used in the class will change in the future (which might break domains that used the old value) I prefer to know about it by having this test fail

PR #1255, opened 2021-01-20
limit number of role and domain tags

dvirguttman:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
@@ -497,6 +497,7 @@ public class JDBCConnection implements ObjectStoreConnection {
     private static final String SQL_GET_PRINCIPAL = "SELECT name FROM principal WHERE system_suspended=?;";
     private static final String SQL_INSERT_ROLE_TAG = "INSERT INTO role_tags"
             + "(role_id, role_tags.key, role_tags.value) VALUES (?,?,?);";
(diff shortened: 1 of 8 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -538,6 +543,8 @@ public JDBCConnection(Connection con, boolean autoCommit) throws SQLException {
         con.setAutoCommit(autoCommit);
         transactionCompleted = autoCommit;
         objectMap = new HashMap<>();
+        roleTagsLimit = Integer.getInteger(ZMSConsts.ZMS_PROP_QUOTA_ROLE_TAG, 25);
```
havetisyan, line 546: we shouldn't keep fetching the system property for every new connection. Instead these values should be fetched as part of the JDBCObjectStore should fetch them as part of its constructor and then set the when a new connection is returned - something similar to the setOperationTimeout method the jdbc connection has. This way we only fetch the settings once and not over and over again.

havetisyan (changes requested):
Looks like the unit tests are failing as well. didn't look into them in detail but It could be because the test is getting a connection object from the pool and not a new one as such the limits are not set to the lower value.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
dvirguttman: sure. originally it was static const.. forgot that this is a heavy created object..

PR #1277, opened 2021-02-03
domain tags ui

dvirguttman:


ui/src/components/header/Tabs.js
```diff
@@ -47,6 +47,10 @@ class Tabs extends React.Component {
             label: 'History',
             name: 'history',
         },
+        {
```
chandrasekhar1996, line 50: I think we are planning to add features to set user authority filters and expiration later on, in that case having a settings tab would be better instead of tags

ui/src/components/tag/AddTagForm.js
```diff
@@ -0,0 +1,223 @@
...
+        background: #ffffff;
+    }
+    font-size: 14px;
+    height: 28px;
+    line-height: 14px;
+    margin: 5px 15px 5px 0;
+    padding: 6px 8px 7px 10px;
+`;
+
+const StyledTagColor = styled(StyledTag)`
```
chandrasekhar1996, line 75: we should add this style in line 64 instead of styling our styled component

ui/src/components/tag/AddTag.js
```diff
@@ -0,0 +1,98 @@
...
+                errorMessage: 'Tag value is required.',
+            });
+            return;
+        }
+
+        this.props.addNewTag(this.state.tagName, this.state.tagValues);
+    }
+
+    onUpdate(tagKey, tagValues) {
+        this.state.tagName = tagKey;
```
chandrasekhar1996, line 69: we should update state variables using this.setState

ui/src/components/tag/AddTagForm.js
```diff
@@ -0,0 +1,223 @@
...
+    background: rgba(53, 112, 244, 0.08);
+`;
+
+const StyledAnchor = styled.a`
+    text-decoration: none;
+`;
+
+const StyledAnchorActiveInline = { color: colors.linkActive };
+
+export default class AddTagForm extends React.Component {
```
chandrasekhar1996, line 85: any reason we want to maintain state variables in this component?
a suggestion would be to just use this to render the form and use the props to populate values because currently you are doing setState here on input changed and again doing a setState in AddTag.js which is repetitive.

ui/src/components/tag/AddTagForm.js
```diff
@@ -0,0 +1,223 @@
...
+
+    inputChanged(key, evt) {
+        let value = '';
+        if (evt.target) {
+            value = evt.target.value;
+        } else {
+            value = evt ? evt : '';
+        }
+        this.setState({ [key]: value }, () =>
+            this.props.onUpdate(this.state.tagName, this.state.tagValues)
```
chandrasekhar1996, line 114: we are not updating the tagValue on inputChanged because you are passing this.state.tagValue, whereas the input changed is this.state.newTagValue.

ui/src/components/tag/AddTag.js
```diff
@@ -0,0 +1,98 @@
...
+            <AddModal
+                isOpen={this.state.showModal}
+                cancel={this.props.onCancel}
+                submit={this.onSubmit}
+                title={title}
+                errorMessage={
+                    this.props.errorMessage || this.state.errorMessage
+                }
+                sections={
+                    <AddTagForm
```
chandrasekhar1996, line 88: refer to 
https://github.com/AthenZ/athenz/pull/1277/files?file-filters%5B%5D=.js#r569955596

ui/src/components/tag/AddTagForm.js
```diff
@@ -0,0 +1,223 @@
...
+                </SectionDiv>
+                <SectionDiv>
+                    <StyledInputLabel />
+                    <ButtonDiv>
+                        <StyledButton
+                            secondary
+                            size={'small'}
+                            onClick={this.addTagValue}
+                        >
+                            Add
```
chandrasekhar1996, line 210: let's have the add button next to the tag value field instead of on the next line

ui/src/components/header/RoleTabs.js
```diff
@@ -43,6 +43,10 @@ class RoleTabs extends React.Component {
             label: 'History',
             name: 'history',
         },
+        {
```
chandrasekhar1996, line 46: let's keep the ordering of the tabs consistent with the ordering of the columns in RoleTable

ui/src/components/tag/TagList.js
```diff
@@ -0,0 +1,483 @@
...
+                showDeleteTagValue: true,
+                deleteTagName: tagKey,
+                deleteTagValue: tagValue,
+                errorMessage: null,
+            });
+        }
+    }
+
+    validateTagExist(tagName) {
+        return this.state.tags[tagName] != null;
```
chandrasekhar1996, line 153: make sure that this.state.tags is not undefined

ui/src/components/tag/TagList.js
```diff
@@ -0,0 +1,483 @@
...
+                    this.setState({
+                        errorMessage: RequestUtils.xhrErrorCheckHelper(err),
+                    });
+                });
+        }
+    }
+
+    updateStateAfterReload(data, successMessage) {
+        this.setState({
+            tags: data.tags,
```
chandrasekhar1996, line 332: tags will be undefined after deleting last tag

ui/src/components/tag/TagList.js
```diff
@@ -0,0 +1,483 @@
...
+            <TagsSectionDiv data-testid='tag-list'>
+                <AddContainerDiv>
+                    <div>
+                        <Button secondary onClick={this.openAddTag}>
+                            Add Tag
+                        </Button>
+                        {addTag}
+                    </div>
+                </AddContainerDiv>
+                <TagTable>
```
chandrasekhar1996, line 431: display the table only if there are tags present otherwise show a message like - Click on Add Tag to create a new tag

ui/src/components/tag/AddTagForm.js
```diff
@@ -0,0 +1,223 @@
...
+            : '';
+
+        return (
+            <SectionsDiv autoComplete={'off'} data-testid='add-tag-form'>
+                <SectionDiv>
+                    <StyledInputLabel htmlFor='tag-name'>
+                        Tag Name
+                    </StyledInputLabel>
+                    <ContentDiv>
+                        <StyledInput
```
chandrasekhar1996, line 179: if the current operation is to edit the tag, let us not show input for the tag name, instead show a text field with the tag name which won't be editable or disable this input field

ui/src/components/tag/TagList.js
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    validateTagExist(tagName) {
+        return this.state.tags[tagName] != null;
+    }
+
+    addNewTag(tagKey, tagValues) {
+        const csrf = this.props._csrf;
+        if (this.state.category === 'domain') {
+            this.api
```
chandrasekhar1996, line 159: instead of doing a getDomain call here, we can just pass the meta object from tags.js page.
Currently we are passing only the tags, instead if we pass the entire meta object we won't have to do another getDomain call

ui/src/components/tag/TagList.js
```diff
@@ -0,0 +1,483 @@
...
+                        : `Successfully added tag ${tagKey}`;
+                    this.updateMeta(domainMeta, csrf, successMessage);
+                })
+                .catch((err) => {
+                    this.setState({
+                        errorMessage: RequestUtils.xhrErrorCheckHelper(err),
+                    });
+                });
+        } else if (this.state.category === 'role') {
+            this.api
```
chandrasekhar1996, line 179: same as above

ui/src/components/tag/AddTagForm.js
dvirguttman: AddTagForm is AddTag child.
On every change in AddTagForm (add tag, delete tag, change tag name) we need to inform AddTag since the submitting done through AddTag.
As a result both should keep state and not `a suggestion would be to just use this to render the form and use the props to populate values`

ui/src/components/tag/AddTagForm.js
dvirguttman: This is relevant only for that:
`let tagNameChanged = this.inputChanged.bind(this, 'tagName');`
Anyway, deleted the inputChange for newTagValue. this is not relevant for parent component

ui/src/components/tag/AddTagForm.js
chandrasekhar1996: Since addTagForm is child of AddTag, we can just update the state variable of AddTag by calling this.props.function()

This would update the state value in AddTag.js and this is what we pass to AddTagForm.js so AddTagForm gets the updates through props.

ui/src/components/tag/AddTagForm.js
chandrasekhar1996: not sure I follow, 
the intended behavior in UI is:
user enters tag value in input field, but doesn't click on Add button. The user directly clicks on Submit button and it should work. This is similar to other add modals

ui/src/components/tag/AddTagForm.js
dvirguttman: other reason for keep it with state since addTagForm uses also for edit existing tag.
will check it again

ui/src/components/tag/AddTagForm.js
dvirguttman: Exactly.
See in AddTag:
`        if (!this.state.tagValues || this.state.tagValues.length === 0) {
            this.setState({
                errorMessage: 'Tag value is required.',
            });
            return;
        }`

ui/src/components/header/RoleTabs.js
dvirguttman: not sure what do you mean

ui/src/components/tag/TagList.js
dvirguttman: this is ok

ui/src/components/tag/AddTagForm.js
chandrasekhar1996: so when the input field of tag value is changed then in addTagForm.js we do a setState on the varibale "newTagValue" but we call this.props.onUpdate with the variable "tagValues" which is empty until you click on Add Button

we don't wan't this

ui/src/components/header/RoleTabs.js
chandrasekhar1996: in the domain page with roleTable, we have "tags" column after "policy-rule" and before "settings".

ui/src/components/tag/TagList.js
chandrasekhar1996: it throws an error, we should handle it similar to what we do in the constructor

ui/src/components/tag/AddTagForm.js
dvirguttman: In my first comment I wrote:
Anyway, deleted the inputChange for newTagValue. this is not relevant for parent component
So, we are updating the parent only on pressing Add button.

dvirguttman: Anyway, I think that we should not keep state of newTagValue

ui/src/components/tag/TagList.js
dvirguttman: Handled that in `render` but it is better

ui/src/components/tag/AddTagForm.js
dvirguttman: This is same pattern as add role member

ui/src/components/tag/AddTag.js
```diff
@@ -0,0 +1,124 @@
...
+                cancel={this.props.onCancel}
+                submit={() => this.onSubmit()}
+                title={title}
+                errorMessage={
+                    this.props.errorMessage || this.state.errorMessage
+                }
+                sections={
+                    <AddTagForm
+                        api={this.api}
+                        onUpdate={(tagKey, newTagValue, tagValues) =>
```
pratikgote, line 113: we should bind this.onUpdate in the constructor and pass it here

pratikgote: trying to avoid creating an anonymous function on every render

ui/src/components/tag/AddTagForm.js
```diff
@@ -0,0 +1,255 @@
...
+                            id='tag-name'
+                            name='tag-name'
+                            placeholder='Enter New Tag Name'
+                            value={this.state.tagName}
+                            onChange={(evt) =>
```
pratikgote, line 212: same here, this.inputChanged should be bind in constructor and pass it to this function to avoid creating anonymous functions on render

ui/src/components/tag/AddTagForm.js
```diff
@@ -0,0 +1,255 @@
...
+                        Tag Value(s)
+                    </StyledInputLabelPadding>
+                    <ContentDiv>
+                        <AddTagValueDiv>
+                            <StyledInputTag
+                                id='tag-val'
+                                name='tag-val'
+                                placeholder='Enter New Tag Value'
+                                value={this.state.newTagValue}
+                                onChange={(evt) =>
```
pratikgote, line 230: same as above

ui/src/components/tag/AddTagForm.js
```diff
@@ -0,0 +1,255 @@
...
+                                    this.inputChanged('newTagValue', evt)
+                                }
+                            />
+                            <ButtonDiv>
+                                <StyledButton
+                                    secondary
+                                    size={'small'}
+                                    onClick={() => this.addTagValue()}
```
pratikgote, line 238: same as above

ui/src/components/tag/TagList.js
```diff
@@ -0,0 +1,443 @@
...
+                editMode={this.state.editMode}
+                onCancel={() => this.closeAddTag()}
+                resource={
+                    this.state.category === 'domain'
+                        ? this.props.domain
+                        : this.props.role
+                }
+                api={this.api}
+                _csrf={this.props._csrf}
+                addNewTag={(tagName, tagValues) =>
```
pratikgote, line 366: same as above, avoiding anonymous function creation in render

pratikgote (changes requested):

ui/src/components/tag/AddTag.js
dvirguttman: Actually, I've changed all new tag components not to use bind.
1. The bind syntax is unreadable.
2. Performance is better (at least not worse than bind).

chandrasekhar1996 (changes requested):
need to add tag columns for expanded roles as well (RoleSectionRow.js)

abvaidya:
Please change the copyright on all the new files from "COPYRIGHT 2021 Verizon Media." To "COPYRIGHT The Athenz Authors"

This should be the go-forward copyright notice recommended by CNCF.

ui/src/components/role/RoleSectionRow.js
```diff
@@ -187,6 +187,10 @@ class RoleSectionRow extends React.Component {
             this,
             `/domain/${this.props.domain}/role/${this.state.name}/policy`
         );
+        let clickTag = this.onClickFunction.bind(
```
pratikgote, line 190: we should also move these to the constructor, I don't see any dynamic state value associated with these

pratikgote (approved):
👍

ui/src/components/tag/AddTagForm.js
chandrasekhar1996: did you try removing state and moving to props?

chandrasekhar1996: following up on this..tried moving to props from state ?

PR #1306, opened 2021-02-20
Log ssl connections

OferLevi85:


containers/jetty/pom.xml
```diff
@@ -180,6 +180,18 @@
       <groupId>com.amazonaws</groupId>
       <artifactId>aws-java-sdk-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>uk.co.datumedge</groupId>
```
OferLevi85, line 184: Added new dependencies in test scope only

containers/jetty/src/main/java/com/yahoo/athenz/container/log/JettyConnectionLogger/ConnectionLogEntry.java
```diff
@@ -0,0 +1,195 @@
...
+import java.time.Instant;
+import java.util.Optional;
+import java.util.UUID;
+
+public class ConnectionLogEntry {
+    private final UUID id;
+    private final Instant timestamp;
+    private final Double durationSeconds;
+    private final String peerAddress;
+    private final Integer peerPort;
```
OferLevi85, line 30: We only record failed connections so some attributes used in Vespa solution were removed as they are only valid for successfully SSL handshakes (such bytes received / number of requests / agreed protocol etc...)

containers/jetty/src/main/java/com/yahoo/athenz/container/log/JettyConnectionLogger/JettyConnectionLogger.java
```diff
@@ -0,0 +1,284 @@
...
+            LOGGER.error(failure.getMessage());
+            ConnectionInfo info = sslToConnectionInfo.remove(sslEngine);
+            info.setSslHandshakeFailure(failure);
+        });
+    }
+    //
+    // SslHandshakeListener methods end
+    //
+
+    private void handleListenerInvocation(
```
OferLevi85, line 152: This method wraps the other "event listeners" in this class and writes the event to the main log (server.log) in debug mode. For example "DEBUG c.y.a.c.l.J.JettyConnectionLogger - Connection.Listener.onClosed(597fea07)".
We might want to consider removing it in the future if it will generate too much "noise".

containers/jetty/src/main/java/com/yahoo/athenz/container/log/JettyConnectionLogger/JettyConnectionLogger.java
```diff
@@ -0,0 +1,284 @@
...
+    // AbstractLifeCycle methods stop
+    //
+
+    //
+    // Connection.Listener methods start
+    //
+    @Override
+    public void onOpened(Connection connection) {
+        handleListenerInvocation("Connection.Listener", "onOpened", "%h", Stream.of(connection).collect(Collectors.toList()), () -> {
+            SocketChannelEndPoint endpoint = findUnderlyingSocketEndpoint(connection.getEndPoint());
```
OferLevi85, line 88: When a connection is opened, get the connection details and put them in the "ConnectionInfo" map. They will be removed from the map when the connection closes. The details of the connection will only be written to log in case of handshake failure (handshakeFailed was called with the exception)

abvaidya:
Please change copyright notices for all new files from "Copyright Verizon Media 2021 All rights reserved" to "Copyright The Athenz Authors".

abvaidya (changes requested):
Copyright needs to be changed

OferLevi85:
> Please change copyright notices for all new files from "Copyright 2020 Verizon Media" to "Copyright The Athenz Authors".

Fixed

containers/jetty/pom.xml
```diff
@@ -180,6 +180,18 @@
...
+      <artifactId>hamcrest-json</artifactId>
+      <version>0.2</version>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-client</artifactId>
+      <version>9.4.36.v20210114</version>
```
havetisyan, line 192: Instead of harding the version here we should use ${jetty.version}

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -458,28 +467,62 @@ public void addHTTPConnectors(HttpConfiguration httpConfig, int httpPort, int ht
...
+                    + " error: " + e.getMessage());
+            throw new IllegalArgumentException("Invalid metric class");
+        }
+
+        return metricFactory.create();
+    }
+
+    private ConnectionLog getSslConnectionLog() {
+        SSLConnectionLogFactory sslConnectionLogFactory;
+        String sslConnectionLogFactoryClass = System.getProperty(AthenzConsts.ATHENZ_PROP_SSL_LOGGER_FACTORY_CLASS,
```
havetisyan, line 514: let's define this as final

havetisyan (changes requested):
a couple of minor fixes

havetisyan (approved):

PR #1308, opened 2021-02-23
Example set up on docker in AWS EC2 

abvaidya:


docker/amazon-trust.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env sh
+
+if test -e /etc/pki/tls/certs/ca-bundle.crt; then
+  BASE_DIR=$(git rev-parse --show-toplevel)
+  mkdir -p "${BASE_DIR}/docker/zts/conf/awscas"
+  cd "${BASE_DIR}/docker/zts/conf/awscas" || true
```
havetisyan, line 6: any reason why not just copy the files into the conf as amazon1.pem, etc instead of using awscas and then splitting them here?

havetisyan (approved):

docker/amazon-trust.sh
abvaidya: files are read from the pem bundle which contains all public CAs and not just amazon CAs. ZTS requires everything in a truststore bundle, so I read only amazon cas and then store them as separate files in "awscas" dir. 

and in zts-auto-config.sh these 4 files are imported into the existing truststore.

abvaidya: in the next PR I will add amazon CA in the conf dir and remove the brittle logic of reading it from local PEM bundle.

PR #1325, opened 2021-03-07
Rewrite SSL Connection Logger

OferLevi85:

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
@@ -19,36 +19,27 @@
 package com.yahoo.athenz.common.server.log.jetty;
 
 import com.yahoo.athenz.common.metrics.Metric;
-import org.eclipse.jetty.io.Connection;
-import org.eclipse.jetty.io.EndPoint;
-import org.eclipse.jetty.io.SocketChannelEndPoint;
-import org.eclipse.jetty.io.ssl.SslConnection;
 import org.eclipse.jetty.io.ssl.SslHandshakeListener;
-import org.eclipse.jetty.server.*;
 import org.eclipse.jetty.util.component.AbstractLifeCycle;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import javax.net.ssl.SSLEngine;
 import javax.net.ssl.SSLHandshakeException;
-import java.net.InetSocketAddress;
 import java.time.Instant;
-import java.util.*;
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
+import java.util.Collections;
+import java.util.List;
+import java.util.UUID;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 import static com.yahoo.athenz.common.server.log.jetty.ExceptionCauseFetcher.getInnerCause;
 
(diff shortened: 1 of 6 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
```diff
@@ -79,83 +70,21 @@ protected void doStart() {
...
     @Override
     public void handshakeFailed(Event event, Throwable failure) {
         SSLEngine sslEngine = event.getSSLEngine();
         handleListenerInvocation("SslHandshakeListener", "handshakeFailed", "sslEngine=%h,failure=%s", Stream.of(sslEngine, failure).collect(Collectors.toList()), () -> {
-            if (LOGGER.isDebugEnabled()) {
-                LOGGER.debug(failure.getMessage());
-            }
-            ConnectionInfo info = sslToConnectionInfo.remove(sslEngine);
+            ConnectionInfo info = ConnectionInfo.from(sslEngine);
             if (info != null) {
```
havetisyan, line 78: when would info be null? looks like we're always going to get a valid info object so we should be able to remove this if check.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
```diff
@@ -79,83 +70,21 @@ protected void doStart() {
...
                 info.setSslHandshakeFailure(failure);
             }
+
+            if (info.sslHandshakeFailureException != null) {
```
havetisyan, line 82: in this method I think we're always going to have the exception check so we should able to remove this check.

havetisyan (changes requested):
A couple of other changes we should make as part of this PR:

a) JsonConnectionLogWriter - lines 41/42 - can we use try with resources for OutputStream so we don't need a finally block

b) JsonConnectionLogWriter - line 38 LOG is not used so let's remove it from the class

c) SSLConnectionLog - let's combine lines 34/35 - no need for a local variable

d) JettyConnectionLogger - where we have "this.sslHandshakeFailureMessage == GENERAL_SSL_ERROR" - it's probably safer to replace this with GENERAL_SSL_ERROR.equals(this.sslHandshakeFailureMessage)

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
OferLevi85: fixed

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
OferLevi85: fixed

OferLevi85:
> A couple of other changes we should make as part of this PR:
> 
> a) JsonConnectionLogWriter - lines 41/42 - can we use try with resources for OutputStream so we don't need a finally block
> 
> b) JsonConnectionLogWriter - line 38 LOG is not used so let's remove it from the class
> 
> c) SSLConnectionLog - let's combine lines 34/35 - no need for a local variable
> 
> d) JettyConnectionLogger - where we have "this.sslHandshakeFailureMessage == GENERAL_SSL_ERROR" - it's probably safer to replace this with GENERAL_SSL_ERROR.equals(this.sslHandshakeFailureMessage)

fixed

havetisyan (approved):

PR #1332, opened 2021-03-08
Enhance security:

gilad-bendor:
- OCSP-Stapling enabled by default
  - Client re-negotiation disabled by default
  - Any response headers can be configured. Suggestion for better security: set the system-property "athenz.response_headers_json" to something like this:
       { "Expect-CT": "max-age=31536000, report-uri=\"http://........\"", "Strict-Transport-Security": "max-age=31536000", "X-Content-Type-Options": "nosniff", "Referrer-Policy": "strict-origin-when-cross-origin", "Cache-Control": "must-revalidate,no-cache,no-store" }

Signed-off-by: Gilad Ben-Dor <gilad.bendor@verizonmedia.com>

### DIFF
--- a/containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
+++ b/containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
@@ -18,10 +18,15 @@ package com.yahoo.athenz.container;
 
 import java.io.File;
 import java.net.InetAddress;
+import java.security.Security;
 import java.util.EnumSet;
+import java.util.HashMap;
+import java.util.Map;
 
 import javax.servlet.DispatcherType;
 
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.yahoo.athenz.common.server.log.jetty.JettyConnectionLogger;
 import com.yahoo.athenz.common.server.log.jetty.JettyConnectionLoggerFactory;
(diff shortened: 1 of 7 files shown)


containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -389,7 +415,15 @@ void loadServicePrivateKey() {
...
         }
 
-        sslContextFactory.setEnableOCSP(enableOCSP);
+        if (enableOCSP) {
+            // See https://stackoverflow.com/questions/49904935/jetty-9-enable-ocsp-stapling-for-domain-validated-certificate
+            Security.setProperty("ocsp.enable", "true");
+            System.setProperty("jdk.tls.server.enableStatusRequestExtension", "true");
+            System.setProperty("com.sun.net.ssl.checkRevocation", "true");
+            sslContextFactory.setEnableOCSP(true);
+            sslContextFactory.setValidateCerts(true);   // not sure what this does... OCSP works even without this - but it is recommended.
```
havetisyan, line 424: other than sslContextFactory.setEnableOCSP(true); I'm not sure why we have the other settings or why we're setting them in the first place. So let's remove those from our code. if the organization wants to enable those, they can do so by adding them to the server startup script as just regular system properties.

havetisyan (changes requested):

gilad-bendor:
Merge hell. Closing pull-request.
Resurrected as https://github.com/AthenZ/athenz/pull/1353

PR #1356, opened 2021-03-16
Log client cert on handshake failure

OferLevi85:


containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -368,7 +368,7 @@ void loadServicePrivateKey() {
         final String excludedProtocols = System.getProperty(AthenzConsts.ATHENZ_PROP_EXCLUDED_PROTOCOLS, ATHENZ_DEFAULT_EXCLUDED_PROTOCOLS);
         final boolean renegotiationAllowed = Boolean.parseBoolean(System.getProperty(AthenzConsts.ATHENZ_PROP_RENEGOTIATION_ALLOWED, "false"));
 
-        SslContextFactory.Server sslContextFactory = new SslContextFactory.Server();
```
OferLevi85, line 371: Override SslContextFactory.Server with our own version that print certificates on failures

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzServer.java
```diff
@@ -0,0 +1,107 @@
...
+            if (X509TrustManager.class.isAssignableFrom(trustManagers[i].getClass())) {
+                X509TrustManagerWithLogs x509TrustManagerWithLogs = new X509TrustManagerWithLogs((X509TrustManager) trustManagers[i]);
+                trustManagers[i] = x509TrustManagerWithLogs;
+            }
+        }
+
+        return trustManagers;
+    }
+
+    public static class X509TrustManagerWithLogs implements X509TrustManager {
```
OferLevi85, line 53: X509TrustManagerWithLogs wrapps the X509TrustManager with the same methods only it print the certs on error

dvirguttman: Can we print it to sslcon.log?

OferLevi85: It does print it to sslconn.log. You'll notice the log is initiated like so:
    private static final Logger LOG = LoggerFactory.getLogger("SSLConnectionLog");

PR #1358, opened 2021-03-16
For SSL errors, always display cause if it's different

OferLevi85:
than the error message

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
@@ -131,8 +131,12 @@ public class JettyConnectionLogger extends AbstractLifeCycle implements SslHands
             this.sslHandshakeFailureException = sslHandshakeException.getClass().getName();
             this.sslHandshakeFailureMessage = sslHandshakeException.getMessage();
             // If the error isn't clear, try to get it from the exception inner cause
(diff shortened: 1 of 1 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/log/jetty/JettyConnectionLogger.java
```diff
@@ -131,8 +131,12 @@ synchronized ConnectionInfo setSslHandshakeFailure(Throwable exception) {
             this.sslHandshakeFailureException = sslHandshakeException.getClass().getName();
             this.sslHandshakeFailureMessage = sslHandshakeException.getMessage();
             // If the error isn't clear, try to get it from the exception inner cause
-            if (exception.getCause() != null && GENERAL_SSL_ERROR.equals(this.sslHandshakeFailureMessage)) {
+            if (exception.getCause() != null) {
                 this.sslHandshakeFailureCause = getInnerCause(exception, sslHandshakeException.getMessage());
+                // If the cause is identical to the message, no need to print it so we'll set it to null
+                if (this.sslHandshakeFailureCause.equals(this.sslHandshakeFailureMessage)) {
```
havetisyan, line 137: are we absolutely sure that this.sslHandshakeFailureCause will never be null? we might be want be safe and do proper checks?

OferLevi85: > are we absolutely sure that this.sslHandshakeFailureCause will never be null? we might be want be safe and do proper checks?

That would mean that one of the Jetty exceptions has a null message which I don't think is likely but I added the check to be on the safe side.

havetisyan (approved):

PR #1377, opened 2021-03-18
send error on exceeding tags limit

dvirguttman:


servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -1081,10 +1081,15 @@ public boolean insertDomainTags(String domainName, Map<String, StringList> tags)
             throw notFoundError(caller, ZMSConsts.OBJECT_DOMAIN, domainName);
         }
         int curTagCount = getDomainTagsCount(domainId);
-        int remainingTagsToInsert = domainTagsLimit - curTagCount;
+        int newTagCount = calculateTagCount(tags);
+        if (curTagCount + newTagCount > domainTagsLimit) {
```
chandrasekhar1996, line 1085: why are we adding the current tag count to new tag count ?

dvirguttman: For tag updates, we delete (if needed) then insert.
While calculating the quota, we should consider existing tags.

chandrasekhar1996: but this will be a put meta call which will have the updated list of tags (including the current tags or deleted tags)

chandrasekhar1996: 👍

chandrasekhar1996:
Currently, if the domain already has tags, you are calculating the tags that need to be removed (`tagsToRemove`) and remove those first before doing the insert operation where you have added the quota check.

We shouldn't delete the tags before knowing whether the new tag count will exceed the quota.
Can't we have a `getTagLimit` method in JDBCObjectStore that can be used to check the quota in DBService?

dvirguttman:
> Currently, if the domain already has tags, you are calculating the tags that need to be removed (`tagsToRemove`) and remove those first before doing the insert operation where you have added the quota check.
> 
> We shouldn't delete the tags before knowing whether the new tag count will exceed the quota.
> Can't we have a `getTagLimit` method in JDBCObjectStore that can be used to check the quota in DBService?

The processing of domains and roles is done in the same transaction and are not auto-committed. 
In case of failure, we will not commit the changes.

Issue #1361, opened 2021-03-17
Support Json format for all zms-cli operations related to domain commands.


jothi-prasad:
Today [zms-cli](https://github.com/AthenZ/athenz/tree/master/utils/zms-cli) produces yaml output for any cli command like below.

```
$ ~/bin/zms-cli -d ${domain} list-role 
roles:
    - admin
    - aws.s3
    - aws_instance_launch_provider
```

add support to also provide json output with  a command line option "-o json"
Note : accept -o param before -d 
for example: the command should be "zms-cli -o json -d ${domain} list-role" 

```
$ ~/bin/zms-cli -o json -d ${domain} list-role
{"names":["admin","aws.s3","aws_instance_launch_provider"]}
```
ZMS API already provides the output in JSON

2 - This PR should include -o json option for given below domain commands
```
$ ~/bin/zms-cli help
 Domain commands:
   list-domain [prefix] | [limit skip prefix depth]
   show-domain [domain]
   lookup-domain-by-aws-account account-id
   lookup-domain-by-azure-subscription subscription-id
   lookup-domain-by-product-id product-id
   lookup-domain-by-role role-member role-name
   add-domain domain product-id [admin ...] - to add top level domains
   add-domain domain [admin ...] - to add sub domains
   set-domain-meta description
   set-audit-enabled audit-enabled
   set-aws-account account-id
   set-azure-subscription subscription-id
   set-product-id product-id
   set-application-id application-id
   set-org-name org-name
   set-cert-dns-domain cert-dns-domain
   set-domain-member-expiry-days user-member-expiry-days
   set-domain-service-expiry-days service-member-expiry-days
   set-domain-group-expiry-days group-member-expiry-days
   set-domain-token-expiry-mins token-expiry-mins
   set-domain-service-cert-expiry-mins cert-expiry-mins
   set-domain-role-cert-expiry-mins cert-expiry-mins
   set-domain-token-sign-algorithm algorithm
   set-domain-user-authority-filter filter
   import-domain domain [file.yaml [admin ...]] - no file means stdin
   export-domain domain [file.yaml] - no file means stdout
   delete-domain domain
   get-signed-domains [matching_tag]
   use-domain [domain]
   check-domain [domain]
   get-quota
   set-quota [attrs ...]
   delete-quota
   overdue-review [domain]

patrasap0908:
Hi @abvaidya, okay for me to take this one up?

es-karthic:
Can I claim this one

PR #1386, opened 2021-03-21
Support JSON output format for domain-related commands

patrasap0908:
This PR adds support for JSON output for `zms-cli` for all domain-related commands, as described by https://github.com/AthenZ/athenz/issues/1361.

Few sample commands:
- `list-domain`:
```
(venv) ~/Projects/athenz : go run utils/zms-cli/zms-cli.go -z https://127.0.0.1:4443/zms/v1 -cert ../test/athenz/docker/sample/domain-admin/team_admin_cert.pem -key ../test/athenz/docker/sample/domain-admin/team_admin_key.pem -c ../test/athenz/docker/sample/CAs/athenz_ca.pem -o json list-domain
{
    "names": [
        "athenz",
        "athenz.aws",
        "athenz.aws-ecs",
        "athenz.aws-lambda",
        "athenz.test",
        "sys",
        "sys.auth",
        "sys.auth.audit",
        "sys.auth.audit.domain",
        "sys.auth.audit.org",
        "user"
    ],
    "next": ""
}
(venv) ~/Projects/athenz : 
(venv) ~/Projects/athenz :
(venv) ~/Projects/athenz : go run utils/zms-cli/zms-cli.go -z https://127.0.0.1:4443/zms/v1 -cert ../test/athenz/docker/sample/domain-admin/team_admin_cert.pem -key ../test/athenz/docker/sample/domain-admin/team_admin_key.pem -c ../test/athenz/docker/sample/CAs/athenz_ca.pem list-domain
domains:
    - athenz
    - athenz.aws
    - athenz.aws-ecs
    - athenz.aws-lambda
    - athenz.test
    - sys
    - sys.auth
    - sys.auth.audit
    - sys.auth.audit.domain
    - sys.auth.audit.org
    - user
```

- `get-quota`
```
(venv) ~/Projects/athenz : go run utils/zms-cli/zms-cli.go -z https://127.0.0.1:4443/zms/v1 -cert ../test/athenz/docker/sample/domain-admin/team_admin_cert.pem -key ../test/athenz/docker/sample/domain-admin/team_admin_key.pem -c ../test/athenz/docker/sample/CAs/athenz_ca.pem -o json -d user get-quota
{
    "name": "server-default",
    "subdomain": 100,
    "role": 1000,
    "roleMember": 100,
    "policy": 1000,
    "assertion": 100,
    "entity": 100,
    "service": 250,
    "serviceHost": 10,
    "publicKey": 100,
    "group": 100,
    "groupMember": 100,
    "modified": "2021-03-21T07:22:00.240Z"
}
(venv) ~/Projects/athenz : 
(venv) ~/Projects/athenz : 
(venv) ~/Projects/athenz : go run utils/zms-cli/zms-cli.go -z https://127.0.0.1:4443/zms/v1 -cert ../test/athenz/docker/sample/domain-admin/team_admin_cert.pem -key ../test/athenz/docker/sample/domain-admin/team_admin_key.pem -c ../test/athenz/docker/sample/CAs/athenz_ca.pem -d user get-quota
quota:
    subdomain: 100
    role: 1000
    role-member: 100
    group: 100
    group-member: 100
    policy: 1000
    assertion: 100
    service: 250
    service-host: 10
    public-key: 100
    entity: 100
    modified: 2021-03-21T07:22:00.240Z
```

Unsupported Format:
```
(venv) ~/Projects/athenz : go run utils/zms-cli/zms-cli.go -z https://127.0.0.1:4443/zms/v1 -cert ../test/athenz/docker/sample/domain-admin/team_admin_cert.pem -key ../test/athenz/docker/sample/domain-admin/team_admin_key.pem -c ../test/athenz/docker/sample/CAs/athenz_ca.pem -o unsupported-format list-domain
*** unsupported output format "unsupported-format"
exit status 1
```


libs/go/zmscli/cli.go
```diff
@@ -14,11 +14,20 @@ import (
...
 
-	"github.com/ardielle/ardielle-go/rdl"
 	"github.com/AthenZ/athenz/clients/go/zms"
+	"github.com/ardielle/ardielle-go/rdl"
 	"golang.org/x/net/proxy"
 )
 
+const (
+	// JSONOutputFormat is the JSON output format for commands.
+	JSONOutputFormat = "json"
```
OferLevi85, line 24: Goodjob having a library to take care of the json output. You'll notice that for yaml the prints are still hardcoded (with spaces for indentations).
Is it possible for you to replace the current yaml printing code with a library that will print the yaml output, similar to how you did in Json?

libs/go/zmscli/cli.go
```diff
@@ -942,8 +957,8 @@ func (cli Zms) HelpSpecificCommand(interactive bool, cmd string) string {
 	switch cmd {
 	case "list-domain":
 		buf.WriteString(" syntax:\n")
-		buf.WriteString("   list-domain [prefix]\n")
-		buf.WriteString("   list-domain [limit skip prefix depth]\n")
+		buf.WriteString("   [-o json] list-domain [prefix]\n")
```
OferLevi85, line 960: @jothi-prasad I noticed you open the issue for domain commands only so the flag is implemented only for domain-related actions. Any reason why the change isn't for all methods?

libs/go/zmscli/domain.go
```diff
@@ -38,7 +60,7 @@ func (cli Zms) DeleteDomain(dn string) (*string, error) {
 		}
 		if err == nil {
 			s := "[Deleted domain " + dn + "]"
-			return &s, nil
+			return cli.switchOverFormats(&StandardJSONMessage{Message: s}, s)
```
OferLevi85, line 63: I noticed that you wrap the output with StandardJSONMessage before calling switchOverFormats.
The assumption here is that there are only 2 formats:
1. The default (yaml) - which will use the second argument not wrapped
2. Json - which will this one.
This will work fine for 2 formats but if we would like to add a third in the future, say XML, this will need to change.
Please try to right it in such a way that adding new formats will not require changing the existing calls.

libs/go/zmscli/cli.go
patrasap0908: So the formatting for JSON works because the APIs return JSON responses, so it's a pretty simple thing to pretty-print those.
But I'm afraid it wouldn't be possible to do so for the YAML output since there are no YAML responses from the APIs, which is probably why they've been hardcoded.

libs/go/zmscli/domain.go
patrasap0908: Makes sense, let me change that function definition up a bit.

libs/go/zmscli/cli.go
OferLevi85: Yeah the API returns it as Json but I thought about having a library that converts from Json to Yaml.
That way adding new commands or changing existing commands will not require fumbling with Yaml indentations.
In any case this can be taken care of in a future PR

patrasap0908: Oh in that way, https://github.com/icza/dyno is one such library that can do that. A separate PR would be the way to go, you're right.

OferLevi85: https://github.com/ghodss/yaml also looks good

patrasap0908: Yes, that'll do it.

libs/go/zmscli/domain.go
patrasap0908: @OferLevi85, take a look at https://github.com/AthenZ/athenz/pull/1386/commits/84490b54393e1d9f4033ee326e622059849498d3#diff-8118320d935b49c9f953eebaf5401929b9737f08b783fe1a160e4421109acbcfL30-L38 and let me know what you think.

patrasap0908: With these new changes, any new output format can be supported without having to change the function call statements for single-line messages or for native Go structs in the future.

OferLevi85: Thanks I'll take a look.
By the way, please implement the flag to all methods, not just domain methods.

patrasap0908: Could you create a separate issue maybe for that? There are quite a few commands to cover.

OferLevi85 (approved):

Issue #1399, opened 2021-03-25
Implement a new API in ZTS to return workloads by IP or by service


abvaidya:
ZTS should expose an API to return workload details either by IP address or by domain and service as input parameters. Results should include workload details such as provider, unique identifier of the workload and last update time for the workload.

Issue #1398, opened 2021-03-25
Implement micro segmentation using Athenz


abvaidya:
Users can come to Athenz UI to define which services can connect to other services and during runtime, ZTS would expose a new API to return the list of ingress and egress transport rules, which will be used by a client utility on the host to update iptable rules.

PR #1400, opened 2021-03-25
RDL and generated files for issues #1398 and #1399

abvaidya:
Related to issue #1398, #1399


clients/java/zts/core/pom.xml
```diff
@@ -38,7 +38,7 @@
   </dependencyManagement>
 
   <properties>
-    <code.coverage.min>0.63</code.coverage.min>
+    <code.coverage.min>0.62</code.coverage.min>
```
abvaidya, line 41: Had to reduce this since PR only contains generated code.

havetisyan: we basically add mock functions to accommodate any new functions. so please make a note to go back and add those.

havetisyan (approved):

PR #1408, opened 2021-03-29
Support using customized PrivateKey object from HSM provider in Crypto.java

WindzCUHK:
## Issue

The implement for signing is hard-coded to use the BouncyCastle provider. As the HSM provider that we are planing to use has its own implementation on the PrivateKey object, which is not compatible with BouncyCastle's implementation for creating signature. Signature cannot be created with HSM provider specific PrivateKey object.

## Fix

1. remove the hard-coded part in `Crypto.java`
1. leverage the provider selection to `java.security` file
1. rename system properties `athenz.crypto.bc_provider` to `athenz.crypto.key_factory_provider`
    - The system properties `athenz.crypto.bc_provider` is not documented or found in the example configuration files. We think that it is better to clarify the usage of this system property.

### DIFF
--- a/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
+++ b/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
@@ -705,7 +705,7 @@ public class Crypto {
                 // Decrypt the private key with the specified password
 
                 InputDecryptorProvider pkcs8Prov = new JceOpenSSLPKCS8DecryptorProviderBuilder()
-                        .setProvider(BC_PROVIDER).build(pwd.toCharArray());
+                        .build(pwd.toCharArray());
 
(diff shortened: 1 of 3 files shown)


abvaidya (changes requested):
Please fix the failing zts-server unit tests because of this change.

havetisyan:
@WindzCUHK  this PR will generate major performance issues for all of our clients. The crypto class is used when our clients verify their access/role tokens and we've seen major performance issues when using the JDK default provider instead of BC when handling EC keys.

So we can't just relegate the provider to the java.security file as we can't require all of our clients to make such a change as their upgrade. So the PR should introduce a logic where based on some configuration setting we don't default to using BC provider.

r-andrew-dev:
To be eligible for swag, please [sign-up](https://docs.google.com/forms/d/e/1FAIpQLScFq4Bjrh5tVxDFWGlXROzNP-biUtVW7ZSNXCNzKVp8tg7Knw/viewform) for Yahoo Hack Together, if you have not done so, yet.

PR #1411, opened 2021-03-29
[HackTogether] ZTS API support for micro segmentation use cases

abvaidya:


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2283,21 +2282,135 @@ public RoleCertificate postRoleCertificateRequestExt(ResourceContext ctx, RoleCe
...
+
+        TransportRules transportRules = new TransportRules();
+        transportRules.setIngressRules(new ArrayList<>());
+        transportRules.setEgressRules(new ArrayList<>());
+
+        Map<String, List<String>> transportRulesBaseData = dataStore.getDataCache(domainName).getTransportRulesInfoForService(serviceName);
+        Workloads workloads;
+        if (transportRulesBaseData != null) {
+            for (Map.Entry<String, List<String>> entry : transportRulesBaseData.entrySet()) {
+                TransportRulesProcessor.TransportDirection transportDirection = TransportRulesProcessor.getDirection(entry.getKey());
```
havetisyan, line 2387: why do we have 2 separate lines 2387-2388? calling this twice requires the server to run the regex match twice. Shouldn't the transportDirection be part of the TransportRule?

servers/zts/src/main/java/com/yahoo/athenz/zts/transportrules/TransportRulesProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+
+public class TransportRulesProcessor {
+    private static final String TRANSPORT_RULES_ACTION_REGEX_PATTERN = "(TCP|UDP)-(IN|OUT):(\\d{1,5}-\\d{1,5}|\\d{1,5}):(\\d{1,5})";
+    private static final Pattern TRANSPORT_RULES_PATTERN = Pattern.compile(TRANSPORT_RULES_ACTION_REGEX_PATTERN);
+
+    public static boolean isTransportRuleAction(String action) {
+        return TRANSPORT_RULES_PATTERN.matcher(action).find();
+    }
+
+    public static TransportDirection getDirection(String key) {
```
havetisyan, line 32: made a comment in the code above - not sure why we have have this as a separate entry rather than part of the parseTransportRuleAction method?

servers/zts/src/main/java/com/yahoo/athenz/zts/workload/impl/DynamoDBWorkloadRecordStoreFactory.java
```diff
@@ -0,0 +1,63 @@
...
+import org.slf4j.LoggerFactory;
+
+public class DynamoDBWorkloadRecordStoreFactory implements WorkloadRecordStoreFactory {
+    private static final Logger LOGGER = LoggerFactory.getLogger(DynamoDBWorkloadRecordStoreFactory.class);
+
+    @Override
+    public WorkloadRecordStore create(PrivateKeyStore keyStore) {
+
+        final String tableName = System.getProperty(ZTSConsts.ZTS_PROP_WORKLOAD_DYNAMODB_TABLE_NAME);
+        if (tableName == null || tableName.isEmpty()) {
```
havetisyan, line 37: one general comment here - we should start using the jetty's StringUtil.isEmpty() method here since it handles the case with a single method and is better for code coverage.

havetisyan (approved):

PR #1412, opened 2021-03-29
Reload the key-store if the file is changed (configurable)

gilad-bendor:
Signed-off-by: Gilad Ben-Dor <gilad.bendor@verizonmedia.com>


containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -469,6 +470,23 @@ void addHTTPSConnector(HttpConfiguration httpConfig, int httpsPort, boolean prox
             sslConnector.addBean(connectionLogger);
         }
         server.addConnector(sslConnector);
+
+        // Reload the key-store if the file is changed
+System.err.println("=====GBENDOR:   Reload the key-store if the file is changed");
```
abvaidya, line 475: please clean the error messages and use logger.

abvaidya (changes requested):

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
gilad-bendor: you caught me in the middle of a "debugging" session...
;-)

containers/jetty/pom.xml
```diff
@@ -39,7 +39,7 @@
   </dependencyManagement>
 
   <properties>
-    <code.coverage.min>0.88</code.coverage.min>
+    <code.coverage.min>0.87</code.coverage.min>
```
OferLevi85, line 42: increasing the coverage is fine but reducing it isn't... add tests to cover the missing lines.

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -469,6 +470,18 @@ void addHTTPSConnector(HttpConfiguration httpConfig, int httpsPort, boolean prox
...
+
+        // Reload the key-store if the file is changed
+        final int reloadSslContextSeconds = Integer.parseInt(System.getProperty(AthenzConsts.ATHENZ_PROP_KEYSTORE_RELOAD_SEC, "0"));
+        if ((reloadSslContextSeconds > 0) && (sslContextFactory.getKeyStorePath() != null)) {
+            try {
+                KeyStoreScanner keystoreScanner = new KeyStoreScanner(sslContextFactory);
+                keystoreScanner.setScanInterval(reloadSslContextSeconds);
+                server.addBean(keystoreScanner);
+            } catch (IllegalArgumentException exception) {
+                LOG.error("Keystore will not be automatically reloaded when \"{}\" is changed: {}", sslContextFactory.getKeyStorePath(), exception.getMessage());
```
OferLevi85, line 482: I think that if reload is requested and there's an error during container boot we should throw an exception and prevent the service from coming up. Better than thinking it will be refreshed when it won't.

OferLevi85 (changes requested):
Please see my comments

containers/jetty/pom.xml
gilad-bendor: Reverted

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
gilad-bendor: Done

OferLevi85 (approved):

PR #1420, opened 2021-04-02
Update SIA AWS to rotate private key daily when refreshing service id…

pratikgote:
…entity certificates

### DIFF
--- a/provider/aws/sia-ec2/authn.go
+++ b/provider/aws/sia-ec2/authn.go
@@ -17,6 +17,8 @@
 package sia
 
 import (
+	"crypto/rsa"
+	"crypto/tls"
 	"crypto/x509"
 	"encoding/json"
 	"encoding/pem"
@@ -29,6 +31,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"os"
 	"os/exec"
 	"strings"
 	"time"
@@ -71,8 +74,18 @@ func GetRoleCertificate(ztsUrl, svcKeyFile, svcCertFile string, opts *options.Op
 			failures += 1
 			continue
 		}
-		certFilePem := util.GetRoleCertFileName(mkDirPath(opts.CertDir), role.Filename, roleName)
 		spiffe := fmt.Sprintf("spiffe://%s/ra/%s", domainNameRequest, roleNameRequest)
+
+		if opts.GenerateRoleKey {
+			var err error
+			key, err = RoleKey(opts.RotateKey, svcKeyFile)
+			if err != nil {
+				logutil.LogInfo(sysLogger, "unable to read generate/read key from %s, err: %v\n", role.Filename, err)
+				failures += 1
+				continue
+			}
+		}
+
 		csr, err := util.GenerateCSR(key, opts.Domain, opts.Services[0].Name, roleName, "", provider, spiffe, opts.ZTSAWSDomain, true)
 		if err != nil {
 			logutil.LogInfo(sysLogger, "unable to generate CSR for %s, err: %v\n", roleName, err)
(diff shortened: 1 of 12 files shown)


provider/aws/sia-ec2/options/options.go
```diff
@@ -170,7 +181,7 @@ func initFileConfig(bytes []byte, accountId string) (*Config, *ConfigAccount, er
 
 // NewOptions takes in sia_config bytes and returns a pointer to Options after parsing and initializing the defaults
 // It uses profile arn for defaults when sia_config is empty or non-parsable. It populates "services" array
-func NewOptions(bytes []byte, accountId, metaEndPoint, siaDir, version, ztsCaCert, ztsServerName, ztsAwsDomain string, sysLogger io.Writer) (*Options, error) {
+func NewOptions(bytes []byte, accountId, metaEndPoint, siaDir, version, ztsCaCert, ztsServerName, ztsAwsDomain, taskIdPrefix string, sysLogger io.Writer) (*Options, error) {
```
havetisyan, line 184: taskIdPrefix is confusing here because it's not. In the caller it's correctly identified as providerPfx. So we should carry that forward. In fact I would suggest replace providerPfx with providerParentDomain since that's exactly what it is. The logic is we generate the provider service name as follows:

<provider-parent-domain>.aws.<region-name>

So: <provider-parent-domain>.aws is the provider domain name
<region-name> is the provider service name

provider/aws/sia-ec2/options/options.go
```diff
@@ -267,15 +307,33 @@ func NewOptions(bytes []byte, accountId, metaEndPoint, siaDir, version, ztsCaCer
...
 }
 
 func GetOSType() int {
-	// this is only needed if sshd on the given os
-	// needs something special for the restart
-	// typically one would parse the /etc/os-release
-	// for now we have a single method for restarting
-	// sshd so we'll just return unknown os type
-
+	// first check to see if it's centos - currently
```
havetisyan, line 318: No changes are needed here. CentosOS6 and Ubuntu14 are no longer supported. So the default way of restarting sshd is sufficient for all OS types we support.

provider/aws/sia-ec2/authn.go
```diff
@@ -71,8 +74,18 @@ func GetRoleCertificate(ztsUrl, svcKeyFile, svcCertFile string, opts *options.Op
 			failures += 1
 			continue
 		}
-		certFilePem := util.GetRoleCertFileName(mkDirPath(opts.CertDir), role.Filename, roleName)
 		spiffe := fmt.Sprintf("spiffe://%s/ra/%s", domainNameRequest, roleNameRequest)
+
+		if opts.GenerateRoleKey == true {
```
havetisyan, line 79: let's just have:

if opts.GenerateRoleKey {
}

havetisyan (changes requested):

provider/aws/sia-ec2/options/options.go
pratikgote: done

provider/aws/sia-ec2/options/options.go
pratikgote: done

provider/aws/sia-ec2/authn.go
pratikgote: done

provider/aws/sia-ec2/cmd/siad/main.go
```diff
@@ -52,20 +55,43 @@ func main() {
 	ztsServerName := flag.String("ztsservername", "", "zts server name for tls connections")
 	ztsCACert := flag.String("ztscacert", "", "zts CA certificate file")
 	ztsAwsDomain := flag.String("ztsawsdomain", "", "ZTS AWS Domain")
+	providrPntDomain := flag.String("providerParentDomain", "athenz", "providerParentDomain")
+	version := flag.String("version", "1.0", "version")
```
havetisyan, line 59: this doesn't look right. the caller should not be able to specify their version number otherwise anyone can use any value they want. This should be hardcoded value with the capability to specify it during the build process.

havetisyan (changes requested):

provider/aws/sia-ec2/cmd/siad/main.go
pratikgote: updated

havetisyan (approved):

PR #1428, opened 2021-04-07
Add group review days to role

OferLevi85:


libs/java/client_common/src/main/java/com/yahoo/athenz/common/utils/SignUtils.java
```diff
@@ -171,6 +173,8 @@ private static Struct asStruct(Role role) {
         appendObject(struct, ATTR_SELF_SERVE, role.getSelfServe());
         appendObject(struct, ATTR_SERVICE_EXPIRY_DAYS, role.getServiceExpiryDays());
         appendObject(struct, ATTR_SERVICE_REVIEW_DAYS, role.getServiceReviewDays());
+        appendObject(struct, ATTR_GROUP_EXPIRY_DAYS, role.getGroupExpiryDays());
```
havetisyan, line 176: the attributes in the signed document are in sorted attribute order. so this not the right place for these values.

havetisyan (changes requested):

libs/java/client_common/src/main/java/com/yahoo/athenz/common/utils/SignUtils.java
OferLevi85: Fixed

havetisyan (approved):

PR #1430, opened 2021-04-07
new DomainMetaStore interface support in ZMS

havetisyan:
A new DomainMetaStore interface that allows the server to verify and update domain's meta attributes in some external store. By default we have a No-Op implementation that does nothing. The current supported attributes are:

businessService
awsAccount
azureSubscription
productId

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -28,6 +28,8 @@ import com.yahoo.athenz.common.server.audit.AuditReferenceValidator;
 import com.yahoo.athenz.common.server.audit.AuditReferenceValidatorFactory;
 import com.yahoo.athenz.common.server.log.AuditLogger;
 import com.yahoo.athenz.common.server.log.AuditLoggerFactory;
+import com.yahoo.athenz.common.server.metastore.DomainMetaStore;
+import com.yahoo.athenz.common.server.metastore.DomainMetaStoreFactory;
 import com.yahoo.athenz.common.server.notification.Notification;
 import com.yahoo.athenz.common.server.notification.NotificationManager;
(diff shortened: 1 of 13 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -1856,6 +1898,79 @@ void validateDomainValues(Domain domain) {
...
+
+        final String caller = "validateDomainRegularMetaStoreValues";
+
+        // we're going to check the meta values for our new domain
+        // requests against our meta store - for now we only have
+        // business service as a regular domain meta attribute
+
+        BitSet changedAttrs = new BitSet();
+        if (ZMSUtils.metaValueChanged(domain.getBusinessService(), meta.getBusinessService())) {
+            if (!domainMetaStore.isValidBusinessService(domain.getName(), meta.getBusinessService())) {
```
abvaidya, line 1967: This question applies to all meta attributes.

Lets say domain A has business service X, and its been changed to value Y, whats the purpose of passing domainName to domainMetaStore.isValidBusinessService()
How does a real life validation look like? e.g Business service not associated with any other domain?

havetisyan: it's completely up to the implementation. For example, I have an external service that already maintains the mapping of athenz domains and business service. Since the domain admin has the rights to set the business service I don't want them to set to some random value. So the implementation will call isValidBusinessService for domain A and value Y. If the external store says it must be service X, then the request will be rejected so that the domain admin can't set invalid value Y in Athenz.

havetisyan: The other use case would be that it's just completely ignored. I only want to check if the value is an actual business service or not. In that case the implementation will just ignore the domainName and will only verify that businessService value is valid or not.

PR #1443, opened 2021-04-15
rdl changes to add hostname in workload response

jothi-prasad:


core/zts/src/test/java/com/yahoo/athenz/zts/WorkloadTest.java
```diff
@@ -81,6 +82,9 @@ public void testWorkloadFields() {
         assertNotEquals(wl1, wl2);
 
         wl2.setUpdateTime(Timestamp.fromMillis(123456789123L));
+        wl2.setHostname("random");
```
havetisyan, line 85: we seem to be missing test cases here. After you set random, you need to verify that assertNotEquals(wl1, wl2);
then you set the value to null, and then verify again that it's assertNotEquals(wl1, wl2);
then you store value back to its original value and verify they're equal.  as in lines 87-88.

This sequence verifies all the lines have code coverage.

Whenever making changes, always run clover in the mode where the changes were mode:

mvn -P coverage clean org.openclover:clover-maven-plugin:instrument org.openclover:clover-maven-plugin:clover

and verify in the clover output that all your new lines have full code coverage.

havetisyan (changes requested):

core/zts/src/test/java/com/yahoo/athenz/zts/WorkloadTest.java
jothi-prasad: fixed the tests and also ran the code coverage plugin. the code coverage is 100% now.

havetisyan (approved):

PR #1445, opened 2021-04-16
Fix invalid environment in athenz_ui script

upgle:
## Changes
I found an invalid environment `ZMS_SERVER_URL` -> `ZMS_SERVER` in athenz_ui script and fixed it.

### DIFF
--- a/ui/scripts/athenz_ui
+++ b/ui/scripts/athenz_ui
@@ -36,7 +36,7 @@ if [ -z "${ZMS_SERVER}" ]; then
     exit 1
 fi
 
-if [ -z "${ZMS_SERVER_URL}" ]; then
+if [ -z "${ZMS_SERVER}" ]; then
     export ZMS_SERVER_URL=https://${ZMS_SERVER}:4443/zms/v1/
 fi
 


upgle:
Do I need to do anything else to get this PR reviewed?
Please let me know if I need anything to do other than a sign.

abvaidya:
Thank you for the PR.

havetisyan:
@abvaidya we should revisit this. I don't think the PR was correct. In fact it removes the capability for the caller to specify the ZMS_SERVER_URL with its own port number. We already verify on line 34 that ZMS_SERVER is set so there is no need to verify again on line 39. In fact the idea, iirc, was the caller to specify the ZMS_SERVER and then if they want they can specify ZMS_SERVER_URL and if one is not set, we'll generate one automatically.

@upgle can you clarify what issue where you having that changing the env variable fixed?

abvaidya:
oh I missed the custom ZMS_SERVER_URL logic, will revert the PR. Thanks for catching this.

upgle:
@havetisyan 
Oh, I'm really sorry. I think I misunderstood the code.

PR #1451, opened 2021-04-21
adding support for hostname field in workload table

jothi-prasad:


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2906,7 +2906,12 @@ public Response postInstanceRegisterInformation(ResourceContext ctx, InstanceReg
 
         if (enableWorkloadStore && !athenzSysDomainCache.isWorkloadStoreExcludedProvider(provider)) {
             // insert into workloads store is on best-effort basis. No errors are thrown if the op is not successful.
-            insertWorkloadRecord(cn, provider, certReqInstanceId, sanIpStrForWorkloadStore);
+            String hostName = info.getHostname();
+            if (hostName == null) {
+                LOGGER.debug("hostname is not set by agent, hence forming the hostname {} with domain {} service {} and sanIpStr {} ..", hostName, info.getDomain(), info.getService(), sanIpStrForWorkloadStore);
```
havetisyan, line 2911: the debug statement with multiple arguments must be enclosed within (LOGGER.isDebugEnabled()) {}

havetisyan: hostName is already null, so not sure why we're including in the debug statement

havetisyan: another suggestion I would make here is the moving this block within the insertWorkloadRecord method itself rather than having it here. We already pass the cn and sanIpStrForWorkloadStore to that function so we can generate the hostname there as well. We also get the benefit that if sanIpStrForWorkloadStore is null, then we won't do any extra work and return right away.

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
jothi-prasad: done. and moved the logic to insertWorkloadRecord method.

havetisyan (approved):

PR #1456, opened 2021-04-25
Add api: get valid values for domain meta attributes 

OferLevi85:
To be used mainly by UI to display list of valid services for a user


core/zms/src/main/rdl/Domain.rdli
```diff
@@ -282,3 +282,18 @@ resource DomainTemplate DELETE "/domain/{name}/template/{template}" {
         ResourceError TOO_MANY_REQUESTS;
     }
 }
+
+//List of Business Services
+type BusinessServiceList Struct {
```
havetisyan, line 287: I don't see why we need to define a specific object type for Business Services. This is not an Athenz object - just list of strings so we should keep at as is - Array<String>, more details on the api call itself

core/zms/src/main/rdl/Domain.rdli
```diff
@@ -282,3 +282,18 @@ resource DomainTemplate DELETE "/domain/{name}/template/{template}" {
...
+    Array<String> businessServices; //list of business services
+}
+
+//List all valid Business Services.
+resource BusinessServiceList GET "/bservices?user={user}" {
```
havetisyan, line 292: We should also not have a specific endpoint for this - it assumes that somehow we're managing business services. We just need a more generic api if we want to implement such functionality. These are all domain meta attributes so it should be something like:

resource Array<String> GET "/domain/metastore?attribute=<attributeName>&user=<userName>

the server can verify that the attributeName is one of the supported meta store attributes (e.g. businessService, azureSubscription, awsAccount, etc. and call the appropriate megastore api.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/metastore/DomainMetaStore.java
```diff
@@ -45,6 +47,13 @@
      */
     void setBusinessServiceDomain(final String domainName, final String businessService);
 
+    /**
+     * Get a list of valid business services
+     * @param userName (optional) if not null, only get business services associated with the user
+     * @return Business Services List
+     */
+    List<String> getValidBusinessServices(final String userName);
```
havetisyan, line 55: we should also add similar api methods for the other attributes that the meta store supports:

getValidAWSAccounts()
getValidAzureSubscriptions(),e tc.

havetisyan (changes requested):

core/zms/src/main/rdl/Domain.rdli
OferLevi85: I’m fixing it so the endpoint will be used to fetch valid values for attribute but I still won’t be able to return a list of strings as you showed in the example:

`resource Array<String> GET "/domain/metastore?attribute=<attributeName>&user=<userName>
`
RDL generator doesn’t accept it. The resource must be a declared type even if it’s just a String array (the name of the function generated in ZMSResources is determined by this type + action) So I’ve created the following type:
```
//List of valid domain meta attribute values
type DomainMetaStoreValidValuesList Struct {
    Array<String> validValues; //list of valid values for attribute
}
```

libs/java/server_common/src/test/java/com/yahoo/athenz/common/server/metastore/DomainMetaStoreTest.java
```diff
@@ -41,6 +44,8 @@ public void testDomainMetaStoreDefaults() {
         assertTrue(metaStore.isValidBusinessService("athenz", "security"));
         assertTrue(metaStore.isValidBusinessService("athenz", null));
 
+        assertEquals(metaStore.getValidBusinessServices(null), new ArrayList<>());
```
havetisyan, line 47: we should add one of those for other attributes as well for code coverage.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2375,6 +2375,52 @@ public void deleteDomainTemplate(ResourceContext ctx, String domainName, String
         dbService.executeDeleteDomainTemplate(ctx, domainName, templateName, auditRef, caller);
     }
 
+    @Override
+    public DomainMetaStoreValidValuesList getDomainMetaStoreValidValuesList(ResourceContext ctx, String attributeName, String userName) {
```
havetisyan, line 2379: Do we have test cases for this new method?

libs/java/server_common/src/test/java/com/yahoo/athenz/common/server/metastore/DomainMetaStoreTest.java
OferLevi85: Added missing calls

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
OferLevi85: Added missing tests

servers/zms/pom.xml
```diff
@@ -39,7 +39,7 @@
   </dependencyManagement>
 
   <properties>
-    <code.coverage.min>0.96</code.coverage.min>
+    <code.coverage.min>0.964</code.coverage.min>
```
OferLevi85, line 42: Increased precision so we'll be alerted immediately on new functionality not covered with tests

havetisyan (approved):

PR #1473, opened 2021-05-03
support for  search filter in service details page

jothi-prasad:


ui/src/components/service/InstanceList.js
```diff
@@ -124,25 +182,65 @@ export default class InstanceList extends React.Component {
...
+                            <SearchInput
+                                dark={false}
+                                name='search'
+                                fluid={true}
+                                value={this.state.searchText}
+                                placeholder={'Search'}
+                                error={this.state.error}
+                                onChange={(event) =>
+                                    this.setState({
+                                        searchText: event.target.value,
```
pratikgote, line 231: instead of using DOM elements, we should either make use of state/props

pratikgote (changes requested):

abvaidya (approved):

PR #1486, opened 2021-05-18
Support using customized PrivateKey object from HSM provider in Crypto.java for signing

WindzCUHK:
## Issue
The implement for signing is hard-coded to use the BouncyCastle provider. As the HSM provider that we are planing to use has its own implementation on the PrivateKey object, which is not compatible with BouncyCastle's implementation for creating signature. Signature cannot be created with HSM provider specific PrivateKey object.

## Fix
1. remove the hard-coded part in Crypto.java
1. replace system properties athenz.crypto.bc_provider
    1. add athenz.crypto.key_factory_provider
    1. add athenz.crypto.signature_provider

### related issue
- #1408

### DIFF
--- a/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
+++ b/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
@@ -116,7 +116,8 @@ public class Crypto {
     public static final String SHA1 = "SHA1";
     public static final String SHA256 = "SHA256";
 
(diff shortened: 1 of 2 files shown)


havetisyan (approved):

havetisyan:
Thanks @WindzCUHK Do you guys need a new library release in Maven Central with this change or you're building your own versions of libraries and not use the official ones from Maven Central?

falz-tino:
Thank you for your confirmation.  We do not need a new library release only for this change.  We can wait for the next planned official release.

PR #1488, opened 2021-05-20
Add api to clear domain meta attribute

OferLevi85:


havetisyan:
I'm not sure if I want a separate endpoint just because the java client is not able to send the empty strings to ZMS. We should try to see if we can fix the java client to handle this case correctly.

PR #1490, opened 2021-05-23
Fix putDomainMeta and enabled rdl structs

OferLevi85:
with empty values

### DIFF
--- a/rdl/rdl-gen-athenz-java-model/javamodel.go
+++ b/rdl/rdl-gen-athenz-java-model/javamodel.go
@@ -654,7 +654,11 @@ func (gen *javaModelGenerator) emitStructFields(fields []*rdl.StructFieldDef, na
 			if optional {
 				gen.emit("    @RdlOptional\n")
 				if gen.jackson {
-					gen.emit("    @JsonInclude(JsonInclude.Include.NON_EMPTY)\n")
+				    if f.Annotations["x_allowempty"] == "true" {
+				        gen.emit("    @JsonInclude(JsonInclude.Include.NON_NULL)\n")
+				    } else {
+					    gen.emit("    @JsonInclude(JsonInclude.Include.NON_EMPTY)\n")
+					}
 				}
 			}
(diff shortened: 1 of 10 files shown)


rdl/rdl-gen-athenz-java-model/javamodel.go
```diff
@@ -647,14 +647,19 @@ func (gen *javaModelGenerator) emitStructFields(fields []*rdl.StructFieldDef, na
 			fnames = append(fnames, fname)
 			optional := f.Optional
 			ftype := javaType(gen.registry, f.Type, optional, f.Items, f.Keys)
+			allowEmpty := strings.HasSuffix(string(f.Type), "AllowEmpty")
```
OferLevi85, line 650: Initially, I wanted to add another attribute to the rdl schema field (like "optional" I wanted to add "allowEmpty"). However, it requires changing the original RDL project. 
So I went with changing our local RDL code and the logic is that we'll allow empty values for structs with the postfix "AllowEmpty". Currently there's only one such structure - BusinessServiceAllowEmpty

havetisyan: @OferLevi85  RDL already provides full support for extensible annotations and since we have our own java model generator it's very simple to implement this support.

Update the field definition in rdl:

`
String businessService (optional, x_allowempty="true"); //associated business service with domain
`

(rdl automatically parses any attribute starting x_ value and stores them in map - the only requirement is that they all need to be strings0.

Then in the java model generator where you include the NON_EMPTY vs NON_NULL you can have something like:

```
if f.Annotations["x_allowempty"] == "true" {
    gen.emit("    @JsonInclude(JsonInclude.Include.NON_NULL)\n")
} else {
    gen.emit("    @JsonInclude(JsonInclude.Include.NON_EMPTY)\n")
}
```

OferLevi85: I wasn't aware of it. Thanks, I'll make the change

havetisyan (approved):

PR #1496, opened 2021-05-30
UI: Fix 3 bugs associated with business service

OferLevi85:
1. Fixed bug where pressing "cancel" will show the business service as empty even
though the value hasn't changed in backend
2. Fixed bug where changing business service and re-opening the dialog will show the old value by default
3. Fixed bug where a default justification message was used for audit enabled domains
4. Fixed error messages (now appear in modal)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,7 +120,6 @@ libs/go/athenzutils/pkg/
 clients/nodejs/zpe/package-lock.json
 clients/nodejs/zts/package-lock.json
 libs/nodejs/auth_core/package-lock.json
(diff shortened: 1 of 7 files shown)


chandrasekhar1996:
If a justification/audit ref isn't provided we should show an error message and not let the user submit the form for audit enabled domains.

OferLevi85:
> If a justification/audit ref isn't provided we should show an error message and not let the user submit the form for audit enabled domains.

Fixed.

chandrasekhar1996 (approved):

PR #1505, opened 2021-06-07
UI: Onboard template to domain

OferLevi85:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -3074,9 +3074,13 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         validateRequest(ctx.request(), caller);
 
         List<TemplateMetaData> serverTemplateMetadataList = serverSolutionTemplates.getTemplates()
-                .values()
+                .entrySet()
                 .stream()
-                .map(template -> template.getMetadata())
+                .map(template -> {
+                    TemplateMetaData metadata = template.getValue().getMetadata();
+                    metadata.setTemplateName(template.getKey());
+                    return metadata;
+                })
                 .collect(Collectors.toList());
 
         DomainTemplateDetailsList serverTemplateDetailsList = new DomainTemplateDetailsList();
--- a/ui/src/components/template/TemplateList.js
+++ b/ui/src/components/template/TemplateList.js
@@ -46,25 +46,79 @@ const TableHeadStyled = styled.th`
     word-break: break-all;
 `;
 
+const TitleDiv = styled.div`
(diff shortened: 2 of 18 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3074,9 +3074,13 @@ public DomainTemplateDetailsList getServerTemplateDetailsList(ResourceContext ct
         validateRequest(ctx.request(), caller);
 
         List<TemplateMetaData> serverTemplateMetadataList = serverSolutionTemplates.getTemplates()
-                .values()
+                .entrySet()
```
OferLevi85, line 3077: Added template names to the response

ui/src/components/template/TemplateList.js
```diff
@@ -46,25 +46,88 @@ const TableHeadStyled = styled.th`
...
+            errorMessage: errorMessage,
+        });
+        // this is to close the failure alert
+        setTimeout(
+            () =>
+                this.setState({
+                    showError: false,
+                    errorMessage: '',
+                }),
+            MODAL_TIME_OUT
```
chandrasekhar1996, line 103: we shouldn't use timed close for errors. It should persist on the screen until user actively dismisses it.

ui/src/components/template/TemplateList.js
```diff
@@ -157,6 +253,28 @@ export default class TemplateList extends React.Component {
...
                     </thead>
                     <tbody>{rows}</tbody>
+                    <TitleDiv>Server Templates</TitleDiv>
+                    <thead>
+                        <tr>
+                            <TableHeadStyled align={left}>
+                                TEMPLATE
+                            </TableHeadStyled>
+                            <TableHeadStyled align={left}>INFO</TableHeadStyled>
+                            <TableHeadStyled align={center}>
```
chandrasekhar1996, line 263: if the server template list only contains the templates which haven't be applied to the domain then do we require CurrentVersion column?

OferLevi85: It's for symmetry with the "existing templates" table above it.

chandrasekhar1996: ohh in that case it would be better to show 0 or N/A as value instead of leaving it empty

OferLevi85: Will place N/A

ui/src/components/template/TemplateList.js
OferLevi85: Fixed

ui/src/components/template/TemplateList.js
OferLevi85: Fixed

chandrasekhar1996 (approved):

PR #1517, opened 2021-06-11
Update Tag Values to include '/', ':'

pratikgote:
Signed-off-by: pgote <pgote@verizonmedia.com>


ui/package-lock.json
```diff
@@ -1,18965 +1,8 @@
 {
     "name": "@athenz/ui",
     "version": "0.4.0",
-    "lockfileVersion": 2,
+    "lockfileVersion": 1,
```
abvaidya, line 4: what makes this change?

abvaidya (changes requested):
what is the change for which all snapshots are regenerated and why there is a difference in package-lock.json version?

ui/package-lock.json
pratikgote: that is because the package-lock.json is generated using a node version and if we run _npm i_ using a different node version, then the lockfileVersion gets updated. I reverted it back to 2 manually, but in my machine, node keeps updating lockfileVersion in package-lock.json. 

```
➜  node --version
v12.20.1
➜ 
```

PR #1520, opened 2021-06-15
New endpoint for UI: Get all valid authrority attributes

OferLevi85:


core/zms/src/main/rdl/Authority.rdli
```diff
@@ -0,0 +1,19 @@
+// Copyright 2021 Athenz Authors
```
havetisyan, line 1: we no longer use year in copyright lines

core/zms/src/main/rdl/Authority.rdli
```diff
@@ -0,0 +1,19 @@
...
+// Licensed under the terms of the Apache version 2.0 license. See LICENSE file for terms.
+
+//List of valid user authority attribute values
+type UserAuthorityValidValuesList Struct {
+    Array<String> validValues; //list of valid values for attribute
+}
+
+//List all valid values for the user authority
+resource UserAuthorityValidValuesList GET "/authority/user?attributeType={attributeType}" {
```
havetisyan, line 10: let's use /authority/user/attribute?type={type} in case we need to add more endpoints for the user authority in the future

core/zms/src/main/rdl/Authority.rdli
```diff
@@ -0,0 +1,19 @@
+// Copyright 2021 Athenz Authors
+// Licensed under the terms of the Apache version 2.0 license. See LICENSE file for terms.
+
+//List of valid user authority attribute values
+type UserAuthorityValidValuesList Struct {
```
havetisyan, line 5: let's replace the Valid since that's kind of implied. So maybe we can use:

UserAuthorityAttributeValueList

havetisyan, line 5: Does it make sense to combine the attributes into a single call and just return the type as part of the response?

Something like:

type UserAuthorityAttributeValue Struct {
    SimpleName name;
    SimpleName type;
}

UserAuthorityAttributeValueList Struct {
    Array<UserAuthorityAttributeValue> values;
}

core/zms/src/main/rdl/Authority.rdli
havetisyan, line 10: take a look at the comment above to see if we can remove the query and just return all attribute in one call?

havetisyan (changes requested):

core/zms/src/main/rdl/Authority.rdli
OferLevi85: Sure but because each type may have several attributes I think I'll hold it as a map (type->List of attributes). Let me know what you think

core/zms/src/main/rdl/Authority.rdli
OferLevi85: Fixed

core/zms/src/main/rdl/Authority.rdli
OferLevi85: Endpoint now return all types so changed it to:
/authority/user/attribute

OferLevi85: Fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -8363,6 +8363,22 @@ public DomainRoleMembership getPendingDomainRoleMembersList(ResourceContext ctx,
...
 
+    @Override
+    public UserAuthorityAttributeMap getUserAuthorityAttributeMap(ResourceContext ctx) {
+        final String caller = ctx.getApiName();
+
+        logPrincipal(ctx);
+
+        validateRequest(ctx.request(), caller);
+
+        Map<String, UserAuthorityAttributes> attriuteMap = new HashMap<>();
```
pratikgote, line 8374: there is a typo here, shouldn't it be attributeMap?

pratikgote (approved):

abvaidya:
@OferLevi85 please take care of the typo in your next PR.

PR #1522, opened 2021-06-21
Allow system admins to set domain meta attribute business service

OferLevi85:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -1966,6 +1966,14 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
                     changedAttrs.set(DomainMetaStore.META_ATTR_PRODUCT_ID);
                 }
                 break;
+            case ZMSConsts.SYSTEM_META_BUSINESS_SERVICE:
+                if (ZMSUtils.metaValueChanged(domain.getBusinessService(), meta.getBusinessService())) {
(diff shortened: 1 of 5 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -1966,6 +1966,13 @@ BitSet validateDomainSystemMetaStoreValues(Domain domain, DomainMeta meta, final
                     changedAttrs.set(DomainMetaStore.META_ATTR_PRODUCT_ID);
                 }
                 break;
+            case ZMSConsts.SYSTEM_META_BUSINESS_SERVICE:
+                if (ZMSUtils.metaValueChanged(domain.getBusinessService(), meta.getBusinessService())) {
+                    if (!domainMetaStore.isValidBusinessService(domain.getName(), meta.getBusinessService())) {
+                        throw ZMSUtils.requestError("invalid business service for domain", caller);
+                    }
+                    changedAttrs.set(DomainMetaStore.META_ATTR_BUSINESS_SERVICE);
+                }
```
havetisyan, line 1975: let's add a break statement here otherwise we might miss in the future when we add additional case lines.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
OferLevi85: Fixed

havetisyan (approved):

PR #1523, opened 2021-06-22
Assertion conditions implementation - core, clients, DB schema changes

abvaidya:


servers/zms/pom.xml
```diff
@@ -39,7 +39,7 @@
   </dependencyManagement>
 
   <properties>
-    <code.coverage.min>0.9652</code.coverage.min>
+    <code.coverage.min>0.9650</code.coverage.min>
```
abvaidya, line 42: temp change to not fail the build for generated method signatures. Will revert this with server side changes

servers/zms/pom.xml
```diff
@@ -39,7 +39,7 @@
   </dependencyManagement>
 
   <properties>
-    <code.coverage.min>0.9652</code.coverage.min>
+    <code.coverage.min>0.9649</code.coverage.min>
```
abvaidya, line 42: temp change to not fail the build for generated method signatures. Will revert this with server side changes

core/zms/src/main/rdl/Policy.tdl
```diff
@@ -16,6 +36,7 @@ type Assertion Struct {
     AssertionEffect effect (optional, default=ALLOW); //the effect of the assertion in the policy language
     Int64 id (optional); //assertion id - auto generated by server. Not required during put operations.
     Bool caseSensitive (optional); // If true, we should store action and resource in their original case
+    AssertionConditions assertionConditions (optional); // optional list of assertion conditions associated with given assertion
```
havetisyan, line 39: we should name this just conditions otherwise we're repeating assertion part and be consistent with other fields (e.g. we're calling effect and not assertionEffect).

core/zms/src/main/rdl/Policy.tdl
```diff
@@ -7,6 +7,26 @@ include "Names.tdl";
...
+
+// A representation of condition associated with an assertion
+type AssertionCondition Struct {
+    Int32 id (optional); //condition id - auto generated by server. Not required during put operations.
+    Map<AssertionConditionKey, AssertionConditionData> conditionsMap; // each key in the map represents a unique condition. All the keys present in the map form a logical condition with AND operation.
+}
+
+//The representation of list of assertion conditions
+type AssertionConditions Struct {
+    Array<AssertionCondition> assertionConditionsList; // list of assertion conditions.
```
havetisyan, line 27: let's name this field just conditionsList and not repeat assertions part.

servers/zms/schema/updates/update-20210621.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE IF NOT EXISTS `zms_server`.`assertion_condition` (
+  `assertion_id` INT UNSIGNED NOT NULL,
+  `condition_id` INT NOT NULL,
+  `key` VARCHAR(64) NOT NULL,
+  `operator` VARCHAR(16) NOT NULL,
+  `value` VARCHAR(2048) NOT NULL,
+  PRIMARY KEY (`assertion_id`, `condition_id`, `key`),
+  INDEX `fk_condition_assertion_idx` (`assertion_id` ASC),
+  INDEX `key_idx` (`key` ASC),
```
havetisyan, line 9: why are we creating an index for the key? what's the use case for this?

havetisyan (changes requested):

servers/zms/schema/updates/update-20210621.sql
abvaidya: that was for before when we wanted to manipulate individual keys, will remove it.

havetisyan (approved):

PR #1525, opened 2021-06-25
Add debug logs to KeyRefresher for Thread identity

ashwintumma23:
In our application code, we are using multiple threads which invoke the KeyRefresher class. But only a subset of them are found to be running after a specific time. So, we want to verify if there are any exceptions in the threads, and which threads are currently running when Trust Managers and Key Managers are refreshed. 

Signed-off-by: Ashwin Tumma atumma@verizonmedia.com


libs/java/cert_refresher/src/main/java/com/oath/auth/KeyRefresher.java
```diff
@@ -160,6 +162,7 @@ private void scanForFileChanges() {
                         Thread.sleep(retryFrequency);
                     }
                 } catch (InterruptedException ignored) {
+                    LOGGER.error(ignored.getMessage());
```
pranavbhole, line 165: print error text with original exception

ashwintumma23: Thanks, @pranavbhole. Incorporated the changes in https://github.com/AthenZ/athenz/pull/1526

PR #1526, opened 2021-06-26
Add debug logs to KeyRefresher for Thread identity

ashwintumma23:
In our application code, we are using multiple threads which invoke the KeyRefresher class. But only a subset of them are found to be running after a specific time. So, we want to verify if there are any exceptions in the threads, and which threads are currently running when Trust Managers and Key Managers are refreshed.

Signed-off-by: Ashwin Tumma <atumma@verizonmedia.com>


ashwintumma23:
Thank you for your comment, @pranavbhole, on https://github.com/AthenZ/athenz/pull/1525 . Incorporated changes in this PR; somehow DCO was failing in the other.

pranavbhole (approved):

havetisyan:
Is there any reason why just using "[%thread]" in your slf4j configuration xml file doesn't give you the needed details?

PR #1527, opened 2021-06-27
zms-cli: generic output conversion

OferLevi85:


libs/go/zmscli/cli.go
```diff
@@ -23,7 +24,9 @@ import (
 const (
 	// JSONOutputFormat is the JSON output format for commands.
 	JSONOutputFormat = "json"
-	// DefaultOutputFormat is the default YAML output format for commands.
+	// YAMLOutputFormat is the YAML output format for commands.
+	YAMLOutputFormat = "yamlNew"
```
OferLevi85, line 28: For now, to show the "generic Yaml" format we'll use the yamlNew as the output flag. Once we get rid of the "manual yaml" we'll replace this flag with a simple "yaml".

libs/go/zmscli/cli.go
```diff
@@ -60,6 +63,30 @@ func (cli Zms) buildJSONOutput(res interface{}) (*string, error) {
 	return &output, nil
 }
 
+func (cli Zms) buildYAMLOutput(res interface{}) (*string, error) {
```
OferLevi85, line 66: buildYAMLOutput builds generic Yaml based on Json

libs/go/zmscli/cli.go
```diff
@@ -60,6 +63,30 @@ func (cli Zms) buildJSONOutput(res interface{}) (*string, error) {
...
+	yamlOutput, err := yaml.Marshal(res)
+	if err != nil {
+		return nil, fmt.Errorf("failed to produce YAML output: %v", err)
+	}
+	output := string(yamlOutput)
+	return &output, nil
+}
+
+type YamlConverter func(res interface{}) (*string, error)
```
OferLevi85, line 75: For old (manual) YAML converter, I'll wrap the logic in a "function pointer" I can pass to dumpByFormat and just call it if the old type is used.
Once we get rid of the old yaml dump we can remove this type along with the old dump logic.

libs/go/zmscli/role.go
```diff
@@ -66,10 +67,24 @@ func (cli Zms) ShowRole(dn string, rn string, auditLog, expand bool, pending boo
 	if err != nil {
 		return nil, err
 	}
-	var buf bytes.Buffer
-	buf.WriteString("role:\n")
-	cli.dumpRole(&buf, *role, auditLog, indentLevel1Dash, indentLevel1DashLvl)
-	return cli.switchOverFormats(role, buf.String())
+
+	oldYamlConverter := func(res interface{}) (*string, error) {
```
OferLevi85, line 71: Right now I've only implemented the "pass as function" logic in the ShowRole method. After this change is approved I'll implement it in the rest of the functions.

libs/go/zmscli/cli.go
```diff
@@ -46,6 +49,11 @@ type Zms struct {
 	AddSelf          bool
 }
 
+type SuccessMessage struct {
```
OferLevi85, line 52: Struct for success message. For now I'm planning to leave the original text in the "Message" field as it might be in use for users.
Example success Json:
` $ target/darwin/zms-cli -o json  -d home.test add-member admin user.example
{
    "Status": 200,
    "Message": "[Added to admin: user.example]"
}`

Example Yaml:
`$ target/darwin/zms-cli -o yaml -d home.test add-member admin user.example
status: 200
message: '[Added to admin: user.example]'`

libs/go/zmscli/role.go
```diff
@@ -231,7 +246,13 @@ func (cli Zms) AddMembers(dn string, rn string, members []string) (*string, erro
 	} else {
 		s = "[Added to " + rn + ": " + strings.Join(ms, ",") + "]"
 	}
-	return cli.switchOverFormats(s)
+
+	message := SuccessMessage{
+		Status:  200,
+		Message: s,
+	}
+
```
OferLevi85, line 254: For "old Yaml" we just printed out the error message without format (unlike in ShowRole function).
So instead of wrapping this logic as the "manual yaml converter", I passed the regular cli.buildYAMLOutput so it will appear in Yaml format.

libs/go/zmscli/cli.go
havetisyan: we should call the format as yaml and then update zms-cli to default to manualYaml (the old format). This way it allows users to start using -o yaml right now with the new format and not change later once it becomes the default option

libs/go/zmscli/cli.go
havetisyan: looks good!

libs/go/zmscli/cli.go
OferLevi85: Fixed

libs/go/zmscli/domain.go
```diff
@@ -38,13 +39,95 @@ func (cli Zms) DeleteDomain(dn string) (*string, error) {
...
+			return cli.dumpByFormat(message, cli.buildYAMLOutput)
 		}
 	}
 	return nil, err
 }
 
 func (cli Zms) ImportDomain(dn string, filename string, admins []string) (*string, error) {
+
+	if cli.OutputFormat == JSONOutputFormat || cli.OutputFormat == YAMLOutputFormat {
+		return cli.ImportDomainNew(dn, filename, admins, true)
```
OferLevi85, line 55: The import domain/update domain / export domain didn't work properly for Json (The json only included the "domain" responses without "roles / policies / services").
For these commands, I split the functions into "Old" versions and "New" versions. The "Old" versions are only used for "manual yaml" and will be removed once we remove the "manual yaml" functionality.

havetisyan (approved):

PR #1528, opened 2021-06-27
assertion condition server changes

abvaidya:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/QuotaChecker.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/QuotaChecker.java
@@ -25,7 +25,8 @@ class QuotaChecker {
 
     private final Quota defaultQuota;
     private boolean quotaCheckEnabled;
-    
+    int assertionConditionsQuota = Integer.parseInt(System.getProperty(ZMSConsts.ZMS_PROP_QUOTA_ASSERTION_CONDITIONS, "10"));
+
     public QuotaChecker() {
         
         // first check if the quota check is enabled or not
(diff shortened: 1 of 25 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/QuotaChecker.java
```diff
@@ -378,4 +379,44 @@ void checkEntityQuota(ObjectStoreConnection con, String domainName, Entity entit
...
+        }
+
+        // if our assertionConditions is null then there is no quota check
+        if (assertionConditions == null || assertionConditions.getConditionsList() == null ||
+                assertionConditions.getConditionsList().isEmpty()) {
+            return;
+        }
+
+        // we're going to check if we'll be allowed to create given assertionConditions
+        countAssertionConditions(con, assertionId, assertionConditions.getConditionsList().size(), caller);
```
havetisyan, line 397: in the method assertionConditions.getConditionsList().size() is treated as newCount - but is it the new count? If i'm updating the condition, does the method correctly handle that there are no new conditions so it's fine?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -522,6 +524,36 @@ void convertToLowerCase(Object obj) {
...
                 }
             }
+        },
+        ASSERTION_CONDITIONS {
+            @Override
+            void convertToLowerCase(Object obj) {
+                AssertionConditions assertionConditions = (AssertionConditions) obj;
+                // lower case key and value, assertion condition operator is an ENUM, hence left unchanged
+                assertionConditions.setConditionsList(assertionConditions.getConditionsList().stream()
+                        .map(c -> new AssertionCondition()
```
havetisyan, line 534: why are we creating new assertion conditions rather than just updating the values in place like for all objects?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -522,6 +524,36 @@ void convertToLowerCase(Object obj) {
...
+            }
+        },
+        ASSERTION_CONDITION {
+            @Override
+            void convertToLowerCase(Object obj) {
+                AssertionCondition assertionCondition = (AssertionCondition) obj;
+                // lower case key and value, assertion condition operator is an ENUM, hence left unchanged
+                if (assertionCondition.getConditionsMap() != null) {
+                    assertionCondition.setConditionsMap(assertionCondition.getConditionsMap().entrySet().stream()
+                            .collect(Collectors.toMap(
```
havetisyan, line 551: we should not create a new map. we can just iterate through and update values in place. let's not create new objects if not necessary

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5711,7 +5724,7 @@ JWSDomain retrieveJWSDomain(final String domainName) {
             LOG.debug("retrieveJWSDomain: retrieving domain {}", domainName);
         }
 
-        AthenzDomain athenzDomain = getAthenzDomain(domainName, true, false);
+        AthenzDomain athenzDomain = getAthenzDomain(domainName, true);
```
havetisyan, line 5727: why was the master flag set to false removed here since it ends up calling the same method?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -529,30 +529,22 @@ void convertToLowerCase(Object obj) {
...
             void convertToLowerCase(Object obj) {
                 AssertionCondition assertionCondition = (AssertionCondition) obj;
                 // lower case key and value, assertion condition operator is an ENUM, hence left unchanged
-                if (assertionCondition.getConditionsMap() != null) {
-                    assertionCondition.setConditionsMap(assertionCondition.getConditionsMap().entrySet().stream()
-                            .collect(Collectors.toMap(
-                                    e -> e.getKey().toLowerCase(),
-                                    e -> e.getValue().setValue(e.getValue().getValue().toLowerCase())
-                            )));
+                Map<String, AssertionConditionData> lowerCasedMap = new HashMap<>();
```
havetisyan, line 542: Here we're still creating a new map set. My original comment was whether or not we can implement this without creating anything new. If that's not possible, then the original implementation using streams is preferred. also, the original code correctly checked for null for conditions map while this version does not.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
abvaidya: validator should take care of null since the individual properties in AssertionCondition are mandatory.

havetisyan (approved):

PR #1531, opened 2021-06-30
configurable location for changelog file store

abvaidya:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/ServerCommonConsts.java
```diff
@@ -38,6 +38,7 @@
     public static final String PROP_ATHENZ_CONF             = "athenz.athenz_conf";
     public static final String ZTS_PROP_FILE_NAME           = "athenz.zts.prop_file";
     public static final String PROP_GET_CONDITIONS_IN_SIGNED_DOMAINS = "athenz.server_common.get_conditions_in_signed_domains";
+    public static final String PROP_DATA_STORE_ROOT_DIR = "athenz.server_common.data_store_root_dir";
```
havetisyan, line 41: it's somewhat misleading to call it root_dir since it's not. before we're adding this to the home dir to get the dir. So we should call this PROP_DATA_STORE_SUBDIR since we're only looking for the subdirectory name

havetisyan (approved):

PR #1536, opened 2021-07-07
Remove grunt dependency from nodejs clients zpe / zts and from nodejs auth_core library

OferLevi85:


clients/nodejs/zpe/Makefile
```diff
@@ -10,16 +10,26 @@ NPM := $(shell command -v npm 2> /dev/null)
 
 ifdef NPM
 
-all:
-	npm install
```
OferLevi85, line 14: For all three "nodejs" projects I did the same change:
1. Replace grunt lint / test calls with regular direct calls from "make" as we did in UI (update package.json + Makefile)
2. Run fix-lint which makes the output a bit different than grunt lint which explains the indentation changes in the project files

pratikgote (approved):
👍

PR #1545, opened 2021-07-11
zts endpoint: Get roles which allow fetching role certs dynamically

OferLevi85:

### DIFF
--- a/clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
+++ b/clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
@@ -2291,6 +2291,22 @@ public class ZTSClient implements Closeable {
     }
 
     /**
+     * Fetch all roles that are tagged as requiring role certificates for principal
+     * @param principal Name of the principal
+     * @return list of roles that are tagged as requiring role certificates for the principal
+     */
+    public RoleAccess getRolesRequireRoleCert(String principal) {
+        updateServicePrincipal();
+        try {
+            return ztsClient.getRolesRequireRoleCert(principal);
+        } catch (ResourceException ex) {
+            throw new ZTSClientException(ex.getCode(), ex.getData());
+        } catch (Exception ex) {
+            throw new ZTSClientException(ResourceException.BAD_REQUEST, ex.getMessage());
+        }
+    }
+
+    /**
      * For a given provider domain get a list of tenant domain names that the user is a member of
      * @param providerDomainName name of the provider domain
      * @param userName is the name of the user to search for in the tenant domains of the provider
--- a/core/zts/src/main/rdl/RoleCert.rdli
+++ b/core/zts/src/main/rdl/RoleCert.rdli
@@ -16,3 +16,16 @@ resource RoleCertificate POST "/rolecert" (name=PostRoleCertificateRequestExt) {
         ResourceError UNAUTHORIZED;
     }
 }
+
+// Fetch all roles that are tagged as requiring role certificates for principal
+resource RoleAccess GET "/role/cert?principal={principal}" (name=getRolesRequireRoleCert) {
+    EntityName principal (optional); //If not present, will return roles for the user making the call
+    authenticate;
+    exceptions {
+        ResourceError BAD_REQUEST;
+        ResourceError NOT_FOUND;
+        ResourceError FORBIDDEN;
+        ResourceError UNAUTHORIZED;
+        ResourceError TOO_MANY_REQUESTS;
+    }
+}
--- a/core/zts/src/main/rdl/RoleCert.tdl
+++ b/core/zts/src/main/rdl/RoleCert.tdl
@@ -17,3 +17,7 @@ type RoleCertificateRequest Struct {
     Timestamp prevCertNotBefore (optional); //previous role certificate not before date
     Timestamp prevCertNotAfter (optional); //previous role certificate not after date
 }
+
+type RoleAccess Struct {
+    Array<EntityName> roles;
+}
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -2338,6 +2338,35 @@ public class ZTSImpl implements KeyStore, ZTSHandler {
     }
 
     @Override
+    public RoleAccess getRolesRequireRoleCert(ResourceContext ctx, String principal) {
+        final String caller = ctx.getApiName();
+
+        final String principalDomain = logPrincipalAndGetDomain(ctx);
+        validateRequest(ctx.request(), principalDomain, caller);
+        validate(principal, TYPE_ENTITY_NAME, principalDomain, caller);
+
+        if (StringUtil.isEmpty(principal)) {
+            // If principal not specified, get roles for current user
+            principal = ((RsrcCtxWrapper) ctx).principal().getFullName();
+        }
+
+        // for consistent handling of all requests, we're going to convert
+        // all incoming object values into lower case since ZTS Server
+        // saves all of its object names in lower case
+
+        principal = principal.toLowerCase();
+
+        List<String> rolesRequireRoleCert = dataStore.getRolesRequireRoleCert(principal);
+        if (rolesRequireRoleCert == null) {
+            rolesRequireRoleCert = new ArrayList<>();
+        }
+
+        RoleAccess roleList = new RoleAccess();
+        roleList.setRoles(rolesRequireRoleCert);
+        return roleList;
+    }
+
+    @Override
     public Workloads getWorkloadsByService(ResourceContext ctx, String domainName, String serviceName) {
 
         final String caller = ctx.getApiName();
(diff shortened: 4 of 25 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
```diff
@@ -73,6 +70,7 @@
     final Cache<String, PublicKey> zmsPublicKeyCache;
     final Cache<String, List<GroupMember>> groupMemberCache;
     final Cache<String, List<GroupMember>> principalGroupCache;
+    final Map<String, Role> requireRoleCertCache;
```
OferLevi85, line 73: Considered using a principal -> "role tagged as dynamic role cert fetch" cache but this meant that for every change in any role I'll need to go over the entire cache to search for members and fix entries accordingly.
I expect the refresh cert calls / bootstrap calls load will be much lighter so Implemented the following solution:
1.  Created a cache that contains all tagged roles. For every role change we just need to check a specific role and not the entire cache.
2. On certificate refresh / bootstrap we'll go over the relatively small roles list that contains the principal as a member and return it.

core/zts/src/main/rdl/RoleCert.tdl
```diff
@@ -17,3 +17,8 @@ type RoleCertificateRequest Struct {
     Timestamp prevCertNotBefore (optional); //previous role certificate not before date
     Timestamp prevCertNotAfter (optional); //previous role certificate not after date
 }
+
+//The representation for an enumeration of roles
```
havetisyan, line 21: Let's not create another object here. We already have one called RoleAccess with the same structure. That object is returned to indicate the list of roles the principal has access to in a domain and it's pretty much the same use.

core/zts/src/main/rdl/RoleCert.rdli
```diff
@@ -16,3 +16,16 @@ resource RoleCertificate POST "/rolecert" (name=PostRoleCertificateRequestExt) {
         ResourceError UNAUTHORIZED;
     }
 }
+
+// Fetch all roles that are tagged as requiring role certificates for service
+resource RoleList GET "/rolecert?service={service}" (name=getRolesRequireRoleCert) {
```
havetisyan, line 21: we already use POST /rolecert to request role certs so with the same URI, it's somewhat not clear that we're asking for a list of roles that require role certs. Maybe we separate the uri: GET /role/cert?...

core/zts/src/main/rdl/RoleCert.rdli
```diff
@@ -16,3 +16,16 @@ resource RoleCertificate POST "/rolecert" (name=PostRoleCertificateRequestExt) {
...
+    ResourceName service;
```
havetisyan, line 22: this should be:

EntityName principal;

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
```diff
@@ -953,6 +969,11 @@ void deleteDomain(final String domainName) {
         changeLogStore.removeLocalDomain(domainName);
     }
 
+    private void processDeleteRolesRequireRoleCert(String domainName) {
+        List<String> rolesToDelete = requireRoleCertCache.asMap().keySet().stream().filter(key -> key.startsWith(domainName)).collect(Collectors.toList());
```
havetisyan, line 973: this kind of logic with starts with will not work when you  have domains that include the same prefix - e.g. yahoo and yahoosports - if you delete yahoo then you'll also match yahoosports as well. In this cases you want to add the domain separator "yahoo." and then do the startsWith search

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
```diff
@@ -1550,6 +1571,23 @@ public void setCloudStore(CloudStore cloudStore) {
         return domainData.getRoles();
     }
 
+    public List<String> getRolesRequireRoleCert(String principal) {
```
havetisyan, line 1574: This implementation is not going to work at large scale. Image if we have thousands of roles setup with the option. It means for every request we're going to iterate through the list of roles just to find a match. Instead we should have a cache with memberName -> roleName. 

However, maintaining such a cache has its own issues where we need to keep updating the cache when the role is updated so we should revisit this logic and see how we should handle it. Maybe write a small design doc and we can discuss the implementation there.

We must also handle the case where the principal includes wildcards. So if I have a role member as athenz.* then any service within athenz domain should match.

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
havetisyan: If we look at the current number of role certificates we issue, it's not a small list. So we need to find a solution to handle these requests efficiently

clients/java/zts/core/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
```diff
@@ -2278,6 +2278,22 @@ public ServiceIdentityList getServiceIdentityList(String domainName) {
         }
     }
 
+    /**
+     * Fetch all roles that are tagged as requiring role certificates for service
+     * @param service
```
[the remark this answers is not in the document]
OferLevi85: Fixed

core/zts/src/main/rdl/RoleCert.tdl
OferLevi85: Moved RoleAccess from Token.rdli to RoleCert.tdl so it can be used by both (Token.rdli already includes RoleCert.tdl)

core/zts/src/main/rdl/RoleCert.rdli
OferLevi85: Fixed

core/zts/src/main/rdl/RoleCert.rdli
OferLevi85: Fixed

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
OferLevi85: Sure thing, working on it.

servers/zts/src/main/java/com/yahoo/athenz/zts/store/RolePrefixTrie.java
```diff
@@ -0,0 +1,122 @@
...
+
+    public Set<String> findRoles(String principal) {
+        try {
+            readLock.lock();
+            Set<String> roles = new HashSet<>();
+            TrieNode current = root;
+            final String[] principalSections = principal.split("\\.");
+
+            for (int i = 0; i < principalSections.length; i++) {
+                if (isLastSection(principalSections, i)) {
```
dvirguttman, line 95: why not just changing the for loop condition to `principalSections.length - 1` ?

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
OferLevi85: Please see the updated version. I'm keeping cache in 3 data structures:
1. For regular members - same as for groups using the existing methods
2. For wildcard members - Set of all roles where a wildcard is a member
3. For prefix members - Use a prefix trie so finding a match to a given principal is efficient.

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
OferLevi85: Please see the updated version. I'm keeping cache in 3 data structures:

For regular members - same as for groups using the existing methods
For wildcard members - Set of all roles where a wildcard is a member
For prefix members - Use a prefix trie so finding a match to a given principal is efficient.

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
OferLevi85: Please see the latest changes (logic in cleanRequireRoleCertCache in DataStore.java).

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2329,6 +2329,30 @@ public RoleCertificate postRoleCertificateRequestExt(ResourceContext ctx, RoleCe
         return roleCertificate;
     }
 
+    @Override
+    public RoleAccess getRolesRequireRoleCert(ResourceContext ctx, String principal) {
+        final String caller = ctx.getApiName();
+
+        final String principalDomain = logPrincipalAndGetDomain(ctx);
+        validateRequest(ctx.request(), principalDomain, caller);
+        validate(principal, TYPE_ENTITY_NAME, principalDomain, caller);
```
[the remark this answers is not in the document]
OferLevi85: I didn't mark it as optional In RoleCert.rdli but I understand it is only enforced for path parameters.
In any case I now made it optional and if it isn't specified I get the roles for the current user (similar to getPrincipalRoles in ZMSImpl).

havetisyan (approved):

PR #1555, opened 2021-07-18
Extending Microsegmentation daemon api

psasidhar:


core/msd/src/main/rdl/Workload.tdl
```diff
@@ -14,9 +14,20 @@ type Workload Struct {
...
 }
 
 // list of workloads
 type Workloads Struct {
-	Array<Workload> workloadList; // list of workloads
+    Array<Workload> workloadList; // list of workloads
+}
+
+// WorkloadRecord type describes the workload to be saved in the storage
+type WorkloadRecord Workload {
```
abvaidya, line 30: why do we need this defined in RDL?

abvaidya (changes requested):

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/workload/WorkloadData.java
```diff
@@ -0,0 +1,44 @@
+package com.yahoo.athenz.common.server.workload;
```
havetisyan, line 1: need copyright

havetisyan (changes requested):

havetisyan:
@psasidhar two files still missing copyright details

core/msd/src/main/rdl/Workload.tdl
psasidhar: removed it

havetisyan (approved):

PR #1559, opened 2021-07-25
UI: Add case-sensitive checkbox for policy assertions

OferLevi85:

### DIFF
--- a/ui/src/components/policy/AddRuleForRoleForm.js
+++ b/ui/src/components/policy/AddRuleForRoleForm.js
@@ -21,6 +21,7 @@ import RadioButtonGroup from '../denali/RadioButtonGroup';
 import { colors } from '../denali/styles';
 import RequestUtils from '../utils/RequestUtils';
 import Color from '../denali/Color';
+import CheckBox from '../denali/CheckBox';
 
 const SectionsDiv = styled.div`
     width: 100%;
@@ -78,6 +79,7 @@ export default class AddRuleForRoleForm extends React.Component {
             name: '',
             action: '',
             resource: '',
+            case: false,
         };
         this.getRoles();
     }
@@ -111,15 +113,22 @@ export default class AddRuleForRoleForm extends React.Component {
     }
 
     inputChanged(key, evt) {
-        this.setState({ [key]: evt.target.value });
-        this.props.onChange(key, evt.target.value);
+        if (key == 'case') {
+            this.setState({ [key]: evt.target.checked });
+            this.props.onChange(key, evt.target.checked);
+        } else {
+            this.setState({ [key]: evt.target.value });
+            this.props.onChange(key, evt.target.value);
+        }
     }
 
     render() {
         let policyNameChanged = this.inputChanged.bind(this, 'name');
         let resourceChanged = this.inputChanged.bind(this, 'resource');
         let actionChanged = this.inputChanged.bind(this, 'action');
+        let caseChanged = this.inputChanged.bind(this, 'case');
         let policy = ``;
+        let id = this.props.isPolicy ? 'new-policy' : this.props.id;
         if (this.props.isPolicy) {
             policy = (
                 <SectionDiv>
@@ -169,7 +178,7 @@ export default class AddRuleForRoleForm extends React.Component {
                             name='rule-action'
                             value={this.state.action}
                             onChange={actionChanged}
-                            placeholder={'Rule Action (Case Sensitive)'}
+                            placeholder={'Rule Action'}
                         />
                     </ContentDiv>
                 </SectionDiv>
@@ -183,7 +192,19 @@ export default class AddRuleForRoleForm extends React.Component {
                             name='rule-resource'
                             value={this.state.resource}
                             onChange={resourceChanged}
-                            placeholder={'Rule Resource (Case Sensitive)'}
+                            placeholder={'Rule Resource'}
+                        />
+                    </ContentDiv>
+                </SectionDiv>
+                <SectionDiv>
+                    <ContentDiv>
+                        <CheckBox
+                            checked={this.state.case}
+                            name={'checkbox-case-sensitive' + id}
+                            id={'checkbox-case-sensitive' + id}
+                            key={'checkbox-case-sensitive' + id}
+                            label='Case Sensitive Action and Resource'
+                            onChange={caseChanged}
                         />
                     </ContentDiv>
                 </SectionDiv>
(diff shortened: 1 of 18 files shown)


ui/src/components/role-policy/AddRuleFormForRole.js
```diff
@@ -91,14 +93,20 @@ export default class AddRuleFormForRole extends React.Component {
     }
 
     inputChanged(key, evt) {
-        this.setState({ [key]: evt.target.value });
-        this.props.onChange(key, evt.target.value);
+        if (key == 'case') {
+            this.setState({ [key]: !this.state.case });
```
chandrasekhar1996, line 97: evt.target.checked can be used here

ui/src/components/policy/AddRuleForm.js
```diff
@@ -122,14 +124,20 @@ export default class AddRuleForm extends React.Component {
     }
 
     inputChanged(key, evt) {
-        this.setState({ [key]: evt.target.value });
-        this.props.onChange(key, evt.target.value);
+        if (key == 'case') {
+            this.setState({ [key]: !this.state.case });
```
chandrasekhar1996, line 128: evt.target.checked

ui/src/components/role-policy/AddRuleFormForRole.js
```diff
@@ -167,7 +175,18 @@ export default class AddRuleFormForRole extends React.Component {
...
                             value={this.state.resource}
                             onChange={resourceChanged}
-                            placeholder={'Rule Resource (Case Sensitive)'}
+                            placeholder={'Rule Resource'}
+                        />
+                    </ContentDiv>
+                </SectionDiv>
+                <SectionDiv>
+                    <ContentDiv>
+                        <CheckBox
```
chandrasekhar1996, line 184: have to give a unique key to the checkbox component 
https://reactjs.org/docs/lists-and-keys.html#keys
This is one option to generate a unique id : https://www.geeksforgeeks.org/lodash-_-uniqueid-method/

ui/src/components/policy/AddRuleForm.js
```diff
@@ -209,7 +217,18 @@ export default class AddRuleForm extends React.Component {
...
                             value={this.state.resource}
                             onChange={resourceChanged}
-                            placeholder={'Rule Resource (Case Sensitive)'}
+                            placeholder={'Rule Resource'}
+                        />
+                    </ContentDiv>
+                </SectionDiv>
+                <SectionDiv>
+                    <ContentDiv>
+                        <CheckBox
```
chandrasekhar1996, line 226: unique key

ui/src/components/policy/AddRuleForRoleForm.js
```diff
@@ -183,7 +192,18 @@ export default class AddRuleForRoleForm extends React.Component {
...
                             value={this.state.resource}
                             onChange={resourceChanged}
-                            placeholder={'Rule Resource (Case Sensitive)'}
+                            placeholder={'Rule Resource'}
+                        />
+                    </ContentDiv>
+                </SectionDiv>
+                <SectionDiv>
+                    <ContentDiv>
+                        <CheckBox
```
chandrasekhar1996, line 201: unique key

ui/src/components/policy/AddRuleForRoleForm.js
```diff
@@ -17,10 +17,12 @@ import React from 'react';
 import styled from '@emotion/styled';
 import InputLabel from '../denali/InputLabel';
 import Input from '../denali/Input';
+import InputDropdown from '../denali/InputDropdown';
```
chandrasekhar1996, line 20: not required

ui/src/components/policy/AddRuleForRoleForm.js
OferLevi85: Not sure what you mean by "unique key"

ui/src/components/policy/AddRuleForm.js
OferLevi85: Why use it? I'll still need to check the key

ui/src/components/policy/AddRuleForm.js
OferLevi85: Not sure what that means

ui/src/components/role-policy/AddRuleFormForRole.js
OferLevi85: Why use it? I'll still need to check the key

chandrasekhar1996: setState is an asynchronous function, 
here you are setting state of 'case' by accessing the current state of 'case' which may not be updated from a previous call. so this can lead to wrong values being set and not showing up properly in UI.
This would be rare but still a possibility and setting the value using event object would be the best way.

OferLevi85: Got it. Thanks for the explanation!

ui/src/components/role-policy/AddRuleFormForRole.js
chandrasekhar1996: also to check the effect of it, in the policy page open add rule forms of two different policies and try clicking on the checkbox.

ui/src/components/role-policy/AddRuleFormForRole.js
OferLevi85: Fixed

ui/src/components/policy/AddRuleForm.js
OferLevi85: Fixed

ui/src/components/role-policy/AddRuleFormForRole.js
OferLevi85: Fixed

ui/src/components/policy/AddRuleForm.js
OferLevi85: Fixed

ui/src/components/policy/AddRuleForRoleForm.js
OferLevi85: Fixed

ui/src/components/policy/AddRuleForRoleForm.js
OferLevi85: Fixed

chandrasekhar1996 (approved):

PR #1561, opened 2021-07-27
For notifications - get recipient email from User Authority

OferLevi85:

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationToEmailConverterCommon.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationToEmailConverterCommon.java
@@ -16,6 +16,7 @@
 
 package com.yahoo.athenz.common.server.notification;
 
+import com.yahoo.athenz.auth.Authority;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -40,6 +41,7 @@ public class NotificationToEmailConverterCommon {
     private static final String PROP_NOTIFICATION_ATHENZ_UI_URL = "athenz.notification_athenz_ui_url";
     private static final String PROP_NOTIFICATION_SUPPORT_TEXT = "athenz.notification_support_text";
     private static final String PROP_NOTIFICATION_SUPPORT_URL = "athenz.notification_support_url";
+    private static final String PROP_NOTIFICATION_USER_AUTHORITY = "athenz.notification_user_authority";
     private static final String EMAIL_TEMPLATE_CSS = "emails/base.css";
 
     private static final String HTML_STYLE_TAG_START = "<style>";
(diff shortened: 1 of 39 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationToEmailConverterCommon.java
```diff
@@ -69,7 +72,26 @@ public NotificationToEmailConverterCommon() {
         supportUrl = System.getProperty(PROP_NOTIFICATION_SUPPORT_URL);
 
         emailBaseCSS = readContentFromFile(getClass().getClassLoader(), EMAIL_TEMPLATE_CSS);
+        this.notificationUserAuthority = loadNotificationUserAuthority();
```
OferLevi85, line 75: The PR main change is in this file:
1. Load User Authority
2. On getFullyQualifiedEmailAddresses, if User Authority is loaded get the email for the user.
The other classes were changes as I no longer instantiate NotificationToEmailConverterCommon for each of them - I just inject the same one.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationToEmailConverterCommon.java
```diff
@@ -69,7 +72,26 @@ public NotificationToEmailConverterCommon() {
...
+    }
+
+    private Authority loadNotificationUserAuthority() {
```
havetisyan, line 78: Was there any reason why we're not just passing the user authority object from the server to the notification interface so we don't have to maintain a separate setting just for the notification library and then load it again?

OferLevi85: There were a couple of reasons I decided to go this route:
1. For ZTS we will need a new setting in any case as we currently don't load a "UserAuthority" there (we load all the authorities but we don't keep a specific one for users like we do in ZMS).
2. I thought it might be possible that we will want to use a separate authority for Notifications than the one we usually use for authentication.
If you think it's best to inject the existing user authority let me know and I'll modify the "loadAuthorities" method in ZTSImpl to load it and later inject it there and from zms.

havetisyan: My concern is that I might have some cases where I don't want to instantiate multiple copies of my user authority class. Those could be doing some heavy operations to get the data, etc, so now we're doing twice the work.  So if there is a use case to have a different user authority then the implementation could be to look at the current setting - if one is defined then it will load that otherwise it should use whatever the server gave it - this way we're only using a single instance.

As far ZTS - we can just introduce the setting in ZTS and then pass the authority to the notification library.

OferLevi85: Sure I'll make the necessary changes.

havetisyan (changes requested):

havetisyan (approved):

PR #1603, opened 2021-08-25
MSD - API changes for static workload support

abvaidya:

When an appliance can not get Athenz service identity via provider powered bootstrap due to limitations, users can manually associate a FQDN ( or IP in some cases ) with the existing service using this new feature to support static workloads.

Changes:
new type StaticWorkload in MSD to support use cases where identity via provider bootstrap is not possible due to limitations
new putStaticWorkload API
Changes to the existing Workloads object to include List of StaticWorkload 
Rename of Workload to DynamicWorkload for clarity


core/msd/src/main/rdl/Workload.tdl
```diff
@@ -4,24 +4,44 @@
 //Name types
 include "Names.tdl";
 
-// workload type describing workload associated with an identity
-type Workload Struct {
-    DomainName domainName; // name of the domain, optional for getWorkloadsByService API call
-    EntityName serviceName; // name of the service, , optional for getWorkloadsByService API call
+type StaticWorkloadType Enum {
```
abvaidya, line 7: Tried to keep this as generic as possible.

core/msd/src/main/rdl/Workload.rdli
```diff
@@ -36,13 +36,29 @@ resource Workloads GET "/workloads/{ip}" (name=getWorkloadsByIP) {
...
+// Api to perform a dynamic workload PUT operation for a domain and service
 // Workload details are obtained from the service certificate
-resource WorkloadOptions PUT "/domain/{domainName}/service/{serviceName}/workload" (name=putWorkload) {
+resource WorkloadOptions PUT "/domain/{domainName}/service/{serviceName}/workload/dynamic" (name=putDynamicWorkload) {
     DomainName domainName; // name of the domain
     EntityName serviceName; // name of the service
-    WorkloadOptions options; // workload store to be inserted
-    authenticate;
+    WorkloadOptions options; // metadata about the dynamic workload
+    authorize ("update", "{domainName}:service.{serviceName}");
```
abvaidya, line 45: Updated authenticate -> authorize

core/msd/src/main/rdl/Workload.tdl
```diff
@@ -4,24 +4,44 @@
...
+
 type WorkloadOptions Struct {
     Bool ipChanged; // boolean flag to signal a change in IP state
 }
 
 // list of workloads
 type Workloads Struct {
-    Array<Workload> workloadList; // list of workloads
+    Array<DynamicWorkload> dynamicWorkloadList; // list of dynamic workloads
+    Array<StaticWorkload> staticWorkloadList; // list of static workloads
```
abvaidya, line 46: Since the Workloads object is changing anyways with Workload -> DynamicWorkload move, haven't kept staticWorkloadList as optional field.

core/msd/src/main/rdl/Workload.rdli
havetisyan: this doesn't look right. Isn't this operation carried out by the pinger? So why would the service itself have update right over itself? The correct authorization must be done in the code where we check that the principal of the client certificate is {domainName}.{serviceName}

core/msd/src/main/rdl/Workload.tdl
```diff
@@ -4,24 +4,44 @@
...
     Timestamp certIssueTime (optional); // certificate issue time (ex: getNotBefore)
 }
 
+// workload type describing workload indirectly associated with an identity ( without bootstrap )
+type StaticWorkload Struct {
+    DomainName domainName; // name of the domain
+    EntityName serviceName; // name of the service
+    StaticWorkloadType type; // value representing one of the StaticWorkloadType enum
+    Array<String> ipAddresses (optional); // list of IP addresses associated with the workload, optional for getWorkloadsByIP API call
+    String hostname (optional); // hostname associated with the workload
```
havetisyan, line 35: so if for the external appliance case - do we have a better option than a hostname since it doesn't have to be hostname and could be a role name? Maybe something more generic?

core/msd/src/main/rdl/Workload.tdl
havetisyan: Is the expectation that the server will just pass back an empty array?

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/msd/MsdStoreConnection.java
```diff
@@ -16,48 +16,52 @@
...
      * @param domain of the service to look up
      * @param service name of the service to look up
-     * @return a List of Workload
+     * @return Workloads object containing static and dynamic workloads
      */
-    default public List<Workload> getWorkloadsBySvc(String domain, String service) {
-        return new ArrayList<>();
+    default Workloads getWorkloadsBySvc(String domain, String service) {
+        Workloads workloads = new Workloads();
+        workloads.setDynamicWorkloadList(new ArrayList<>());
```
havetisyan, line 47: shouldn't we set the static list also to empty list since the rdl has it as mandatory?

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/msd/MsdStoreConnection.java
```diff
@@ -16,48 +16,52 @@
...
     /**
      * getWorkloadsByIp looks up all workloads for the requested ip in the MSD Storage
      * @param ip to lookup
      * @return a List of Workload
      */
-    default public List<Workload> getWorkloadsByIp(String ip) {
-        return new ArrayList<>();
+    default Workloads getWorkloadsByIp(String ip) {
+        Workloads workloads = new Workloads();
+        workloads.setDynamicWorkloadList(new ArrayList<>());
```
havetisyan, line 58: shouldn't we set the static list also to empty list since the rdl has it as mandatory?

havetisyan (changes requested):

core/msd/src/main/rdl/Workload.tdl
abvaidya: Thats the behavior today with dynamic workloads ( empty array ). Would you prefer to make staticWorkloadList  optional instead?

core/msd/src/main/rdl/Workload.tdl
havetisyan: sure, that's generic enough and in comments you can identity that in most cases this is going to be a hostname depending on the type.

core/msd/src/main/rdl/Workload.tdl
havetisyan: no, but there are a couple of places where we don't set it so we must need to make sure the server always returns an empty array.

PR #1605, opened 2021-08-28
Implement multi-version policies endpoints

OferLevi85:


servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -2603,6 +2618,32 @@ public boolean insertPolicy(String domainName, Policy policy) {
         return (affectedRows > 0);
     }
 
+    @Override
```
havetisyan, line 2621: Why do we have a separate API just for insertPolicyVersion? Isn't this the same as insertPolicy where policy.version is not null? So we can just update insertPolicy to make handle the version number accordingly. In the DB Service code we can make sure whenever we need to call just insertPolicy the version is set to null and we don't take the value from the user

servers/zms/src/main/java/com/yahoo/athenz/zms/store/ObjectStoreConnection.java
```diff
@@ -117,18 +117,22 @@
 
     // Policy commands
 
-    Policy getPolicy(String domainName, String policyName);
+    Policy getPolicy(String domainName, String policyName, String version);
     boolean insertPolicy(String domainName, Policy policy);
+    boolean insertPolicyVersion(String domainName, String version, Policy policy);
```
havetisyan, line 122: why do we have this api instead of just using insertPolicy? 

see my comments in the implementation

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -150,21 +150,28 @@
...
     private static final String SQL_UPDATE_POLICY = "UPDATE policy SET name=? WHERE policy_id=?;";
     private static final String SQL_UPDATE_POLICY_MOD_TIMESTAMP = "UPDATE policy "
             + "SET modified=CURRENT_TIMESTAMP(3) WHERE policy_id=?;";
-    private static final String SQL_GET_POLICY_ID = "SELECT policy_id FROM policy WHERE domain_id=? AND name=?;";
+    private static final String SQL_SET_ACTIVE_POLICY_VERSION = "UPDATE policy SET active = CASE WHEN version=? then true ELSE false END WHERE domain_id=? AND name=?;";
+    private static final String SQL_GET_ACTIVE_POLICY_ID = "SELECT policy_id FROM policy WHERE domain_id=? AND name=? AND active=true;";
+    private static final String SQL_GET_POLICY_VERSION_ID = "SELECT policy_id FROM policy WHERE domain_id=? AND name=? AND version=?;";
     private static final String SQL_DELETE_POLICY = "DELETE FROM policy WHERE domain_id=? AND name=?;";
+    private static final String SQL_DELETE_POLICY_VERSION = "DELETE FROM policy WHERE domain_id=? AND name=? AND version=? and active=false;";
     private static final String SQL_LIST_POLICY = "SELECT name FROM policy WHERE domain_id=? AND active=1";
```
havetisyan, line 166: can we also review this line. here we're using active=1 so let's make it consistent and use true/false

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -150,21 +150,28 @@
...
+    private static final String SQL_LIST_POLICY_VERSION = "SELECT version FROM policy WHERE domain_id=? AND name=?";
     private static final String SQL_COUNT_POLICY = "SELECT COUNT(*) FROM policy WHERE domain_id=?";
     private static final String SQL_LIST_ASSERTION = "SELECT * FROM assertion WHERE policy_id=?";
     private static final String SQL_COUNT_ASSERTION = "SELECT COUNT(*) FROM assertion WHERE policy_id=?";
     private static final String SQL_GET_ASSERTION = "SELECT * FROM assertion "
             + "JOIN policy ON assertion.policy_id=policy.policy_id "
             + "JOIN domain ON policy.domain_id=domain.domain_id "
-            + "WHERE assertion.assertion_id=? AND domain.name=? AND policy.name=?;";
+            + "WHERE assertion.assertion_id=? AND domain.name=? AND policy.name=? AND policy.active=true;";
```
havetisyan, line 174: Can we review this usage - this looks suspicious to me. for the given request we already have the assertion_id which means we know exactly what version of the policy this belongs to, so why are we adding an active check?

Looking at this a little bit more - this seems to be used in the delete assertion api call from zmsimpl/dbservice so how are we deleting assertions from non-active versions of the policy

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -978,6 +981,83 @@ boolean shouldRetryOperation(ResourceException ex, int retryCount) {
...
+                // open our audit record
+
+                auditDetails.append("{\"name\": \"").append(policyName);
+                auditDetails.append("\", \"version\": \"").append(version);
+
+                // now we need process our policy assertions
+
+                List<Assertion> newAssertions = originalPolicy.getAssertions();
+                if (newAssertions != null) {
+                    for (Assertion assertion : newAssertions) {
```
havetisyan, line 1031: for completeness sake we must copy all the assertion conditions as well when we're making a copy of the policy. at some point we might use this functionality in microsegmentation as well so we need to make sure all conditions are available as well.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1876,14 +2008,22 @@ void executeDeletePolicy(ResourceContext ctx, String domainName, String policyNa
 
                 // extract the current policy for audit log purposes
 
-                Policy policy = getPolicy(con, domainName, policyName);
-                if (policy == null) {
+                List<Policy> policyVersions = new ArrayList<>();
```
havetisyan, line 2011: this should be moved to line 2017 so we're not wasting creating a new list if versions ends up being null or empty and we return 404

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4393,6 +4390,184 @@ public PolicyList getPolicyList(ResourceContext ctx, String domainName, Integer
...
+
+    @Override
+    public Policy getPolicyVersion(ResourceContext ctx, String domainName, String policyName, String version) {
+        final String caller = ctx.getApiName();
+        logPrincipal(ctx);
+
+        validateRequest(ctx.request(), caller);
+        validate(domainName, TYPE_DOMAIN_NAME, caller);
+        validate(policyName, TYPE_ENTITY_NAME, caller);
+        validate(version, TYPE_ENTITY_NAME, caller);
```
havetisyan, line 4432: this must be TYPE_SIMPLE_NAME

havetisyan (changes requested):

core/zms/src/main/rdl/Policy.rdli
```diff
@@ -218,4 +219,91 @@ resource AssertionCondition DELETE "/domain/{domainName}/policy/{policyName}/ass
         ResourceError CONFLICT;
         ResourceError TOO_MANY_REQUESTS;
     }
+}
+
+//List policy versions.
+resource PolicyList GET "/domain/{domainName}/policy/{policyName}/version?limit={limit}&skip={skip}" (name=getPolicyVersionList) {
```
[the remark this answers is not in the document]
OferLevi85: Got it. Thought that because 3 is a configurable value we might want to keep this open for future updates but if we never planning on having more than a few versions I'll remove it.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
OferLevi85: Well at the moment we don't. I assumed that we can only modify active policy versions as the new endpoints didn't mention providing the option to add / remove assertions for non-active versions. So if we want to add or remove assertions we need to make the policy active first.
I understand now that this should be supported and will update the API accordingly

core/zms/src/main/rdl/Policy.rdli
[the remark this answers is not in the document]
OferLevi85: Fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
OferLevi85: Fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
OferLevi85: Fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
OferLevi85: Fixed (added 2 new endpoints to delete / put assertions for policy versions.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
OferLevi85: Fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
OferLevi85: Fxied

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -978,6 +983,111 @@ boolean shouldRetryOperation(ResourceException ex, int retryCount) {
...
+                        // insert assertion (and get new assertion id)
+
+                        if (!con.insertAssertion(domainName, policyName, version, assertion)) {
+                            con.rollbackChanges();
+                            throw ZMSUtils.internalServerError("unable to put policy: " + originalPolicy.getName() + ", version: " + version + ", fail inserting assertion", caller);
+                        }
+
+                        // copy assertion conditions for new assertion id
+
+                        if (assertionConditions != null && assertionConditions.getConditionsList() != null && !assertionConditions.getConditionsList().isEmpty()) {
```
havetisyan, line 1052: on line 1038 we have: AssertionConditions assertionConditions = new AssertionConditions(); 

so there is no need to call assertionConditions != null since it will never be null

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/store/ObjectStoreConnection.java
OferLevi85: fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
OferLevi85: fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -439,15 +444,15 @@ boolean processPolicy(ObjectStoreConnection con, Policy originalPolicy, String d
 
             if (!ignoreDeletes) {
                 for (Assertion assertion : delAssertions) {
-                    if (!con.deleteAssertion(domainName, policyName, assertion.getId())) {
+                    if (!con.deleteAssertion(domainName, policyName, null, assertion.getId())) {
```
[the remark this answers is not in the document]
OferLevi85: The function processPolicy is only called from APIs that use the active version only.
However, to prevent bugs in case the code changes in the future and this function will be used for both active and disable policy versions, I'll change it as requested.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -439,15 +444,15 @@ boolean processPolicy(ObjectStoreConnection con, Policy originalPolicy, String d
...
                         return false;
                     }
                 }
                 auditLogAssertions(auditDetails, "deleted-assertions", delAssertions);
             }
 
             for (Assertion assertion : addAssertions) {
-                if (!con.insertAssertion(domainName, policyName, assertion)) {
+                if (!con.insertAssertion(domainName, policyName, null, assertion)) {
```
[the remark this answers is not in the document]
OferLevi85: Same as in previous comment, the function processPolicy is only called from APIs that use the active version only.
However, to prevent bugs in case the code changes in the future and this function will be used for both active and disable policy versions, I'll change it as requested.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
OferLevi85: fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -999,7 +1109,7 @@ void executePutPolicy(ResourceContext ctx, String domainName, String policyName,
 
                 // retrieve our original policy
 
-                Policy originalPolicy = getPolicy(con, domainName, policyName);
+                Policy originalPolicy = getPolicy(con, domainName, policyName, null);
```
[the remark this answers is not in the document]
OferLevi85: The function executePutPolicy is only called from endpoints that use the active version.
I prefer not to use the version specified by the user because if he used these endpoints he meant to use the active version.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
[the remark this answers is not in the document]
havetisyan: @OferLevi85  That is not correct. processPolicy is called from executePutPolicy which is called from putPolicy. If I create a new version of the policy I can manage that version from either adding new assertions or just using putPolicy by providing the full object. In the putPolicy we don't have any restrictions where we check if you're modifying the active policy. So if the user uses that API to modify a non-active version, the code will update the active version which is not the user was doing. I'll merge the PR and will do a subsequent fix for this issue.

PR #1619, opened 2021-09-14
Domain changes messaging skeleton

dvirguttman:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/DomainChangeMessage.java
```diff
@@ -0,0 +1,156 @@
...
+    
+    private String domainName;
+    
+     // Domain change message id     
+    private String uuid;
+   
+    // The number of milliseconds since the epoch 
+    private long published;
+    
+    // Change in domain meta
```
havetisyan, line 33: @dvirguttman how about making this more generic? Instead of having each one as a separate field, why not just have a couple of attribute with enums with additional details. I'm thinking about:

objectType: { DOMAIN, ROLE, POLICY, SERVICE, ENTITY}
objectName: here we can be more specific -e.g. role or policy name based on the object Type
operation: {api name - putRole, putRoleMeta, putPolicy, etc} - this way I know what type of operation was carried out - .e.g if I have:

objectType: ROLE
objectName: admin
api: putRoleMeta

then I know that metadata for the admin role was modified. This should give us more details and flexibility to add details

havetisyan (approved):

PR #1622, opened 2021-09-15
repository and validator interfaces for MSD static workload implementation

abvaidya:
StaticWorkloadDataRepository represents various data repositories corresponding to different types of static workloads. ( CloudLBRepository, CloudNATRepository etc.)

StaticWorkloadValidator represents various validators corresponding to different types of static workloads. The validators use corresponding repository objects to validate against them.


havetisyan (changes requested):
1) we need all interfaces to be documented - javadocs
2) we also need unit tests for the no op implementation classes.

havetisyan (approved):

PR #1624, opened 2021-09-19
rdl publishChangeEvents generator

dvirguttman:


rdl/rdl-gen-athenz-server/main.go
```diff
@@ -258,6 +258,7 @@ public interface {{cName}}Handler {{openBrace}} {{range .Resources}}
     {{methodSig .}};{{end}}
     ResourceContext newResourceContext(HttpServletRequest request, HttpServletResponse response, String apiName);
     void recordMetrics(ResourceContext ctx, int httpStatus);
+    void publishChangeEvents(ResourceContext ctx, int httpStatus);
```
dvirguttman, line 261: Do we want to add if block to include only ZMSHandler?

rdl/rdl-gen-athenz-server/main.go
```diff
@@ -480,6 +481,9 @@ func (gen *javaServerGenerator) handlerBody(r *rdl.Resource) string {
 	s += "            }\n"
 	s += "        } finally {\n"
 	s += "            this.delegate.recordMetrics(context, code);\n"
+	if r.Method == "POST" || r.Method == "PUT" || r.Method == "DELETE" {
```
havetisyan, line 484: let's change the order of calling metrics and then publish, so that when we're recording metrics we also include the time we spent on publishing events.

havetisyan (changes requested):

rdl/rdl-gen-athenz-server/main.go
dvirguttman: sure,
I am about to push new commit soon with many changes..

libs/java/client_common/src/main/java/com/yahoo/athenz/common/messaging/DomainChangeMessage.java
```diff
@@ -23,12 +23,14 @@
...
         DOMAIN, 
-        ROLE, 
+        ROLE,
+        GROUP,
         POLICY, 
-        SERVICE, 
-        ENTITY
+        SERVICE,
+        ENTITY,
+        USER
```
havetisyan, line 33: Not sure if this is the right object since we don't manage USERs. When we manage users, it's typically to delete members from roles/groups, etc. So I don't expect any notifications where a user object was modified. The correct events would be either a role or a group was modified in a given domain.

servers/zms/src/main/java/com/yahoo/athenz/zms/messaging/DomainChangeMessageGenerator.java
```diff
@@ -0,0 +1,132 @@
+package com.yahoo.athenz.zms.messaging;
+
+import com.yahoo.athenz.common.messaging.DomainChangeMessage;
+
+import java.lang.reflect.InvocationTargetException;
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+public class DomainChangeMessageGenerator {
```
havetisyan, line 10: I'm not sure if I like the functionality in this class. Nowhere in the code we look at the URIs and make assumptions about the components and here looks like we're parsing and handling accordingly. This could lead to incorrect events possibly being generated. Also at this point, we have really verified and/or normalized the arguments so we don't want to base our events on not-normalized argument values.

I'd prefer that in each implementation method we actually create the event object since we're guaranteed to know exactly what the the event is.

havetisyan (changes requested):

libs/java/client_common/src/main/java/com/yahoo/athenz/common/messaging/DomainChangeMessage.java
```diff
@@ -23,12 +23,14 @@
...
         DOMAIN, 
-        ROLE, 
+        ROLE,
+        GROUP,
         POLICY, 
-        SERVICE, 
-        ENTITY
+        SERVICE,
+        ENTITY,
+        TEMPLATE
```
havetisyan, line 33: Template is also not an object type so I don't expect to notify on templates. Instead when applying templates, we'll be updating roles/policies/services/groups.

rdl/rdl-gen-athenz-server/main.go
```diff
@@ -479,11 +485,25 @@ func (gen *javaServerGenerator) handlerBody(r *rdl.Resource) string {
...
+	args := ""
+	for _, in := range inputs {
+		if string(in.Name) != "auditRef" {
+			args += string(in.Name) + ", "
+		}
+	}
+	args = strings.TrimSuffix(args, ", ")
+	return args
+}
+
```
dvirguttman, line 506: will be deleted in the next commit

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -657,6 +662,34 @@ public ZMSImpl() {
...
+        
+        // load the DomainChangePublisher
+        
+        loadDomainChangePublisher();
+    }
+
+    void loadDomainChangePublisher() {
+        String topicNames = System.getProperty(ZMS_PROP_DOMAIN_CHANGE_TOPIC_NAMES, "");
+        for (String topic : topicNames.split(",")) {
+            if (!topic.isEmpty()) {
```
havetisyan, line 674: instead of just calling topic.trim() on line 678, it's better that we trim it first, then check if it's empty, otherwise, we'll support values like: "message1, , message2" and we'll end up calling createPublisher with an empty string.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -9984,16 +10017,35 @@ public void run() {
...
+            LOG.warn("Got exception during recordMetrics: {}", e.getMessage(), e);
+        }
     }
 
+    @Override
+    public void publishChangeMessage(ResourceContext ctx, int httpStatus) {
+        try {
+            if (httpStatus == 200) {
+                for (DomainChangeMessage changeMessage : ctx.getDomainChangeMessages()) {
+                    for (ChangePublisher<DomainChangeMessage> publisher : domainChangePublishers) {
```
havetisyan, line 10039: if our configured topic names is empty, then we'll end up having a null domainChangePublishers which will cause null pointer exception so we should check if it's null or not at the beginning of the method.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -9984,16 +10017,35 @@ public void run() {
...
+            metric.stopTiming(timerMetric, domainName, principalDomainName, httpMethod, httpStatus, timerName);
+        } catch (Exception e) {
+            LOG.warn("Got exception during recordMetrics: {}", e.getMessage(), e);
+        }
     }
 
+    @Override
+    public void publishChangeMessage(ResourceContext ctx, int httpStatus) {
+        try {
+            if (httpStatus == 200) {
```
havetisyan, line 10037: with status only checking for 200, we'll miss notifications for created (201) and no content (204) which happens with deletes. Here we should check if httpStatus >= 200 && httpStatus <= 299

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -9984,16 +10017,35 @@ public void run() {
...
+                for (DomainChangeMessage changeMessage : ctx.getDomainChangeMessages()) {
+                    for (ChangePublisher<DomainChangeMessage> publisher : domainChangePublishers) {
+                        publisher.publish(changeMessage);
+                    }
+                }
+            }
+        } catch (Exception e) {
+            LOG.warn("Got exception during publishChangeMessage: {}", e.getMessage(), e);
```
havetisyan, line 10045: let's make it as error so they're logged in production otherwise we'll never know that something is broken

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -9984,16 +10017,35 @@ public void run() {
...
+            final String principalDomainName = getPrincipalDomain(ctx);
+            final String domainName = getRequestDomainName(ctx);
+            final Object timerMetric = getTimerMetric(ctx);
+            final String httpMethod = (ctx != null) ? ctx.getHttpMethod() : null;
+            final String apiName = (ctx != null) ? ctx.getApiName() : null;
+            final String timerName = (apiName != null) ? apiName + "_timing" : null;
+            metric.increment("zms_api", domainName, principalDomainName, httpMethod, httpStatus, apiName);
+            metric.stopTiming(timerMetric, domainName, principalDomainName, httpMethod, httpStatus, timerName);
+        } catch (Exception e) {
+            LOG.warn("Got exception during recordMetrics: {}", e.getMessage(), e);
```
havetisyan, line 10030: let's make it as error so they're logged in production otherwise we'll never know that something is broken

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
dvirguttman: added null check in the creation instead.

havetisyan: I was actually referring to the domainChangePublishers object itself. If the topic property setting is null or empty string, then domainChangePublishers itself is null. It only gets initialized if we have at least one entry. So at the beginning of the method we can just check 

if (domainChangePublishers == null) {
   return;
}

havetisyan:
@dvirguttman please see my reply about the null ptr check

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
dvirguttman: missed that ,thanks
fixed

PR #1629, opened 2021-09-27
Defining interfaces for consuming/publishing changes

psasidhar:

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangePublisherFactory.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangePublisherFactory.java
@@ -1,0 +1,29 @@
+/*
+ * Copyright The Athenz Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.yahoo.athenz.common.messaging;
+
+import com.yahoo.athenz.auth.PrivateKeyStore;
+
+public interface ChangePublisherFactory<T> {
+    /**
+     * creates a ChangePublisher
+     * @param keyStore private keystore object for fetching any secrets if needed
+     * @param topicName of the topic to subscribe to
+     * @return ChangePublisher that was just created
+     */
+     ChangePublisher<T> create(PrivateKeyStore keyStore, String topicName);
+}
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriber.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriber.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright The Athenz Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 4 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriber.java
```diff
@@ -0,0 +1,33 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.yahoo.athenz.common.messaging;
+
+import java.util.function.Consumer;
+
+/**
+ * Domain change subscriber interface
```
abvaidya, line 22: Domain is not necessary here

libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriberFactory.java
```diff
@@ -0,0 +1,33 @@
...
+ * limitations under the License.
+ */
+
+package com.yahoo.athenz.common.messaging;
+
+import com.yahoo.athenz.auth.PrivateKeyStore;
+
+public interface ChangeSubscriberFactory<T> {
+    /**
+     * creates a DomainChangeSubscriber
```
abvaidya, line 23: Domain is not necessary here

libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriberFactory.java
```diff
@@ -0,0 +1,33 @@
...
+     * @param keyStore private keystore object for fetching any secrets if needed
+     * @param topicName of the topic to subscribe to
+     * @param SubscriptionName of the individual subscriber
```
abvaidya, line 26: SubscriptionName - first char S should be in lower case

libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriberFactory.java
```diff
@@ -0,0 +1,33 @@
...
+     * @param type subscription mode as recognized by the underlying pub/sub system
+     * @return DomainChangeSubscriber that was just created
```
abvaidya, line 28: Domain is not necessary here

libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriberFactory.java
```diff
@@ -0,0 +1,33 @@
...
+     */
+    default ChangeSubscriber<T> create(PrivateKeyStore keyStore, String topicName, String SubscriptionName, String type) {
```
abvaidya, line 30: can we rename type to subscriptionType or something to make it explicit?

abvaidya (changes requested):

libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriberFactory.java
psasidhar: fixed the typo

libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangeSubscriberFactory.java
psasidhar: done

abvaidya (commented):
Do we need default interface methods if we are not providing any reference / no-op implementations?

abvaidya (approved):
LGTM. @havetisyan could you please review?

libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/ChangePublisherFactory.java
```diff
@@ -0,0 +1,29 @@
...
+ */
+
+package com.yahoo.athenz.common.messaging;
+
+import com.yahoo.athenz.auth.PrivateKeyStore;
+
+public interface ChangePublisherFactory<T> {
+    /**
+     * creates a ChangePublisher
+     * @param keyStore private keystore object for fetching any secrets if needed
```
dvirguttman, line 24: Why do we need it?
I think it should be implementation specific, and not part of the interface

havetisyan (approved):

PR #1637, opened 2021-10-03
Pulsar client implementation

dvirguttman:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/messaging/pulsar/PulsarChangeSubscriber.java
```diff
@@ -0,0 +1,64 @@
...
+  }
+
+  @Override
+  public void run() {
+    while (true) {
+      LOG.debug("looping over the consumer receive method");
+      try {
+        Message<byte[]> msg = this.pulsarConsumer.receive(1, TimeUnit.SECONDS);
+        if (msg != null) {
+          LOG.debug("received message: {}", new String(msg.getData()));
```
OferLevi85, line 52: surround with "if (LOG.isDebugEnabled())" to avoid creation of string in production (new String(msg.getData()))

libs/java/server_common/pom.xml
```diff
@@ -31,6 +31,11 @@
   </properties>
 
   <dependencies>
+    <dependency>
+      <groupId>org.apache.pulsar</groupId>
```
OferLevi85, line 35: What if we want to use a different messaging service such as RabbitMQ, ActiveMQ, or something else already in use by whoever using Athenz? Maybe we should make it configurable?

dvirguttman: This pulsar implementation implements:
`PulsarFactory<T> implements ChangePublisherFactory<T>, ChangeSubscriberFactory<T>`

We chose to add specific pub/sub implementation, pulsar, here.
Additional MQ implementation can be added easily.

OferLevi85: Got it

havetisyan:
@dvirguttman haven't looked at the the contents yet, but please add the Athenz Authors copyright to all new files.

dvirguttman:
> @dvirguttman haven't looked at the the contents yet, but please add the Athenz Authors copyright to all new files.

Thanks, done

havetisyan (approved):

PR #1638, opened 2021-10-04
Fix setting policy version active as null return 500 instead of 404 w…

OferLevi85:
…hen trying to create a new non active versioned policy


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4986,6 +4986,9 @@ public void putPolicy(ResourceContext ctx, String domainName, String policyName,
         setRequestDomain(ctx, domainName);
         policyName = policyName.toLowerCase();
         AthenzObject.POLICY.convertToLowerCase(policy);
+        if (policy.getActive() == null) {
```
havetisyan, line 4989: not sure if this is the right place for this update. this kind of breaks backward compatibility if the customer is using their own data store since it implies that all new policies are not active by default which is not correct. if the active field is not specified then the default is it is active. if we have a different logic when we specify a versioned policy then we should do the appropriate check possibly in DBService

havetisyan (changes requested):

havetisyan (approved):

PR #1641, opened 2021-10-05
Integrating msd client in UI

pratikgote:


ui/src/server/handlers/api.js
```diff
@@ -1291,7 +1291,7 @@ Fetchr.registerService({
 Fetchr.registerService({
     name: 'add-service-host',
     update(req, resource, params, body, config, callback) {
-        req.clients.zms.putServiceIdentity(
+        req.clients.msd.putServiceIdentity(
```
abvaidya, line 1294: is this intended change? MSD does not have this API.

abvaidya (changes requested):
Lint needs to be fixed.
Are we doing next 11 and react 17 upgrades in a separate PR?

PR #1671, opened 2021-11-07
Support priority handling for cert signing

OferLevi85:


havetisyan (changes requested):
we also need to update the sia-eks, sia-fargate and sia-azure vm agents with the logic to send the role cert details.

havetisyan (approved):

PR #1684, opened 2021-11-29
additional checks when convering sigs from der to p1363 format

havetisyan:


pratikgote (approved):
👍

PR #1693, opened 2021-12-09
Metrics for ZPU

OferLevi85:


utils/zpe-updater/cmd/tools/main.go
```diff
@@ -87,7 +90,24 @@ func main() {
...
 	}
 
-	// first check if we're just asked to view a local domain
+	// first, if running check we need to verify policy files
+	// validity and generate a json metrics that can be pushed to
+	// a monitoring service
+	if check {
+		policyStatus, err := zpu.CheckState(zpuConfig)
+		if err != nil {
+			log.Fatalf("Unable to check policy files state, %v", err)
```
havetisyan, line 99: I don't think we want the check function to make calls to log.Fatalf and exit. We should always just return a json output and the output would include the overall status code and the message in the response. Additionally, the return code would be an error so the caller can just look at the exit code and include the json in the message (so we always get back a json response)

utils/zpe-updater/cmd/tools/main.go
```diff
@@ -87,7 +90,24 @@ func main() {
...
+		}
+		policyMetrics := metrics.FormPolicyMetrics(policyStatus)
+		jsonMetrics, err := json.Marshal(policyMetrics)
```
havetisyan, line 102: Similarly here. If any of the policy files is not updated as expected, the return code should be 1 instead of 0 with the json output. It would simplify where I only need to check the return code from the call to see if everything is ok or not.

havetisyan (changes requested):

utils/zpe-updater/cmd/tools/main.go
OferLevi85: fixed

utils/zpe-updater/cmd/tools/main.go
OferLevi85: fixed

utils/zpe-updater/cmd/tools/main.go
```diff
@@ -87,7 +90,36 @@ func main() {
...
+		for _, policyMetric := range policyMetrics {
+			metricBytes, err := metrics.DumpMetric(policyMetric, nil)
+			if err != nil {
+				errorMessages = append(errorMessages, err)
+			}
+			bytes = append(bytes, metricBytes...)
+		}
+		statusBytes, ok, err := metrics.DumpStatus(errconv.Reduce(errorMessages))
+		if err != nil {
+			byteString := fmt.Sprintf("{\"application\":\"zpu-check\",\"status_code\":1,\"status_msg\":\"%s\"}", err.Error())
```
havetisyan, line 109: rather than having the json hardcoded here, let's move this into a method in the metrics file. This way the definition is in the same file so we know what we're generating something like:

byteString := metrics.GetFailedStatus(err.Error())

havetisyan (changes requested):
I'm missing where we're marking the check as failure if the expiry is less than we expected. I see where the expiry is calculated but no logic to mark it as failure if wasn't refreshed property.

utils/zpe-updater/cmd/tools/main.go
OferLevi85: fixed

OferLevi85:
> I'm missing where we're marking the check as failure if the expiry is less than we expected. I see where the expiry is calculated but no logic to mark it as failure if wasn't refreshed property.

fixed

havetisyan (approved):

PR #1705, opened 2021-12-16
Use Config Manager to refresh configuration

OferLevi85:
variables from AWS Parameter Store

### DIFF
--- a/containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
+++ b/containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
@@ -28,7 +28,7 @@ import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.yahoo.athenz.common.server.log.jetty.JettyConnectionLogger;
 import com.yahoo.athenz.common.server.log.jetty.JettyConnectionLoggerFactory;
-import com.yahoo.athenz.common.server.util.config.ConfigManager;
+import com.yahoo.athenz.common.server.util.config.providers.ConfigProviderAwsParametersStore;
 import com.yahoo.athenz.common.server.util.config.providers.ConfigProviderFile;
 import org.eclipse.jetty.deploy.DeploymentManager;
 import org.eclipse.jetty.deploy.PropertiesConfigurationManager;
@@ -67,6 +67,8 @@ import com.yahoo.athenz.auth.PrivateKeyStoreFactory;
 import com.yahoo.athenz.common.server.util.ConfigProperties;
 import com.yahoo.athenz.container.filter.HealthCheckFilter;
 import com.yahoo.athenz.container.log.AthenzRequestLog;
+
+import static com.yahoo.athenz.common.server.util.config.ConfigManagerSingleton.CONFIG_MANAGER;
 
 public class AthenzJettyContainer {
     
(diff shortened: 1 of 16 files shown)


containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -589,6 +591,23 @@ public static AthenzJettyContainer createJettyContainer() {
...
+        System.getProperties().remove("socksProxyHost");
+
+        // First load initial properties (to get parameter store path)
+        String propFile = System.getProperty(AthenzConsts.ATHENZ_PROP_FILE_NAME,
+                getRootDir() + "/conf/athenz/athenz.properties");
+        ConfigProperties.loadProperties(propFile);
+
+        // Manage AWS parameter store configurations (as the first config source, AWS parameters will take precedence over parameters
+        // configured in the properties file.
+        String awsParameterStorePath = System.getProperty(AthenzConsts.ATHENZ_PROP_AWS_PARAM_STORE_PATH, "athenz");
```
havetisyan, line 604: why do we have a default path here? by default the feature should not be enabled since we may not be running zms in aws.

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceAWSProvider.java
```diff
@@ -61,13 +64,17 @@
 
     static final String AWS_PROP_CERT_VALIDITY_STS_ONLY = "athenz.zts.aws_cert_validity_sts_only";
 
-    PublicKey awsPublicKey = null;      // AWS public key for validating instance documents
-    long bootTimeOffset;                // boot time offset in milliseconds
-    long certValidityTime;              // cert validity for STS creds only case
+    PublicKey awsPublicKey = null;           // AWS public key for validating instance documents
+    DynamicConfigLong bootTimeOffsetSeconds; // boot time offset in milliseconds
```
havetisyan, line 68: the variable name and comment do not match - is it seconds or milliseconds?

havetisyan (changes requested):
couple of initial comments.

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -67,6 +67,9 @@
 import com.yahoo.athenz.common.server.util.ConfigProperties;
 import com.yahoo.athenz.container.filter.HealthCheckFilter;
 import com.yahoo.athenz.container.log.AthenzRequestLog;
+import software.amazon.awssdk.utils.StringUtils;
```
havetisyan, line 70: for consistency, we should use the StringUtils from jetty and not from aws sdk

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -589,6 +592,25 @@ public static AthenzJettyContainer createJettyContainer() {
...
 
+    public static void initConfigManager() {
+        System.getProperties().remove("socksProxyHost");
+
+        // First load initial properties (to get parameter store path)
+        String propFile = System.getProperty(AthenzConsts.ATHENZ_PROP_FILE_NAME,
+                getRootDir() + "/conf/athenz/athenz.properties");
+        ConfigProperties.loadProperties(propFile);
+
+        String awsParameterStorePath = System.getProperty(AthenzConsts.ATHENZ_PROP_AWS_PARAM_STORE_PATH, "");
```
havetisyan, line 603: let's make the variable final and since we're checking for isEmpty below which handles nulls, there is no need to pass a default value

havetisyan (changes requested):
@OferLevi85 a couple of minor changes

havetisyan (approved):

PR #1718, opened 2021-12-28
sia agents - dynamic role cert

OferLevi85:

### DIFF
--- a/libs/go/sia/aws/agent/agent.go
+++ b/libs/go/sia/aws/agent/agent.go
@@ -91,11 +91,37 @@ func GetRoleCertificate(ztsUrl, svcKeyFile, svcCertFile string, opts *options.Op
 		return false
 	}
 
-	//initialize our return state to success
 	failures := 0
+	if opts.DynamicRoleCerts {
+		//fetch dynamic roles
+		fullServiceName := fmt.Sprintf("%s.%s", opts.Domain, opts.Services[0].Name)
+		log.Printf("Fetching dynamic role cert for service: %s", fullServiceName)
+		rolesRequireRoleCert, err := client.GetRolesRequireRoleCert(zts.EntityName(fullServiceName))
+		if err != nil {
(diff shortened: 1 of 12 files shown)


libs/go/sia/aws/agent/agent.go
```diff
@@ -91,11 +91,35 @@ func GetRoleCertificate(ztsUrl, svcKeyFile, svcCertFile string, opts *options.Op
 		return false
 	}
 
-	//initialize our return state to success
-	failures := 0
+	//fetch static roles
+	failures := fetchRoles(opts.Roles, opts, key, svcKeyFile, client)
```
havetisyan, line 95: I'm somewhat confused with the lines 95-101. 

1. In the config we claim that //if true, fetch role certs dynamically. Roles attribute will be ignored. Based on the comments and this code we're fetching static roles and then dynamic roles.

2. when we address the fetch dynamic role part - why aren't we checking if the option is enabled in the first place? it looks like we always fetch the dynamic role certs.

provider/azure/sia-vm/authn.go
```diff
@@ -77,12 +78,40 @@ func GetRoleCertificate(ztsUrl, svcKeyFile, svcCertFile string, opts *options.Op
 		return false
 	}
 
-	// initialize our return state to success
-	failures := 0
+	// fetch static roles
+	failures:= fetchRoles(opts.Roles, opts, key, provider, client)
```
havetisyan, line 82: same comments as the one for aws above.

havetisyan (changes requested):

libs/go/sia/aws/agent/agent.go
OferLevi85: Fixed

provider/azure/sia-vm/authn.go
OferLevi85: Fixed

PR #1727, opened 2022-01-02
role consistency check when deleting roles or adding assertions

mendi160:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -3829,7 +3829,38 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
             throw ZMSUtils.requestError("deleteRole: admin role cannot be deleted", caller);
         }
 
+        /* we are not going to allow any user to delete
+         * role that related to any assertion of any policy */
+
+        List<Policy> policyList = getPolicyList(domainName, caller);
+
+        if (policyList != null) {
+            validateRoleNotAssociatedToPolicy(policyList, roleName, domainName, caller);
+        }
+
         dbService.executeDeleteRole(ctx, domainName, roleName, auditRef, caller);
+    }
+
+    private List<Policy> getPolicyList(String domainName, String caller) {
+        AthenzDomain domain = getAthenzDomain(domainName, false);
+        if (domain == null) {
+            throw ZMSUtils.notFoundError(caller + ": Domain not found: '" + domainName + "'", caller);
+        }
+        return setupPolicyList(domain, true, true);
+    }
+
+    void validateRoleNotAssociatedToPolicy(List<Policy> policies, String roleName, String domainName, String caller) {
+       roleName = ResourceUtils.roleResourceName(domainName, roleName);
+        for (Policy policy : policies) {
+            if (policy.getAssertions() == null) {
+                continue;
+            }
+            for (Assertion assertion : policy.getAssertions()) {
+                if (assertion.getRole().equals(roleName)) {
+                    throw ZMSUtils.requestError("deleteRole: the role: " + roleName + " associated to policy: " + policy.name + ", it cannot be deleted", caller);
+                }
+            }
+        }
     }
 
     boolean isMemberOfRole(Role role, String member) {
@@ -4767,11 +4798,21 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         }
 
         // validate to make sure we have expected values for assertion fields
+        // - and also to make sure that the associated role exists
 
         validatePolicyAssertion(assertion, caller);
 
         dbService.executePutAssertion(ctx, domainName, policyName, null, assertion, auditRef, caller);
         return assertion;
+    }
+
+    private List<String> getRoleNameList(String domainName, String caller) {
+
+        List<String> names = dbService.listRoles(domainName);
+        if (names.isEmpty()) {
+            throw ZMSUtils.requestError("The domain: " + domainName + "has no roles", caller);
+        }
+        return names;
     }
 
     @Override
@@ -4814,11 +4855,25 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         }
 
         // validate to make sure we have expected values for assertion fields
+        // - and also to make sure that the associated role exists
 
         validatePolicyAssertion(assertion, caller);
 
         dbService.executePutAssertion(ctx, domainName, policyName, version, assertion, auditRef, caller);
         return assertion;
+    }
+
+    void validateRoleAssociatedIsExist(Set<String> roleNames, String lookingForRole, String domainName, String caller) {
+        String shortRoleName = ZMSUtils.removeDomainPrefix(lookingForRole, domainName, ROLE_PREFIX);
+        if (shortRoleName.lastIndexOf("*") != -1) {
+            if (!isWildCardMatch(roleNames, shortRoleName)) {
+                throw ZMSUtils.requestError("The role: " + lookingForRole + " that associated to an assertion does not match to any existing role", caller);
+            }
+        } else {
+            if (!roleNames.contains(shortRoleName)) {
+                throw ZMSUtils.requestError("The role: " + lookingForRole + " that associated to an assertion does not exist", caller);
+            }
+        }
     }
 
     public void deleteAssertion(ResourceContext ctx, String domainName, String policyName,
@@ -4904,12 +4959,19 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
             return;
         }
 
+        // send this to func 'validatePolicyAssertion' since it doesn't need to retrieve the roleList for every assertion
+        Set<String> roleNamesCache = new HashSet<>();
+
         for (Assertion assertion : assertions) {
-            validatePolicyAssertion(assertion, caller);
+            validatePolicyAssertion(assertion, caller, roleNamesCache);
         }
     }
 
     void validatePolicyAssertion(Assertion assertion, String caller) {
+        validatePolicyAssertion(assertion, caller, new HashSet<>());
+    }
+
+    void validatePolicyAssertion(Assertion assertion, String caller, Set<String> roleNamesSet) {
 
         // extract the domain name from the resource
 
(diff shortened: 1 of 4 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3829,9 +3829,32 @@ public void deleteRole(ResourceContext ctx, String domainName, String roleName,
...
+
+        Policies policies = getPolicies(ctx, domainName, true, true);
+
+        // size > 1 because of the admin policy
+        if (policies.getList().size() > 1) {
+            validateRoleNotAssociatedToPolicy(policies.getList(), roleName, domainName, caller);
+        }
+
         dbService.executeDeleteRole(ctx, domainName, roleName, auditRef, caller);
     }
```
mendi160, line 3844: In this function i don't check the case that the role that is about to delete is the last one that match to an assertion with a wild card. Do it?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3829,9 +3829,32 @@ public void deleteRole(ResourceContext ctx, String domainName, String roleName,
             throw ZMSUtils.requestError("deleteRole: admin role cannot be deleted", caller);
         }
 
+        /* we are not going to allow any user to delete
+         * role that related to any assertion of any policy */
+
+        Policies policies = getPolicies(ctx, domainName, true, true);
+
+        // size > 1 because of the admin policy
```
dvirguttman, line 3837: shouldn't we validate it is indeed admin policy?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4770,6 +4793,11 @@ public Assertion putAssertion(ResourceContext ctx, String domainName, String pol
 
         validatePolicyAssertion(assertion, caller);
 
+        // we are not going to allow any user to add/update
+        // an assertion that the associated role does not exist
+        Set<String> roleNames = new HashSet<>((getRoleList(ctx, domainName, null, null).getNames()));
```
dvirguttman, line 4798: `getRoleList` / `getNames` will always return non-null or empty values?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
mendi160: fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
mendi160: fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3829,9 +3829,31 @@ public void deleteRole(ResourceContext ctx, String domainName, String roleName,
             throw ZMSUtils.requestError("deleteRole: admin role cannot be deleted", caller);
         }
 
+        /* we are not going to allow any user to delete
+         * role that related to any assertion of any policy */
+
+        Policies policies = getPolicies(ctx, domainName, true, true);
```
havetisyan, line 3835: getPolicies is an API endpoint so we can't call it within another api endpoint. This will screw up our counters as each deleteRole API will also count as getPolicies API as well. Take a look at the implementation of the getPolicies API and replace with api call with an appropriate internal method implementation.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3829,9 +3829,31 @@ public void deleteRole(ResourceContext ctx, String domainName, String roleName,
...
+
+        if (policies != null && policies.getList() != null) {
+            validateRoleNotAssociatedToPolicy(policies.getList(), roleName, domainName, caller);
+        }
+
         dbService.executeDeleteRole(ctx, domainName, roleName, auditRef, caller);
     }
 
+    void validateRoleNotAssociatedToPolicy(List<Policy> policies, String roleName, String domainName , String caller) {
+        if (!roleName.startsWith(domainName)) {
```
havetisyan, line 3845: not sure why're doing this check here. this method is only called within deleteRole API and there the user is required to pass the role name only (any other invalid input will be rejected eventually). So we can just generate the roleResourceName without calling startsWith first.

One minor comment as well - remove the extra space after domainName in the signature.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4770,6 +4792,14 @@ public Assertion putAssertion(ResourceContext ctx, String domainName, String pol
 
         validatePolicyAssertion(assertion, caller);
 
+        // we are not going to allow any user to add/update
+        // an assertion that the associated role does not exist
+
+        RoleList roleList = getRoleList(ctx, domainName, null, null);
```
havetisyan, line 4798: Same comment here as above. getRoleList is an API endpoint. Let's replace this with an internal implementation and not call another API endpoint.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5026,6 +5077,15 @@ public void putPolicy(ResourceContext ctx, String domainName, String policyName,
 
         validatePolicyAssertions(policy.getAssertions(), caller);
 
+        // we are not going to allow any user to add/update a policy
+        // that one of its assertions has role associated that does not exist
+
+        RoleList roleList = getRoleList(ctx, domainName, null, null);
```
havetisyan, line 5083: Same comment here as above. getRoleList is an API endpoint. Let's replace this with an internal implementation and not call another API endpoint.

havetisyan, line 5083: Same comment here as above. getRoleList is an API endpoint. Let's replace this with an internal implementation and not call another API endpoint.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5026,6 +5077,15 @@ public void putPolicy(ResourceContext ctx, String domainName, String policyName,
...
+        if (roleList != null && roleList.getNames() != null) {
+            String finalDomainName = domainName;
+            policy.getAssertions().forEach(assertion -> validateRoleAssociatedIsExist(new HashSet<>((roleList.getNames())), assertion.getRole(), finalDomainName, policy.getName(), caller));
```
havetisyan, line 5086: policy.getAssertions() I believe could be null so we must check here before.

Take a look at implementing this functionality within validatePolicyAssertions/validatePolicyAssertion calls.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
havetisyan, line 4798: 1) Take a look at implementing this functionality within validatePolicyAssertions/validatePolicyAssertion calls.

2) the logic with if (roleList != null && roleList.getNames() != null) is actually incorrect since if the domain for some reason has no role then we're allowing it to add any assertion without any checks. If the domain has no roles then we must always reject all assertions since there is something wrong with the domain.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4817,10 +4847,31 @@ public Assertion putAssertionPolicyVersion(ResourceContext ctx, String domainNam
 
         validatePolicyAssertion(assertion, caller);
 
+        // we are not going to allow any user to add/update
+        //  an assertion that the associated role does not exist
+
+        RoleList roleList = getRoleList(ctx, domainName, null, null);
```
havetisyan, line 4853: Same comment as above for the putAssertion call.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4767,13 +4799,30 @@ public Assertion putAssertion(ResourceContext ctx, String domainName, String pol
...
         // validate to make sure we have expected values for assertion fields
+        // - and also to make sure that the associated role exists
 
         validatePolicyAssertion(assertion, caller);
 
         dbService.executePutAssertion(ctx, domainName, policyName, null, assertion, auditRef, caller);
         return assertion;
     }
 
+    private List<String> getRoleNameList(String domainName, String caller) {
```
havetisyan, line 4810: @mendi160 we don't really need this function. All this function can just be simply replaced by a single line:

dbService.listRoles(domainName)'

that gives all the roles for the domain without creating additional objects, etc.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4949,6 +5019,17 @@ void validatePolicyAssertion(Assertion assertion, String caller) {
             throw ZMSUtils.requestError("Assertion action contains control characters: "
                     + resource, caller);
         }
+
+        // we are not going to allow any user to add/update
+        // an assertion that the associated role does not exist
+        int endDomainName = assertion.getRole().indexOf(":");
+        String domain = assertion.getRole().substring(0, endDomainName);
```
havetisyan, line 5026: 1) we need to check here for -1. if the user just passes "role1" for the role name instead of "athenz:role.role1" then we'll get index out of bounds exception.

2) also please add a unit test case for that scenario.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3829,9 +3829,41 @@ public void deleteRole(ResourceContext ctx, String domainName, String roleName,
...
         dbService.executeDeleteRole(ctx, domainName, roleName, auditRef, caller);
     }
 
+    private List<Policy> getPolicyList(String domainName, String caller) {
+        Policies result = new Policies();
+        AthenzDomain domain = getAthenzDomain(domainName, false);
+        if (domain == null) {
+            throw ZMSUtils.notFoundError(caller + ": Domain not found: '" + domainName + "'", caller);
+        }
+        return result.setList(setupPolicyList(domain, true, true)).getList();
```
havetisyan, line 3850: why are creating a Policies object, do setList and then do getList()?

we should able just return setupPolicyList(domain, true, true) without creating the result object.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4767,13 +4798,23 @@ public Assertion putAssertion(ResourceContext ctx, String domainName, String pol
...
 
         validatePolicyAssertion(assertion, caller);
 
         dbService.executePutAssertion(ctx, domainName, policyName, null, assertion, auditRef, caller);
         return assertion;
     }
 
+    private List<String> getRoleNameList(String domainName, String caller) {
+
+        List<String> names = new ArrayList<>(dbService.listRoles(domainName));
```
havetisyan, line 4811: there is no need to actually create yet another array list here. Having just a simple:

List<String> names = dbService.listRoles(domainName);

is sufficient since the function already creates the list and returns to us.

havetisyan (changes requested):

havetisyan (approved):

PR #1739, opened 2022-01-06
UI: setup config for csp local source url instead of hardcoded localhost

havetisyan:


abvaidya (approved):

pratikgote (approved):
👍

PR #1773, opened 2022-02-04
Adding methods and a utility program to parse host document

psasidhar:

### DIFF
--- a/libs/go/sia/host/provider/provider.go
+++ b/libs/go/sia/host/provider/provider.go
@@ -1,0 +1,66 @@
+//
+// Copyright The Athenz Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package provider
+
+import (
+	"crypto"
+	"crypto/x509"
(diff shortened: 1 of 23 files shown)


utils/hostdoc/Makefile
```diff
@@ -0,0 +1,57 @@
+#
+# Makefile to build ZTS OIDC ID Token utility
```
havetisyan, line 2: let's  update the comment here

libs/go/sia/host/provider/provider.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/AthenZ/athenz/libs/go/sia/host/ip"
+	"github.com/AthenZ/athenz/libs/go/sia/host/signature"
+)
+
+// Provider is the interface which wraps various Providers known to ZTS
+// It has methods for providing attestationdata depending on provider type
+// and generating sub-parts of DN to be including in the CSR and San DNS and URI entries
+type Provider interface {
+	// PrepareKey creates/setup up a private key for use with the provider
+	PrepareKey(string) (*rsa.PrivateKey, error)
```
havetisyan, line 35: isn't there an interface that we can use here instead of rsa? We should be able to support both rsa and ec keys.

libs/go/sia/host/provider/provider.go
```diff
@@ -0,0 +1,66 @@
...
+
+	// AttestationData returns the attestation data that can be used in the ZTS api
+	AttestationData(string, *rsa.PrivateKey, *signature.SignatureInfo) (string, error)
```
havetisyan, line 38: same comment about private keys here as well.

libs/go/sia/host/provider/provider.go
psasidhar: switching to crypto.Privatekey

libs/go/sia/host/provider/provider.go
psasidhar: switched to crypto.Privatekey

havetisyan (approved):

PR #1777, opened 2022-02-08
Add domain dependency functions to zms java and go clients

OferLevi85:

### DIFF
--- a/libs/go/zmscli/cli.go
+++ b/libs/go/zmscli/cli.go
@@ -384,6 +384,10 @@ func (cli *Zms) EvalCommand(params []string) (*string, error) {
 				dn = args[0]
 			}
 			return cli.GetStats(dn)
+		case "get-dependent-domain-list":
+			if argc == 1 {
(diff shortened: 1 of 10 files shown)


libs/go/zmscli/cli.go
```diff
@@ -1077,6 +1077,22 @@ func (cli *Zms) EvalCommand(params []string) (*string, error) {
...
+			}
+		case "delete-domain-dependency":
+			if argc == 1 {
+				return cli.DeleteDomainDependency(dn, args[0])
+			}
+		case "get-dependent-service-list":
+			if argc == 1 {
+				return cli.GetDependentServiceList(dn)
+			}
+		case "get-dependent-domain-list":
```
havetisyan, line 1092: because this command does not require a domain, it's in a  wrong switch statement. it should be handled in the previous switch block where commands without the domain arguments are handled.

havetisyan (changes requested):

havetisyan (approved):

PR #1789, opened 2022-02-20
Domain Dependency for Tenancy

OferLevi85:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -6726,6 +6726,30 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         if (authzServiceTokenOperation) {
             setupTenantAdminPolicyInProvider(ctx, provSvcDomain, provSvcName, tenantDomain,
                     false, auditRef, caller);
+
+            // If listed as service provider, register dependency between the provider service and the tenant domain (unless dependency exist
+            // following creation of a previous resource group)
+
+            tenancyRegisterDomainDependency(ctx, tenantDomain, provider, auditRef, caller);
+        }
+    }
+
+    private void tenancyRegisterDomainDependency(ResourceContext ctx, String tenantDomain, String provider, String auditRef, String caller) {
+        if (serviceProviderManager.isServiceProvider(provider)) {
+            DomainList domainList = dbService.listDomainDependencies(provider);
+            if (!domainList.getNames().contains(tenantDomain)) {
+                dbService.putDomainDependency(ctx, tenantDomain, provider, auditRef, caller);
+            }
+        }
+    }
+
+    private void tenancyDeregisterDomainDependency(ResourceContext ctx, String tenantDomain, String provSvcDomain, String provSvcName, String auditRef, String caller) {
+        String serviceToDeregister = provSvcDomain + "." + provSvcName;
+        if (serviceProviderManager.isServiceProvider(serviceToDeregister)) {
+            boolean tenantDomainRolesExist = isTenantDomainRolesExist(tenantDomain, provSvcDomain, provSvcName);
+            if (!tenantDomainRolesExist) {
+                dbService.deleteDomainDependency(ctx, tenantDomain, serviceToDeregister, auditRef, caller);
+            }
         }
     }
 
@@ -6776,6 +6800,10 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         if (authzServiceTokenOperation) {
             dbService.executeDeleteTenantRoles(ctx, provSvcDomain, provSvcName, tenantDomain, null,
                 auditRef, caller);
+
+            // If listed as service provider, check if there are remaining resource group roles. If not - de-register dependency between the provider service and the tenant domain
+
+            tenancyDeregisterDomainDependency(ctx, tenantDomain, provSvcDomain, provSvcName, auditRef, caller);
         }
 
         // now clean-up local domain roles and policies for this tenant
@@ -6820,7 +6848,8 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
 
         // validate our detail object against uri components
 
-        if (!validateTenancyObject(detail, tenantDomain, providerDomain + "." + providerService)) {
+        final String provider = providerDomain + "." + providerService;
+        if (!validateTenancyObject(detail, tenantDomain, provider)) {
             throw ZMSUtils.requestError("Invalid tenancy object", caller);
         }
 
@@ -6834,6 +6863,11 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
 
         setupTenantAdminPolicyInProvider(ctx, providerDomain, providerService, tenantDomain,
                 false, auditRef, caller);
+
+        // If listed as service provider, register dependency between the provider service and the tenant domain (unless dependency exist
+        // following creation of a previous resource group)
+
+        tenancyRegisterDomainDependency(ctx, tenantDomain, provider, auditRef, caller);
     }
 
     @Override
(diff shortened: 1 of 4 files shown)


servers/zms/src/test/java/com/yahoo/athenz/zms/ZMSImplTest.java
```diff
@@ -6564,7 +6569,134 @@ public void testGetTenantResourceGroupRoles() {
...
+        assertEquals(ZMSTestInitializer.TABLE_PROVIDER_ROLE_ACTIONS.size(), tRoles.getRoles().size());
+
+        // Now remove the first resource group and verify dependency remains until the second resource group and admin role is removed
+
+        zmsImpl.deleteTenantResourceGroupRoles(serviceProviderCtx, domain, serviceName, tenantDomain, resourceGroup, zmsTestInitializer.getAuditRef());
+        dependentDomainList = zmsImpl.getDependentDomainList(zmsTestInitializer.getMockDomRsrcCtx(), domain + "." + serviceName);
+        assertEquals(dependentDomainList.getNames().size(), 1);
+        assertEquals(dependentDomainList.getNames().get(0), tenantDomain.toLowerCase());
+
+        // Now delete the second resouce group and verify dependency still remains until the admin role is removed
```
OferLevi85, line 6679: @havetisyan Please note that putTenantResourceGroupRoles and deleteTenantResourceGroupRoles are not symmetrical in their creation and deletion of roles.
An admin role is created in putTenantResourceGroupRoles the first time that a resource group is created for a domain. It isn't removed when deleteTenantResourceGroupRoles is called.
To actually remove the role, we use the "deleteTenant" call. 
For domain dependency - I keep the dependency until all roles are removed, including the admin. Let me know if you think the behavior should be different.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6726,6 +6726,49 @@ public void putTenancy(ResourceContext ctx, String tenantDomain, String provider
...
+            // following creation of a previous resource group)
+
+            tenancyRegisterDomainDependency(ctx, tenantDomain, provider, true, auditRef, caller);
+        }
+    }
+
+    private void tenancyRegisterDomainDependency(ResourceContext ctx, String tenantDomain, String provider, boolean authzServiceTokenOperation, String auditRef, String caller) {
+        String serviceToRegister = provider;
+        try {
+            serviceToRegister = getAuthorizedProviderService((RsrcCtxWrapper) ctx, serviceToRegister, authzServiceTokenOperation, caller);
```
havetisyan, line 6740: I'm not sure if I understand why are we making this call here? It looks to be in all cases we're dealing with the provider value that is passed in explicitly in the url or the body of the object. So why are we trying to possible change it - what is the use case where the provider value is not correct and we need to change it?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6726,6 +6726,49 @@ public void putTenancy(ResourceContext ctx, String tenantDomain, String provider
...
+                return;
+            }
+            throw ex;
+        }
+    }
+
+    private void tenancyDeregisterDomainDependency(ResourceContext ctx, String tenantDomain, String provSvcDomain, String provSvcName, boolean authzServiceTokenOperation, String auditRef, String caller) {
+        String serviceToDeregister = provSvcDomain + "." + provSvcName;
+        try {
+            serviceToDeregister = getAuthorizedProviderService((RsrcCtxWrapper) ctx, serviceToDeregister, authzServiceTokenOperation, caller);
```
havetisyan, line 6760: same comment as above. what is the use case where the serviceToDeregister is not the correct value?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6834,6 +6881,11 @@ public void putTenant(ResourceContext ctx, String providerDomain, String provide
 
         setupTenantAdminPolicyInProvider(ctx, providerDomain, providerService, tenantDomain,
                 false, auditRef, caller);
+
+        // If listed as service provider, register dependency between the provider service and the tenant domain (unless dependency exist
+        // following creation of a previous resource group)
+        String provider = providerDomain + "." + providerService;
```
havetisyan, line 6887: in the same function further up from there we have

`if (!validateTenancyObject(detail, tenantDomain, providerDomain + "." + providerService)) {`

So rather than doing the same providerDomain + "." + providerService; twice, define the value before the validateTenancyObject line and then use it here as well.

We should also make it provider as final String.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6991,6 +7047,13 @@ public TenantResourceGroupRoles putTenantResourceGroupRoles(ResourceContext ctx,
 
         dbService.executePutTenantRoles(ctx, provSvcDomain, provSvcName, tenantDomain,
                 resourceGroup, detail.getRoles(), false, auditRef, caller);
+
+        // If listed as service provider, register dependency between the provider service and the tenant domain (unless dependency exist
+        // following creation of a previous resource group)
+
+        String provider = provSvcDomain + "." + provSvcName;
```
havetisyan, line 7054: we can mark this as final

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
OferLevi85: For example, the putTenant call can be made by a service provider without an authorized service token. In this case, wouldn't we want to use the principal as the dependent service provider instead of the service specified in the body? or would we like to use the domain and service the principal passed? (as long as it is authorized to do the "putTenant" call for that specific domain and service)

OferLevi85: Well the putTenant call can be made by a principal which is different than the service provider it passed in the request. But I see now that even if that is the case, we would still want to use the provider it passed in the request and not itself as the dependent service (as long as the principal is allowed to call putTenant with a specific provider service, we can list that provider as the dependent)

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6726,6 +6726,46 @@ public void putTenancy(ResourceContext ctx, String tenantDomain, String provider
...
+    private void tenancyRegisterDomainDependency(ResourceContext ctx, String tenantDomain, String provider, String auditRef, String caller) {
+        try {
+            if (serviceProviderManager.isServiceProvider(provider)) {
+
+                DomainList domainList = dbService.listDomainDependencies(provider);
+                if (!domainList.getNames().contains(tenantDomain)) {
+                    dbService.putDomainDependency(ctx, tenantDomain, provider, auditRef, caller);
+                }
+            }
+        } catch (ResourceException ex) {
```
havetisyan, line 6746: do we still need this try/catch block? I don't see which api is returning the unauthorized exception.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -6726,6 +6726,46 @@ public void putTenancy(ResourceContext ctx, String tenantDomain, String provider
...
+    private void tenancyDeregisterDomainDependency(ResourceContext ctx, String tenantDomain, String provSvcDomain, String provSvcName, String auditRef, String caller) {
+        String serviceToDeregister = provSvcDomain + "." + provSvcName;
+        try {
+            if (serviceProviderManager.isServiceProvider(serviceToDeregister)) {
+                boolean tenantDomainRolesExist = isTenantDomainRolesExist(tenantDomain, provSvcDomain, provSvcName);
+                if (!tenantDomainRolesExist) {
+                    dbService.deleteDomainDependency(ctx, tenantDomain, serviceToDeregister, auditRef, caller);
+                }
+            }
+        } catch (ResourceException ex) {
```
havetisyan, line 6764: do we still need this try/catch block? I don't see which api is returning the unauthorized exception.

havetisyan (changes requested):

havetisyan (approved):

PR #1790, opened 2022-02-22
Adding SecureBoot Provider

psasidhar:

Main changes:

1) Added helper methods in Crypto.java and X509CertUtils to extract Subject DN, Subject CN, Issuer DN. The methods are designed to work with javax.servlet.request.X509Certificate attribute.

2) Modified ZTSImpl to send IssuerDN, SubjectDN, RSAModulusHash ONLY for register requests. This is deliberate so that the refresh requests can continue to come in with Athenz Service Identity Certificates and won't have any impact. This is done through a helper method, rather than adding another argument to generateInstanceConfirmObject, which already has many parameters.

3) SecureBoot provider looks at request certificate attributes (for CN and IssuerDN) only for register requests. And for refresh requests, the provider will look at certHostname filled in by ZTS.

4) AttrValidator interface and factory definitions.


libs/java/client_common/src/main/java/com/yahoo/athenz/common/utils/X509CertUtils.java
```diff
@@ -133,6 +136,132 @@ public static String extractRequestInstanceId(X509Certificate cert) {
...
+     */
+    public static String extractSubjectDn(X509Certificate[] certs) {
+        if (certs == null || certs.length == 0) {
+            return "";
+        }
+
+        return extractSubjectDn(certs[0]);
+    }
+
+    public static String extractSubjectDn(X509Certificate cert) {
```
havetisyan, line 183: please provide a java doc documentation for a public static method

libs/java/client_common/src/main/java/com/yahoo/athenz/common/utils/X509CertUtils.java
```diff
@@ -133,6 +136,132 @@ public static String extractRequestInstanceId(X509Certificate cert) {
...
+     */
+    public static String extractIssuerDn(X509Certificate[] certs) {
+        if (certs == null || certs.length == 0) {
+            return "";
+        }
+
+        return extractIssuerDn(certs[0]);
+    }
+
+    public static String extractIssuerDn(X509Certificate cert) {
```
havetisyan, line 219: please provide a java doc documentation for a public static method

havetisyan (changes requested):

psasidhar:
added the javadoc comments for the new methods in Crypto.java as well

havetisyan (approved):

PR #1805, opened 2022-03-06
disable role member expiry notification

mendi160:


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSConsts.java
```diff
@@ -343,4 +343,6 @@
     public static final String ZMS_PROP_DOMAIN_CHANGE_PUBLISHER_FACTORY_CLASS = "athenz.zms.domain_change_publisher_factory_class";
     public static final String ZMS_PROP_DOMAIN_CHANGE_PUBLISHER_DEFAULT = "com.yahoo.athenz.common.messaging.impl.NoOpDomainChangePublisherFactory";
```
OferLevi85, line 345: Let's keep the same structure for the const values.
DISABLE_EXPIRATION_NOTIFICATIONS_TAG = "zms.DisableExpirationNotifications";
DISABLE_REMINDER_NOTIFICATIONS_TAG = "zms.DisableReminderNotifications"

mendi160: My thought was to give it a more general name because we use it in "zms.DisableExpirationNotifications" also yo disable group member expiry notifications. I'll change it

PR #1810, opened 2022-03-09
Add provider webhook for verifying removal

OferLevi85:
of dependency before domain deletion


servers/zms/src/main/java/com/yahoo/athenz/zms/provider/ServiceProviderClient.java
```diff
@@ -0,0 +1,162 @@
...
+        String body = jsonMapper.writeValueAsString(providerDependencyRequest);
+        StringEntity stringEntity = new StringEntity(body);
+        httpPost.setEntity(stringEntity);
+        HttpDriverResponse httpResponse = httpDriver.doPostHttpResponse(httpPost);
+        if (httpResponse.getStatusCode() >= 500 && httpResponse.getStatusCode() <= 599) {
+            // provider error - principal should retry operation
+            domainDependencyProviderResponse.setStatus(PROVIDER_RESPONSE_DENY);
+            domainDependencyProviderResponse.setMessage("Http Status: " + httpResponse.getStatusCode() + ", error: " + httpResponse.getMessage());
+        } else if (httpResponse.getStatusCode() >= 400 && httpResponse.getStatusCode() <= 499) {
+            // TODO: Log error and send notification for athenz team on request failure
```
OferLevi85, line 144: Will be implemented in an upcoming pr

servers/zms/src/main/java/com/yahoo/athenz/zms/provider/ServiceProviderManager.java
```diff
@@ -0,0 +1,184 @@
...
+                    })
+                    .filter(Objects::nonNull)
+                    .collect(Collectors.toMap(DomainDependencyProvider::getProvider, Function.identity()));
+
+            serviceProviders = serviceProvidersUpdatedMap;
+        }
+    }
+
+    private boolean isServiceProviderAuthorizedLaunch(ServiceIdentity provSvcIdentity) {
+        // TODO: do access check for "launch" action in "sys.auth:instance"
```
OferLevi85, line 118: Will be implemented in an upcoming pr

servers/zms/src/main/java/com/yahoo/athenz/zms/provider/ProviderHostnameVerifier.java
```diff
@@ -0,0 +1,61 @@
...
+public class ProviderHostnameVerifier implements HostnameVerifier {
+    private final ServiceProviderManager serviceProviderManager;
+
+    public ProviderHostnameVerifier(ServiceProviderManager serviceProviderManager) {
+
+        this.serviceProviderManager = serviceProviderManager;
+    }
+
+    @Override
+    public boolean verify(String hostname, SSLSession session) {
```
havetisyan, line 39: This method requires that the caller is running athenz issued CA certs which is not the case. It could run both with an athenz issued certs or with a cert issued by a public ca (athenz issued ca is only a requirement for the ZTS instance endpoint). So the method should just do a standard validation (maybe somehow call the standard jetty std hostname verifier) and if that fails then try to fetch the service name and compare that.

havetisyan, line 39: This method requires that the caller is running athenz issued CA certs which is not the case. It could run both with an athenz issued certs or with a cert issued by a public ca (athenz issued ca is only a requirement for the ZTS instance endpoint). So the method should just do a standard validation against the hostname extracted from the uri and if that fails then try to fetch the service name and compare that.

servers/zms/src/main/java/com/yahoo/athenz/zms/provider/ServiceProviderClient.java
```diff
@@ -0,0 +1,162 @@
...
+
+    private String getProviderEndpoint(ServiceProviderManager.DomainDependencyProvider domainDependencyProvider, String domain) {
+        String url = domainDependencyProvider.getProviderEndpoint();
+        if (!url.endsWith("/")) {
+            url += "/";
+        }
+        if (domainDependencyProvider.isInstanceProvider()) {
+            url += "dependency-check/";
+        }
+        url += domain;
```
havetisyan, line 159: We don't need to include the domainName in the URI. it's a POST request and the payload includes the required domain name.

havetisyan, line 159: We don't need to include the domainName in the URI. it's a POST request and the payload includes the required domain name.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -1543,24 +1555,60 @@ void deleteDomain(ResourceContext ctx, String auditRef, String domainName, Strin
...
 
-    private void verifyNoServiceDependenciesOnDomain(String domainName, String caller) {
-        ServiceIdentityList serviceIdentityList = dbService.listServiceDependencies(domainName);
-        if (serviceIdentityList.getNames() != null && !serviceIdentityList.getNames().isEmpty()) {
-            StringBuilder msgBuilder = new StringBuilder("Remove domain '");
-            msgBuilder.append(domainName);
-            msgBuilder.append("' dependency from the following service(s):");
-            msgBuilder.append(String.join(", ", serviceIdentityList.getNames()));
-            throw ZMSUtils.forbiddenError(msgBuilder.toString(), caller);
+    private void tryRemoveServiceDependenciesForDomain(ResourceContext ctx, String auditRef, String domainName, String caller, ServiceIdentityList serviceIdentityList) {
```
havetisyan, line 1570: I'm not sure if we're following the design here. There are two cases that need to be addressed here:
a) services that are using our dependency / tenancy api - these are using the dependency table as was implemented before and no changes are necessary. these providers will not register a provider endpoint
b) services that are not using any of the dependency / tenancy api - these will never be referenced in our dependency table but we want to do a real-time check before we allow the deletion to happen. In this case we never touch the dependency table as these services will never be referenced there.

havetisyan, line 1570: I'm not sure if we're following the design here. There are two cases that need to be addressed here:
a) services that are using our dependency / tenancy api - these are using the dependency table as was implemented before and no changes are necessary. these providers will not register a provider endpoint
b) services that are not using any of the dependency / tenancy api - these will never be referenced in our dependency table but we want to do a real-time check before we allow the deletion to happen. In this case we never touch the dependency table as these services will never be referenced there.

havetisyan (changes requested):

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/http/HttpDriver.java
```diff
@@ -296,19 +305,11 @@ public String doPost(HttpPost httpPost) throws IOException {
...
-                            //received bad statuscode, don't bother resending request.
-                            LOGGER.error("Received bad status code: {} from: {} reason: {}", statusCode, url, response.getStatusLine());
-                            //before breaking out. We must close inputstream to prevent from leaking.
-                            response.getEntity().getContent().close();
-                            return "";
-                    }
+                    String out = EntityUtils.toString(response.getEntity());
+                    LOGGER.debug("StatusCode: {} Data received: {}", statusCode, out);
+                    //before breaking out. We must close inputstream to prevent from leaking.
+                    response.getEntity().getContent().close();
```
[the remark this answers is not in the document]
OferLevi85: Yes, I agree. I haven't written the original implementation, just wrapped it so I can return HttpDriverResponse and didn't want to make an "out of scope" change that might break the HttpDriver in case I'm missing something.
I went over the implementation again and confirmed that it is closed by a "finally" statement. I'll remove the redundant close call.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
OferLevi85: I didn't understand it that way. In the design I wrote the following regarding deletion of domains:

"When a user attempts to delete a domain, athenz will query the service_domain_dependency table for any registered services associated with the domain.
If there are any registered services - athenz will contact the listed provider endpoint to check if the dependency can be removed.
If the provider denies deletion (or if there is no endpoint listed for the provider), athenz will block the delete attempt and return an error containing all the registered services (and relevant resource groups if they exist)."

I was under the impression that we only care about services listed in the dependency table. I'll change the implementation so before deleting a domain, we will go over all the services in the domain:
 - If a dependency exists in the dependency table, prevent deletion and return an error
 - If not, check if there is a provider endpoint listed
       - If there isn't, continue to the next service
       - if there is, contact the provider to get permissions to delete the domain.

Finally, if none of the services prevented deletion of the domain - delete it.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/http/HttpDriver.java
```diff
@@ -321,6 +322,28 @@ public String doPost(HttpPost httpPost) throws IOException {
...
 
+    /**
+     * doPost performs post operation and returns a string
+     * @param httpPost post request to process
+     * @return response string
+     * @throws IOException in case of any errors
+     */
+    public String doPost(HttpPost httpPost) throws IOException {
+        HttpDriverResponse httpDriverResponse = doPostHttpResponse(httpPost);
+        String url = httpPost.getURI().toString();
```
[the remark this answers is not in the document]
OferLevi85: Fixed

servers/zms/src/main/java/com/yahoo/athenz/zms/provider/ProviderHostnameVerifier.java
OferLevi85: If we accept certificates signed by non-athenz CA, the common name may not be equal to the service. So I'm not sure why we should check the service name.
For now, I'll remove the ProviderHostNameVerified so the default verifier in apache will be used (org.apache.http.conn.ssl.DefaultHostnameVerifier).
Please note that Jetty doesn't have its own implementation for HostnameVerifier.

servers/zms/src/main/java/com/yahoo/athenz/zms/provider/ServiceProviderClient.java
OferLevi85: EDIT: implemented. An error message was added which can be tracked by cloud watch.

servers/zms/src/main/java/com/yahoo/athenz/zms/provider/ServiceProviderManager.java
OferLevi85: Edit: implemented

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -1568,6 +1583,38 @@ private void verifyNoServiceDependenciesOnDomain(String domainName, String calle
         }
     }
 
+    private void verifyServiceProvidersApproveDeletion(ResourceContext ctx, String auditRef, AthenzDomain domain, String caller) {
+
+        List<ServiceIdentity> servicesWithProviderEndpoint = domain.getServices()
```
havetisyan, line 1588: Not sure why we're looking at the services in the domain? The list of provider services that have an endpoint defined got nothing to do with the domain itself. The services are already defined in the service_providers role name in the sys.auth domain and we only need to pick the ones that have a provider endpoint set. So I'm expecting that when we call the ServiceProviderManager::refreshServiceProviders method to fetch the service providers, we also read each service identity and if the provider is set then we keep that in a separate set that we use here for the check.

havetisyan (changes requested):

PR #1816, opened 2022-03-16
Add provider webhook for verifying removal

OferLevi85:
of dependency before domain deletion


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -1568,6 +1582,39 @@ private void verifyNoServiceDependenciesOnDomain(String domainName, String calle
         }
     }
 
+    private void verifyServiceProvidersAuthorizeDelete(ResourceContext ctx, String domainName, String caller) {
+        Principal principal = ((RsrcCtxWrapper) ctx).principal();
+        Map<String, ServiceProviderManager.DomainDependencyProvider> serviceProvidersWithEndpoints = serviceProviderManager.getServiceProvidersWithEndpoints();
+        List<String> serviceDependenciesDenied = serviceProvidersWithEndpoints.values()
+                .parallelStream()
```
OferLevi85, line 1589: I'm making concurrent calls to the provider services. No need to do them one at a time as the endpoints should mostly be different between services.

OferLevi85, line 1589: I'm making concurrent calls to the provider services. No need to do them one at a time as the endpoints should mostly be different between services.

havetisyan (approved):

PR #1817, opened 2022-03-17
Ability to validate CA issuers

pratikgote:
Signed-off-by: pgote <pgote@verizonmedia.com>


libs/java/auth_core/pom.xml
```diff
@@ -92,6 +92,12 @@
       <version>${mockito.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-util</artifactId>
+      <version>${jetty.version}</version>
+      <scope>compile</scope>
```
havetisyan, line 99: why only compile?

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/Authority.java
```diff
@@ -197,4 +200,8 @@ default String getUserEmail(final String username) {
     default List<Principal> getPrincipals(EnumSet<Principal.State> principalStates) {
         return Collections.emptyList();
     }
+
+    default String getCAPathProperty() {
```
havetisyan, line 204: I'm not sure why we have this method in the interface? I expect this type of functionality be within the authority implementation itself and not in the interface.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/CertificateAuthority.java
```diff
@@ -116,6 +120,11 @@ public Principal authenticate(X509Certificate[] certs, StringBuilder errMsg) {
 
         // create principal
         X509Certificate x509Cert = certId.getX509Certificate();
+
+        if (issuerDNs != null && !issuerDNs.isEmpty() && !issuerDNs.contains(Crypto.extractIssuerDn(x509Cert))) {
```
havetisyan, line 124: Rather than having this functionality here, have you looked at the CertificateIdentityParser class. I expect the extra validation of issuer DN to happen within this class and not anywhere else.

libs/java/client_common/src/main/java/com/yahoo/athenz/common/utils/X509CertUtils.java
```diff
@@ -223,16 +223,7 @@ public static String extractIssuerDn(X509Certificate[] certs) {
...
 
-        return extractIssuerDn(certs[0]);
-    }
-
-    /**
-     * extractIssuerDn returns the IssuerDN from the certificate passed in
-     * @param cert X509Certificate to extract the DN from
-     * @return string with Issuer DN
-     */
-    public static String extractIssuerDn(X509Certificate cert) {
```
havetisyan, line 234: we cannot move public methods from common client libraries since it will break any users of this method. So we can change the implementation to call the method fro Crypto class but not remove it.

havetisyan (changes requested):

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/CertificateAuthorityValidator.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.security.cert.X509Certificate;
+import java.util.Set;
+
+public class CertificateAuthorityValidator {
+
+    private Set<String> issuerDNs;
+
+    public CertificateAuthorityValidator() {
+        String trustStorePath = System.getProperty("athenz.authority.truststore.path");
```
havetisyan, line 30: let's define a private final static String for the property name.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/impl/CertificateAuthorityValidator.java
```diff
@@ -0,0 +1,61 @@
...
+        extractIssuerDNs(trustStorePath);
+    }
+
+    private void extractIssuerDNs(String trustStorePath) {
```
havetisyan, line 34: let's mark our trustStorePath in this and next function as final

havetisyan (approved):

PR #1863, opened 2022-05-08
Add athenz service access token support to SIA EKS sidecar

dvirguttman:


libs/go/sia/aws/agent/agent.go
```diff
@@ -567,6 +579,32 @@ func RunAgent(siaCmd, siaDir, ztsUrl string, opts *options.Options) {
 			stop <- true
 		}()
 
+		go func() {
```
havetisyan, line 582: how do we guarantee that when the goroutine is processed, there are valid identity certs already fetched otherwise our token fetch operation will not succeed?

dvirguttman: done, thx

libs/go/sia/aws/agent/agent.go
```diff
@@ -525,6 +543,7 @@ func RunAgent(siaCmd, siaDir, ztsUrl string, opts *options.Options) {
 						errors <- fmt.Errorf("refresh identity failed: %v\n", err)
 						return
 					}
+					identityRefreshed <- true
```
havetisyan, line 546: this seems to be in the wrong place and will cause a hang in the agent since we can't keep writing to the channel if no-one is reading it. e.g. if we have no token setup, after the initial setup, during the 2nd this this call will hang since the initial write hasn't been read anyone. 

we should make this after line 512 after we call RegisterInstance. Even then we should only do if we have valid token options configured (similar check like on line 590).

havetisyan (changes requested):

havetisyan:
@dvirguttman  now if I look at the flow of the code this is what we get when we have an initial service startup with no identity:

The "default:" block in RunAgent runs. right away it determines that there are no files in the cert directory so it goes ahead and calls RegisterInstance. 

Then we call the first go subroutine where we skip the refresh part and call GetRoleCertificate to get role certs. Then we go ahead and fetch the access tokens if options are set. At the same time, the token go subroutine is called. So what's going to happen here is that we're refreshing the tokens again because we have the identityCreated channel created and we pushed true to that channel. So basically we have an issue where we have 2 go subroutines running the same fetch token code. During normal server processing, there is no issue since we must upgrade right after certs are refreshed, but during the initial startup we should probably skip the call so the go subroutine that handles the refresh just does its job. We already have logic in the code with initialSetup block so we maybe we move the accessTokenRequest call within that block so it's called only during refresh case and not during the initial setup since the go subroutine will handle that case.

now with all these changes I don't think we need to the identityCreated channel anymore. I don't see what value it provides since we either register the service or refresh the certs and after each one the token call happens so let's remove that channel.

PR #1865, opened 2022-05-09
update to jetty 11, remove jersey from client impl

havetisyan:
all libraries are now compiled and published with JDK 11
update server to use jetty 11.x with jakarta namespace
new java client generator based on Apache http client and no dependency on jersey


libs/go/athenzconf/Makefile
```diff
@@ -10,18 +10,19 @@ GOPKGNAME = github.com/AthenZ/athenz/libs/go/athenzconf
 
 # check to see if go utility is installed
 GO := $(shell command -v go 2> /dev/null)
-export GOPATH=$(PWD)
+GOPATH := $(shell pwd)
```
OferLevi85, line 13: Noticed that in the clients Makefiles, the GOPATH was changed from local dir (PWD) to (/tmp/go):
export GOPATH ?= /tmp/go
Not sure why the directory was changed there but not here

clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -322,15 +317,42 @@ String lookupZMSUrl() {
...
+            return null;
+        }
+        Registry<ConnectionSocketFactory> registry = RegistryBuilder.<ConnectionSocketFactory>create()
+                .register("https", new SSLConnectionSocketFactory(sslContext))
+                .register("http", new PlainConnectionSocketFactory())
+                .build();
+        PoolingHttpClientConnectionManager poolingHttpClientConnectionManager
+                = new PoolingHttpClientConnectionManager(registry, dnsResolver);
+
+        //route is host + port.  Since we have only one, set the max and the route the same
```
OferLevi85, line 337: Not sure about the comment. It says "set the max and the route the same" but then default max per route is 30 while max total is 20

clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -190,19 +193,11 @@ public void close() {
...
-     * If already set, the existing value of the property will be updated.
-     * Setting a null value into a property effectively removes the property
-     * from the property bag.
-     *
-     * @param name  property name.
-     * @param value property value. null value removes the property with the given name.
+     * Set the DNSResolver to be used by the ZMS Client
+     * @param resolver user supplied dns resolver
      */
-    public void setProperty(String name, Object value) {
```
OferLevi85, line 202: This public method was removed from ZMSClient which was probably used by existing clients of the library. Once they will update to the new version their code will not compile. 
If this function can be safely ignored we can consider marking it as Deprecated and have it do nothing. Otherwise we should consider updating the README file with the API changes in the new version.

clients/java/zts/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
```diff
@@ -524,18 +522,11 @@ public void setEnablePrefetch(boolean state) {
...
-     * internal javax.ws.rs.client.Client client's property method.
-     * If already set, the existing value of the property will be updated.
-     * Setting a null value into a property effectively removes the property
-     * from the property bag.
-     * @param name property name.
-     * @param value property value. null value removes the property with the given name.
+     * Set the DNSResolver to be used by the ZTS Client
+     * @param resolver user supplied dns resolver
      */
-    public void setProperty(String name, Object value) {
```
OferLevi85, line 535: This public method was removed from ZTSClient which was probably used by existing clients of the library. Once they will update to the new version their code will not compile.
If this function can be safely ignored we can consider marking it as Deprecated and have it do nothing. Otherwise we should consider updating the README file with the API changes in the new version.

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
```diff
@@ -351,8 +327,8 @@ void loadServicePrivateKey() {
         PrivateKeyStoreFactory pkeyFactory;
         try {
             pkeyFactory = (PrivateKeyStoreFactory) Class.forName(pkeyFactoryClass).newInstance();
-        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
-            LOG.error("Invalid PrivateKeyStoreFactory class: {} error: {}", pkeyFactoryClass, e.getMessage());
+        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException ex) {
+            LOG.error("Invalid PrivateKeyStoreFactory class: {}", pkeyFactoryClass, ex);
```
OferLevi85, line 331: Log only has one {} (the ex will be ignored)

OferLevi85 (changes requested):
Please see minor comments.

containers/jetty/src/main/java/com/yahoo/athenz/container/AthenzJettyContainer.java
havetisyan: it won't be. if the last argument is an exception, then it includes the full backtrace automatically. I replaced the message with the full trace since sometimes we just get a null message and would have preferred to see the full backtrace in the log to see what was the failure.

clients/java/zts/src/main/java/com/yahoo/athenz/zts/ZTSClient.java
havetisyan: there are multiple changes in the client libraries since we're moving from jax-rs to apache directly and we will have a separate section in the changelog when 1.0 is released.

libs/go/sia/Makefile
```diff
@@ -0,0 +1,66 @@
+#
+# Makefile to build ZMS Go cli library
+# Prerequisite: Go development environment
+#
+# Copyright 2016 Yahoo Inc.
```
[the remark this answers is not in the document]
havetisyan: fixed

OferLevi85 (approved):

PR #1875, opened 2022-05-17
Get Auth History implementation

OferLevi85:


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -956,13 +959,27 @@ void loadObjectStore() {
...
     }
 
+    void loadAuthHistoryStore() {
+        String authHistoryFactoryClass = System.getProperty(ZMSConsts.ZMS_PROP_AUTH_HISTORY_STORE_FACTORY_CLASS,
+                ZMSConsts.ZMS_AUTH_HISTORY_STORE_FACTORY_CLASS);
+        AuthHistoryStoreFactory authHistoryStoreFactory;
+        try {
+            authHistoryStoreFactory = (AuthHistoryStoreFactory) Class.forName(authHistoryFactoryClass).newInstance();
+        } catch (InstantiationException | IllegalAccessException | ClassNotFoundException e) {
+            LOG.error("Invalid AuthHistoryStoreFactory class: {} error: {}", authHistoryFactoryClass, e.getMessage());
```
havetisyan, line 976: let's start just including the exception as the last argument so the full backtrace is logged instead of just the message. when the server fails sometimes these error messages are not that useful and having the full stack is better.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2638,6 +2655,48 @@ public DomainMetaStoreValidValuesList getDomainMetaStoreValidValuesList(Resource
...
+                    return authHistory;
+                })
+                .collect(Collectors.toList());
+        return new AuthHistoryList().setAuthHistoryList(authHistoryList);
+    }
+
+    private Timestamp getAuthHistoryTimeStamp(AuthHistoryRecord authHistoryRecord, String caller) {
+        try {
+            SimpleDateFormat formatter = new SimpleDateFormat("dd/MMM/yyyy:HH:mm:ss");
+            Date date = formatter.parse(authHistoryRecord.getTimestamp());
```
havetisyan, line 2691: what format are we storing the timestamps in the store? can't we just assume the use of unix epoch time so we don't have to deal with formatting and parsing of these values?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2638,6 +2655,48 @@ public DomainMetaStoreValidValuesList getDomainMetaStoreValidValuesList(Resource
...
+            Timestamp timestamp = Timestamp.fromDate(date);
+            return timestamp;
+        } catch (ParseException e) {
+            LOG.error("Error parsing timestamp for authHistoryRecord, timestamp will be empty: " + authHistoryRecord);
```
havetisyan, line 2695: Use {} instead of +

servers/zms/src/main/java/com/yahoo/athenz/zms/store/AuthHistoryRecord.java
```diff
@@ -0,0 +1,111 @@
...
+
+package com.yahoo.athenz.zms.store;
+
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey;
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbSortKey;
+
+import java.util.Objects;
+
+@DynamoDbBean
```
havetisyan, line 27: This should be a generic class regardless of where the data is stored. why are we using dynamodb annotations here?

servers/zms/src/main/java/com/yahoo/athenz/zms/store/AuthHistoryStoreFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+package com.yahoo.athenz.zms.store;
+
+public interface AuthHistoryStoreFactory {
+
+    /**
+     * Create and return a new AuthHistoryStore instance
+     * @return AuthHistoryStore instance
+     */
+    AuthHistoryStore create();
```
havetisyan, line 24: we should consider what the interface would require for other implementations. Maybe at a minimum pass the private key store so the plugin can read username/password details if necessary.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -956,13 +959,27 @@ void loadObjectStore() {
...
-        dbService = new DBService(objectStore, auditLogger, zmsConfig, auditReferenceValidator);
+        dbService = new DBService(objectStore, auditLogger, zmsConfig, auditReferenceValidator, authHistoryStore);
 
         // create our group fetcher based on the db service
 
         groupMemberFetcher = new ZMSGroupMembersFetcher(dbService);
     }
 
+    void loadAuthHistoryStore() {
+        String authHistoryFactoryClass = System.getProperty(ZMSConsts.ZMS_PROP_AUTH_HISTORY_STORE_FACTORY_CLASS,
```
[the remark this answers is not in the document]
OferLevi85: I thought it was required functionality. I'll make the necessary changes.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
OferLevi85: During sync from Access logs to DynamoDB, I'm extracting the time directly from the log and push it as is to DynamoDB (the format is dd/MMM/yyyy:HH:mm:ss).
The reason is that I expect the syncer to be in heavier load during the sync than Athenz when a user fetches these records for a specific domain. So I prefer the parsing to take place once we pull the records instead of doing it when the records are pushed to the DynamoDB.
Let me know if you think I should change the date to UNIX epoch time during push.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/AuthHistoryRecord.java
OferLevi85: The annotations are required for the "mapping table" functionality of the dynamoDB client:
        this.mappedTable = client.table(tableName, TableSchema.fromBean(AuthHistoryRecord.class));
It allows writing queries in a much more convenient way.
I agree that the class should be local to the dynamoDB implementation and the conversion from this structure to AuthHistory.java should take place as part of DynamoDBAuthHistoryStoreConnection and not in the ZMSImpl level.

havetisyan (approved):

PR #1876, opened 2022-05-17
Athens JWK config

dvirguttman:


clients/go/zts/model.go
```diff
@@ -2351,92 +2351,64 @@ func (self *SSHCertificates) Validate() error {
 }
 
 //
```
dvirguttman, line 2353: @havetisyan not sure why the order has changed. maybe different rdl/go version?

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -834,30 +844,114 @@ ServiceIdentity generateZTSServiceIdentity(com.yahoo.athenz.zms.ServiceIdentity
...
     }
-    
+
+    protected AthenzJWKConfig getAthenzJWKConfig(ResourceContext ctx) {
+        validateAthenzJWKConfigIsUpdated(ctx);
+        return athenzJWKConfig;
+    }
+
+    protected void validateAthenzJWKConfigIsUpdated(ResourceContext ctx) {
+        boolean rfc = false; // todo: configuration? something else?
```
dvirguttman, line 877: should we take it from system properties? what should be the default value?

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -834,30 +844,114 @@ ServiceIdentity generateZTSServiceIdentity(com.yahoo.athenz.zms.ServiceIdentity
...
+                        athenzJWKConfig = newJWKConf;
+                        lastAthenzJWKConfUpdateTime = now;
+                    }
+                }
+            }
+        }
+    }
+
+    protected boolean athenzJWKConfChanged(AthenzJWKConfig newJWKConf) {
+        //TODO: do we have a better way to know these kind of changes?
```
havetisyan, line 911: the timestamp we send back to the client should be the latest update time of the zms and zts services. so we take a look at the last modified timestamp for those services and which one is older is set as the value instead of the current value. Similarly we should use the same logic to find out if any value was changed or not - just take a look at the last modified timestamp of the zms/zts services.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
dvirguttman: @havetisyan what about that?

havetisyan: we should always use it as true. when we write the client part, at that we'll make sure we're handling the keys corre

dvirguttman: 👍🏻

core/zts/src/test/java/com/yahoo/athenz/zts/AthenzJWKConfigTest.java
```diff
@@ -0,0 +1,44 @@
+package com.yahoo.athenz.zts;
```
havetisyan, line 1: Need copyright header section

havetisyan (approved):

PR #1880, opened 2022-05-22
Fix for role member expiry notifications

OferLevi85:
When soon to be expired role member is a group, only send notification
to the domain admin

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationCommon.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationCommon.java
@@ -20,6 +20,7 @@ import com.yahoo.athenz.auth.AuthorityConsts;
 import com.yahoo.athenz.auth.util.AthenzUtils;
 import com.yahoo.athenz.common.ServerCommonConsts;
 import com.yahoo.athenz.common.server.util.ResourceUtils;
+import com.yahoo.athenz.zms.ResourceException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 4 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationCommon.java
```diff
@@ -97,19 +98,27 @@ public Notification createNotification(final String recipient,
     }
 
     void addDomainRoleRecipients(Notification notification, final String domainName, final String roleName) {
-        Set<String> domainRoleMembers = domainRoleMembersFetcher.getDomainRoleMembers(domainName, roleName);
-        if (domainRoleMembers == null || domainRoleMembers.isEmpty()) {
+        try {
```
OferLevi85, line 101: This try catch is not really necessary now that we handle groups as role members. However, it's best to catch any exception thrown for a specific recipient so the notification will still be sent to other recipients.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/NotificationCommon.java
```diff
@@ -97,19 +98,27 @@ public Notification createNotification(final String recipient,
...
     void addNotificationRecipient(Notification notification, final String recipient, boolean ignoreService) {
 
-        int idx = recipient.indexOf(AuthorityConsts.ROLE_SEP);
-        if (idx != -1) {
-            addDomainRoleRecipients(notification, recipient.substring(0, idx), recipient);
+        int roleDomainIndex = recipient.indexOf(AuthorityConsts.ROLE_SEP);
+        int groupDomainIndex = recipient.indexOf(AuthorityConsts.GROUP_SEP);
+        if (roleDomainIndex != -1) {
+            addDomainRoleRecipients(notification, recipient.substring(0, roleDomainIndex), recipient);
+        } else if (groupDomainIndex != -1) {
```
OferLevi85, line 120: Domain admins will get the notification but I don't think it's reasonable to send every member of a group notification if it expires on a specific role.

havetisyan (approved):

PR #1896, opened 2022-06-16
when athenz jwk conf file does not exist we should return zero time instead current time

dvirguttman:


dvirguttman:
@havetisyan we need this fix due to:
https://github.com/AthenZ/athenz/blob/master/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java#L3966

havetisyan (approved):

PR #1901, opened 2022-06-26
Set the "Referrer-Policy: strict-origin-when-cross-origin" header

OferLevi85:


havetisyan:
We don't need this functionality since we already have a generic functionality providing this same feature. If we look at a couple of lines above, the server has a config setting: athenz.response_headers_json where you can specify list of headers and static values and we automatically add those as HeaderPatternRule rules

PR #1905, opened 2022-06-27
Java zpe

dvirguttman:


clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
```diff
@@ -282,7 +289,21 @@ public static void setZPEClientClass(final String className) {
...
+            //  fetch all zts jwk keys and update config and try again
+            accessSignKeyResolver.loadPublicKeysFromServer();
+            lastZtsJwkFetchTime = System.currentTimeMillis();
+            publicKey = accessSignKeyResolver.getPublicKey(keyId); 
+        }
+        return publicKey;
+    }
+    
+    protected static boolean canFetchLatestJwksFromZts() {
+        int millisBetweenZtsCalls = Integer.parseInt(System.getProperty(ZPE_PROP_MILLIS_BETWEEN_ZTS_CALLS, Integer.toString(30 * 1000 * 60)));
```
havetisyan, line 303: every time we come across an invalid key we're going to get a system property parse it and then use the value. kind of wasting resources. we should get the system property in the static section for the class and use it all the time.

clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/pkey/file/FilePublicKeyStore.java
```diff
@@ -89,18 +131,26 @@ void loadPublicKeys(ArrayList<PublicKeyEntry> publicKeys, Map<String, PublicKey>
...
+    private PublicKey getPublicKey(String keyId, Map<String, PublicKey> ztsPublicKeyMap) {
         if (keyId == null) {
             return null;
         }
-        return zmsPublicKeyMap.get(keyId);
+        PublicKey publicKey = ztsPublicKeyMap.get(keyId);
+        if (publicKey == null) {
+            // reload athenz jwks from disk and try again
+            LOG.debug("key id: {} does not exist in public keys map, reload athenz jwks from disk", keyId);
+            initAthenzJWKConfig();
```
havetisyan, line 150: let's also have similar logic not read the file every time - e.g. once every 30 mins

havetisyan (changes requested):

havetisyan (approved):

PR #1915, opened 2022-07-11
Add athenz-zms-aws-domain-syncer

OferLevi85:


syncers/zms_aws_domain_syncer/pom.xml
```diff
@@ -0,0 +1,188 @@
...
+  </parent>
+
+  <artifactId>athenz-zms-aws-domain-syncer</artifactId>
+  <name>athenz-zms-aws-domain-syncer</name>
+  <description>Athenz ZMS AWS Domain Syncer</description>
+  <packaging>jar</packaging>
+
+  <properties>
+    <code.coverage.min>0.9300</code.coverage.min>
+    <jackson.core.version>2.13.2</jackson.core.version>
```
havetisyan, line 19: we don't need 19 and 20 since those are defined in the parent pom.xml

syncers/zms_aws_domain_syncer/pom.xml
```diff
@@ -0,0 +1,188 @@
...
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
+        <version>3.8.1</version>
```
havetisyan, line 112: we should use maven-compiler-plugin.version

syncers/zms_aws_domain_syncer/pom.xml
```diff
@@ -0,0 +1,188 @@
...
+            <configuration>
+              <includeScope>runtime</includeScope>
+              <overWriteReleases>false</overWriteReleases>
+              <overWriteSnapshots>false</overWriteSnapshots>
+              <overWriteIfNewer>true</overWriteIfNewer>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+      <plugin>
```
havetisyan, line 157: let's get rid of shade plugin as we should not use it.

syncers/zms_aws_domain_syncer/src/main/java/com/yahoo/athenz/zms_aws_domain_syncer/AwsSyncer.java
```diff
@@ -0,0 +1,98 @@
+package com.yahoo.athenz.zms_aws_domain_syncer;
```
havetisyan, line 1: need copyright message

syncers/zms_aws_domain_syncer/src/main/java/com/yahoo/athenz/zms_aws_domain_syncer/Config.java
```diff
@@ -0,0 +1,290 @@
+package com.yahoo.athenz.zms_aws_domain_syncer;
```
havetisyan, line 1: need copyright message

havetisyan (changes requested):
I made only comments on a couple of files but we need the license/copyright message on all files.

PR #1922, opened 2022-07-18
return obj with put role method

royelbaum:


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3707,7 +3707,8 @@ public void putRole(ResourceContext ctx, String domainName, String roleName, Str
 
         // process our request
 
-        dbService.executePutRole(ctx, domainName, roleName, role, auditRef, caller);
+        Role dbRole = dbService.executePutRole(ctx, domainName, roleName, role, auditRef, caller);
+        return returnObj != null && returnObj ? dbRole : null;
```
dvirguttman, line 3711: consider that instead of adding this line and pass the `returnObj` parameter in each PUT function, it can be part of the rdl code generation in the ZMSResources class

royelbaum: the rdl doest know how to generate this code right now, it means that we will need to add it to the rdl project

PR #1935, opened 2022-08-04
finish rdl changes in ZMSResources and ZMSClient

royelbaum:
Signed-off-by: Roy Elbaum roy.elbaum@yahooinc.com


clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -1064,20 +1064,36 @@ public Role getRole(String domainName, String roleName, boolean auditLog, boolea
      * @param domainName name of the domain
      * @param roleName   name of the role
      * @param auditRef   string containing audit specification or ticket number
+     * @param returnObj Boolean returns the updated object from the database if true
      * @param role       role object to be added to the domain
      * @throws ZMSClientException in case of failure
      */
-    public void putRole(String domainName, String roleName, String auditRef, Role role) {
+    public void putRole(String domainName, String roleName, String auditRef, Boolean returnObj, Role role) {
```
havetisyan, line 1071: So if the return value is still void how would we get the object back in Java? Shouldn't this return the object back to the client which would be either null if the retrunObj is false or the actual updated object?

same comment applies to all new api calls

clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -1064,20 +1064,36 @@ public Role getRole(String domainName, String roleName, boolean auditLog, boolea
...
         updatePrincipal();
         try {
-            client.putRole(domainName, roleName, auditRef, role);
+            client.putRole(domainName, roleName, auditRef, returnObj, role);
```
havetisyan, line 1074: this should return the object back to the client

rdl/rdl-gen-athenz-java-client/javaclient.go
```diff
@@ -332,6 +337,8 @@ func (gen *javaClientGenerator) clientMethodBody(r *rdl.Resource) string {
 	expected = append(expected, rdl.StatusCode(r.Expected))
 	couldBeNoContent := "NO_CONTENT" == r.Expected
 	couldBeNotModified := "NOT_MODIFIED" == r.Expected
+	couldBeReturned := "OK" == r.Expected
+	fmt.Sprintf("hey %b",couldBeReturned)
```
havetisyan, line 341: should be removed

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1340,8 +1347,12 @@ void executePutRole(ResourceContext ctx, String domainName, String roleName, Rol
 
                 // add domain change event
                 addDomainChangeMessage(ctx, domainName, roleName, DomainChangeMessage.ObjectType.ROLE);
-                
-                return;
+
+                if (returnObj == Boolean.TRUE){
```
havetisyan, line 1351: need a space after )

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1340,8 +1347,12 @@ void executePutRole(ResourceContext ctx, String domainName, String roleName, Rol
...
+                    return getRole(con, domainName, roleName, true,false, true);
```
havetisyan, line 1352: need a space after "true,"

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1138,9 +1138,12 @@ public void executePutPolicyVersion(ResourceContext ctx, String domainName, Stri
 
                 // add domain change event
                 addDomainChangeMessage(ctx, domainName, policyName, DomainChangeMessage.ObjectType.POLICY);
-                
-                return;
 
+                if (returnObj == Boolean.TRUE) {
```
havetisyan, line 1142: consider combining these next 5 lines into a single step:

return (returnObj == Boolean.TRUE) ? getPolicy(con, domainName, policyName, version) : null;

(same comment applies to all other sections in the code).

havetisyan (changes requested):
a couple of additional comments:

when rdl is updated, we must also generate the go client code and since we can't create wrappers like java in the go code,  we need to make sure update all the go utilities that use the zms client accordingly. just make sure you run "mvn clean install" from the root directory which should build all the components.

havetisyan:
merged as part of https://github.com/AthenZ/athenz/pull/1942

PR #1939, opened 2022-08-10
For sia access tokens - check expiry in claims

OferLevi85:

### DIFF
--- a/libs/go/sia/access/tokens/tokens.go
+++ b/libs/go/sia/access/tokens/tokens.go
@@ -20,6 +20,7 @@ import (
 	"crypto/tls"
 	"encoding/json"
 	"fmt"
+	"gopkg.in/square/go-jose.v2/jwt"
 	"io/ioutil"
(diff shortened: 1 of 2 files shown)


havetisyan:
@OferLevi85 I don't follow this change? looks like after I get the initial token of 6 hours, I won't get to refresh it until 5:45 mins later? If yes, then it's not correct. We want to make sure to refresh the token after 3 hours.

OferLevi85:
> @OferLevi85 I don't follow this change? looks like after I get the initial token of 6 hours, I won't get to refresh it until 5:45 mins later? If yes, then it's not correct. We want to make sure to refresh the token after 3 hours.

Yeah, I mentioned this change in the slack channel when I posted the PR as I wanted to get your / Sasi's input. I'll change the calculation so it will be refreshed after half the validity time but instead of basing the start time on modification time, I'll base it on the "iat" claim

havetisyan (approved):

PR #1941, opened 2022-08-15
auto purge expired role/group members based on server/domain config

mendi160:


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7994,6 +8033,117 @@ public DomainList listDomainDependencies(String service) {
...
+            // process our delete role member operation
+
+            if (!con.deleteExpiredRoleMember(roleId, principalId, normalizedMember, principal, auditRef)) {
+                throw ZMSUtils.notFoundError(caller + ": unable to delete role member: " +
+                        normalizedMember + " from role: " + roleName, caller);
+            }
+
+            return;
+
+        } catch (ResourceException ex) {
```
havetisyan, line 8050: why are we catching and throwing the exception. We should just remove the try / cache block.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7994,6 +8033,117 @@ public DomainList listDomainDependencies(String service) {
         }
     }
 
+    void executeDeleteExpiredMemberships(ResourceContext ctx, ObjectStoreConnection con, int roleId, String roleName,
```
havetisyan, line 8036: We should return to executeDeleteExpiredMembership since we're processing only a single member and not members

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7994,6 +8033,117 @@ public DomainList listDomainDependencies(String service) {
...
+
+            return;
+
+        } catch (ResourceException ex) {
+            throw ex;
+        }
+    }
+
+    public void executeDeleteAllExpiredRoleMemberships(ResourceContext ctx, String auditRef, String caller) {
+        final int maxDbCallsPerRun = new DynamicConfigInteger(CONFIG_MANAGER, ZMSConsts.ZMS_PROP_PURGE_TASK_MAX_DB_CALLS_PER_RUN, ZMSConsts.PURGE_TASK_MAX_DB_CALLS_PER_RUN_DEF).get();
```
havetisyan, line 8056: we should not create dynamic config objects within the method. we can create them in the DBService constructor and then use them here with .get()s.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7994,6 +8033,117 @@ public DomainList listDomainDependencies(String service) {
...
+        final int limitPerCall = new DynamicConfigInteger(CONFIG_MANAGER, ZMSConsts.ZMS_PROP_PURGE_TASK_LIMIT_PER_CALL, ZMSConsts.PURGE_TASK_LIMIT_PER_CALL_DEF).get();
+        int offset = 0;
+        List<PurgeMember> allExpiredRoleMembers = new ArrayList<>();
+        List<PurgeMember> changedList = new ArrayList<>();
+        ObjectStoreConnection con;
+        try {
+            con = store.getConnection(false, false);
```
havetisyan, line 8063: all of our getConnections should be done win the try-with-resources block so they're automatically closed and returned to the pool

servers/zms/src/main/java/com/yahoo/athenz/zms/purge/PurgeMember.java
```diff
@@ -0,0 +1,57 @@
+package com.yahoo.athenz.zms.purge;
+
+public class PurgeMember {
+    private String domainName;
+    private int collectionId;
```
havetisyan, line 5: we should never expose the collectionId and principalId. The server does not know what the underlying data storage is thus it can't make assumptions about it. This is why the DB layer has its own internal cache

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -6837,6 +6878,71 @@ public List<String> listDomainDependencies(String service) {
...
+        List<PurgeMember> members = new ArrayList<>();
+        try (PreparedStatement ps = con.prepareStatement(GET_ALL_EXPIRED_ROLE_MEMBERS)) {
+            ps.setInt(1, DELAY_PURGE_EXPIRED_MEMBERS_DAYS_DEFAULT);
+            ps.setInt(2, limit);
+            ps.setInt(3, offset);
+            try (ResultSet rs = executeQuery(ps, caller)) {
+                while (rs.next()) {
+                    PurgeMember purgeMember = new PurgeMember()
+                            .setDomainName(rs.getString("domain_name"))
+                            .setCollectionId(rs.getInt("role_id"))
```
havetisyan, line 6895: see the comment above - this should not be exposed.

core/zms/src/main/rdl/Domain.rdli
```diff
@@ -319,4 +319,20 @@ resource AuthHistoryDependencies GET "/domain/{domainName}/history/auth" {
...
         ResourceError TOO_MANY_REQUESTS;
     }
+}
+
+type ExpiredMembers Struct {
+    Array<MemberName> expiredRoleMembers; //list of deleted expired role members names
+    Array<GroupMemberName> expiredGroupMembers; //list of deleted expired groups members names
+}
+
+// Delete expired principals
```
mendi160, line 329: There is a conflict with the deleteTopLevelDomain endpoint, so /domain/expired-members will be changed to /expired-members

core/zms/src/main/rdl/Domain.rdli
```diff
@@ -319,4 +319,20 @@ resource AuthHistoryDependencies GET "/domain/{domainName}/history/auth" {
...
+resource ExpiredMembers DELETE "/domain/expired-members?purgeResources={purgeResources}" {
```
mendi160, line 330: There is a conflict with the deleteTopLevelDomain endpoint, so /domain/expired-members will be changed to /expired-members

havetisyan: what is the use of the purgeResources? wasn't clear from the description. Is it a boolean flag or includes multiple values? if multiple values then update the description to include what those values are

core/zms/src/main/rdl/Domain.rdli
```diff
@@ -319,4 +319,20 @@ resource AuthHistoryDependencies GET "/domain/{domainName}/history/auth" {
...
+}
+
+type ExpiredMembers Struct {
+    Array<MemberName> expiredRoleMembers; //list of deleted expired role members names
+    Array<GroupMemberName> expiredGroupMembers; //list of deleted expired groups members names
+}
+
+// Delete expired principals
+resource ExpiredMembers DELETE "/expired-members?purgeResources={purgeResources}" {
+    Int32 purgeResources (optional); //defining which resources will be purged. by default all resources will be purged
```
mendi160, line 331: 0 - none of them will be purged
1 - only roles will be purged
2 - only groups will be purged
3/null - both of them will be purged

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
```diff
@@ -608,6 +608,36 @@ public AuthHistoryDependencies getAuthHistoryDependencies(
...
+    @DELETE
+    @Path("/expired-members")
+    @Produces(MediaType.APPLICATION_JSON)
+    @Operation(description = "Delete expired principals")
+    public void deleteExpiredMembers(
+        @Parameter(description = "defining which resources will be purged. by default all resources will be purged", required = false) @QueryParam("purgeResources") Integer purgeResources) {
+        int code = ResourceException.OK;
+        ResourceContext context = null;
+        try {
+            context = this.delegate.newResourceContext(this.servletContext, this.request, this.response, "deleteExpiredMembers");
```
mendi160, line 620: an authorization check should be added here

core/zms/src/main/rdl/Domain.rdli
```diff
@@ -319,4 +319,20 @@ resource AuthHistoryDependencies GET "/domain/{domainName}/history/auth" {
...
     }
+}
+
+type ExpiredMembers Struct {
+    Array<MemberName> expiredRoleMembers; //list of deleted expired role members names
+    Array<GroupMemberName> expiredGroupMembers; //list of deleted expired groups members names
+}
+
+// Delete expired principals
+resource ExpiredMembers DELETE "/expired-members?purgeResources={purgeResources}" {
```
havetisyan, line 330: the definition of ExpiredMembers looks incomplete. We're just returning a list of names but that's not sufficient. What roles where those users members of? We need to specify the role/domain names and maybe even the expiration date.

I see we have defined PurgeMember in the code. Why not define it in the RDL maybe with a more generic name and include the expiration field as well

core/zms/src/main/rdl/Domain.rdli
```diff
@@ -319,4 +319,20 @@ resource AuthHistoryDependencies GET "/domain/{domainName}/history/auth" {
...
+    Int32 purgeResources (optional); //defining which resources will be purged. by default all resources will be purged
+    expected NO_CONTENT;
```
havetisyan, line 332: I think here it would be useful to return OK. NO_CONTENT and use the support we added for UI to ask the list of expired members to be returned or not.

clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -3594,4 +3594,22 @@ public DomainList getDependentDomainList(String service) {
...
     }
 
+    /**
+     * Delete expired members from roles and groups.
+     *
+     * @param purgeResources indicates which resource will be purged
+
+     * @throws ZMSClientException in case of failure
+     */
+    public void deleteExpiredMembers(Integer purgeResources) {
```
havetisyan, line 3604: see the comment below for the RDL to optionally return the list of expired members (similar to the support we added for the redux work in UI).

core/zms/src/main/rdl/Domain.rdli
havetisyan, line 330: we're missing the authorization statement in the definition

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7996,6 +8039,147 @@ public DomainList listDomainDependencies(String service) {
...
+
+        // process our delete role member operation
+
+        if (!con.deleteRoleMember(domainName, roleName, normalizedMember, principal, auditRef)) {
+            throw ZMSUtils.notFoundError(caller + ": unable to delete role member: " +
+                    normalizedMember + " from role: " + roleName, caller);
+        }
+    }
+
+    public void executeDeleteAllExpiredRoleMemberships(ResourceContext ctx, String auditRef, String caller) {
```
havetisyan, line 8055: we should somewhat careful about having all this in a single transaction. if our set contains users from 100 different domains, are we effectively blocking changes to those 100 domains until we commit our changes? How about if we fetch the list of users that need to be expunged, separate them per domain and then process each domain at a time. This way only a single domain is blocked and not all of them

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7996,6 +8039,147 @@ public DomainList listDomainDependencies(String service) {
...
+        final String principal = getPrincipalName(ctx);
+
+        // process our delete group member operation
+        if (!con.deleteGroupMember(domainName, groupName, normalizedMember, principal, auditRef)) {
+            throw ZMSUtils.notFoundError("unable to delete group member: " +
+                    normalizedMember + " from group: " + groupName, ctx.getApiName());
+        }
+    }
+
+    public void executeDeleteAllExpiredGroupMemberships(ResourceContext ctx, String auditRef, String caller) {
```
havetisyan, line 8124: same comment as above.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2681,6 +2683,37 @@ public AuthHistoryDependencies getAuthHistoryDependencies(ResourceContext ctx, S
         return dbService.getAuthHistory(domainName);
     }
 
+    @Override
+    public void deleteExpiredMembers(ResourceContext ctx, Integer purgeResources) {
+        final String caller = ctx.getApiName();
+        logPrincipal(ctx);
+        validateRequest(ctx.request(), caller);
+
+        LOG.info("deleteExpiredMembers: purgeResources value: {}", purgeResources);
```
havetisyan, line 2692: this seems to be unnecessary since we already have the other info line below and if the value is invalid then the error response will log it as well.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2681,6 +2683,37 @@ public AuthHistoryDependencies getAuthHistoryDependencies(ResourceContext ctx, S
...
+
+        if (!List.of(0, 1, 2, 3).contains(purgeResources)) {
+            throw ZMSUtils.requestError("Invalid purgeResources value, error: should be a number in [0-3]", caller);
+        }
+
+        EnumSet<PurgeResourcesEnum>  purgeResourcesEnumSet = PurgeResourcesEnum.getPurgeResourcesState(purgeResources);
+
+        LOG.info("deleteExpiredMembers: purgeResources: {}", purgeResourcesEnumSet);
+
+        if (purgeResourcesEnumSet.contains(PurgeResourcesEnum.ROLES)) {
```
havetisyan, line 2707: a follow up from the comment in the RDL, let's collect and return the value if requested

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
mendi160: This could lead to an edge case where someone will update the expiration in the meantime.
Do we need to consider this scenario?

core/zms/src/main/rdl/Domain.rdli
mendi160: My thought was, a list of strings would look as follows: <domain_name>:<collection_name>:<member_name>
However, it might be better to return PurgeMember instead of String.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
havetisyan: yes, we should. if we generate our delete sql commands such that to include the expiry time in the where clause when then we'll never delete a member incorrectly where the expiry was changed after we read the record.

core/zms/src/main/rdl/Domain.rdli
```diff
@@ -319,4 +319,29 @@ resource AuthHistoryDependencies GET "/domain/{domainName}/history/auth" {
...
+type ExpiredMembers Struct {
+    Array<ExpiryMember> expiredRoleMembers; //list of deleted expired role members
+    Array<ExpiryMember> expiredGroupMembers; //list of deleted expired groups members
+}
+
+// Delete expired principals
+resource ExpiredMembers DELETE "/expired-members?purgeResources={purgeResources}" {
+    Bool returnObj (optional, default=false, header="Athenz-Return-Object"); //Return object param updated object back.
+    Int32 purgeResources (optional); //defining which resources will be purged. by default all resources will be purged
+    authorize ("purge", "sys.auth:role.purge_expired_members");
```
havetisyan, line 340: let's replace this with:
authorize("delete", "sys.auth:expired_members");

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSConsts.java
```diff
@@ -245,6 +245,7 @@ public final class ZMSConsts {
     public static final String DB_COLUMN_AS_DOMAIN_NAME     = "domain_name";
     public static final String DB_COLUMN_AS_ROLE_NAME       = "role_name";
     public static final String DB_COLUMN_AS_GROUP_NAME      = "group_name";
+    public static final String DB_COLUMN_AS_PRINCIPAL_NAME     = "principal_name";
```
havetisyan, line 248: let's remove the extra space to align the values

servers/zms/src/main/java/com/yahoo/athenz/zms/purge/PurgeResourcesEnum.java
```diff
@@ -0,0 +1,16 @@
+package com.yahoo.athenz.zms.purge;
```
havetisyan, line 1: we need to add copyright notice

core/zms/src/test/java/com/yahoo/athenz/zms/ExpiryMemberTest.java
```diff
@@ -0,0 +1,71 @@
+package com.yahoo.athenz.zms;
```
havetisyan, line 1: we need to add copyright notice

core/zms/src/test/java/com/yahoo/athenz/zms/ExpiredMembersTest.java
```diff
@@ -0,0 +1,63 @@
+package com.yahoo.athenz.zms;
```
havetisyan, line 1: we need to add copyright notice

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1797,6 +1823,11 @@ void executePutEntity(ResourceContext ctx, String domainName, String entityName,
         }
     }
 
+//    void executeDeleteMembership(ResourceContext ctx, String domainName, String roleName,
```
havetisyan, line 1826: let's remove this if it's commented out

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -151,6 +151,7 @@ public class JDBCConnection implements ObjectStoreConnection {
     private static final String SQL_INSERT_PENDING_ROLE_MEMBER = "INSERT INTO pending_role_member "
             + "(role_id, principal_id, expiration, review_reminder, audit_ref, req_principal) VALUES (?,?,?,?,?,?);";
     private static final String SQL_DELETE_ROLE_MEMBER = "DELETE FROM role_member WHERE role_id=? AND principal_id=?;";
+    private static final String SQL_DELETE_EXPIRED_ROLE_MEMBER = "DELETE FROM role_member WHERE role_id=? AND principal_id=? AND ? >= expiration;";
```
havetisyan, line 154: any reason why we just don't have `expiration=?`

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -6838,6 +6951,59 @@ public List<String> listDomainDependencies(String service) {
...
+            throw sqlError(ex, caller);
+        }
+        return members;
+    }
+
+    @Override
+    public List<ExpiryMember> getAllExpiredGroupMembers(int limit, int offset) {
+        final String caller = "getAllExpiredGroupMembers";
+
+//        List<ExpiryMember> members = new ArrayList<>();
```
havetisyan, line 6984: let's remove any commented code

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7996,6 +8027,171 @@ public DomainList listDomainDependencies(String service) {
...
+        LOG.info("delete all expired role members started");
+
+        // delete all expired role members. for blocking only one domain at a time, for each domain, its expired members will be deleted in a separate transaction.
+        for (Map.Entry<String, List<ExpiryMember>> entry: allExpiredRoleMembersMap.entrySet()) {
+            List<ExpiryMember> removedDomainList = executeDeleteDomainExpiredRoleMemberships(ctx, entry.getKey(), entry.getValue(), auditRef, caller);
+            removedList.addAll(removedDomainList);
+        }
+
+        if (numOfExpiredMembersRetrieved == removedList.size()) {
+        LOG.info("delete all expired role members done successfully: {} expired role members were deleted", removedList.size());
```
havetisyan, line 8106: missing tab for indentation

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7996,6 +8027,171 @@ public DomainList listDomainDependencies(String service) {
...
+
+        if (!con.deleteExpiredRoleMember(domainName, roleName, normalizedMember, principal, expiration, auditRef)) {
+            throw ZMSUtils.notFoundError(caller + ": unable to delete role member: " +
+                    normalizedMember + " from role: " + roleName + ". this happened either because the member was deleted already or his expiration was updated very recently", caller);
+        }
+    }
+
+    List<ExpiryMember> executeDeleteDomainExpiredRoleMemberships(ResourceContext ctx, String domainName, List<ExpiryMember> members, String auditRef, String caller) {
+        List<ExpiryMember> removedList = new ArrayList<>();
+        for (int retryCount = defaultRetryCount; ; retryCount--) {
```
havetisyan, line 8045: because this is not an end user operation, let's not complicate the code with the retry logic. we should just try once and if it fails, just log it and move on since during next run we'll try to clean up again

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -7996,6 +8027,171 @@ public DomainList listDomainDependencies(String service) {
...
+        // process our delete expired group member operation
+        if (!con.deleteExpiredGroupMember(domainName, groupName, normalizedMember, principal, expiration, auditRef)) {
+            throw ZMSUtils.notFoundError("unable to delete group member: " +
+                    normalizedMember + " from group: " + groupName + ". this happened either because the member was deleted already or his expiration was updated very recently", ctx.getApiName());
+        }
+    }
+
+    List<ExpiryMember> executeDeleteDomainExpiredGroupMemberships(ResourceContext ctx, String domainName, List<ExpiryMember> members, String auditRef, String caller) {
+        List<ExpiryMember> removedList = new ArrayList<>();
+        for (int retryCount = defaultRetryCount; ; retryCount--) {
```
havetisyan, line 8126: same comment as above, let's simplify and remove the retry logic

havetisyan (changes requested):
@mendi160 also please make sure you have go 1.19 installed which is what should be used for generating the code. also merge the master into your branch since there were some changes in the go client generation with the move to 1.19.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -476,6 +477,7 @@ public class JDBCConnection implements ObjectStoreConnection {
     private static final String SQL_INSERT_PENDING_GROUP_MEMBER = "INSERT INTO pending_principal_group_member "
             + "(group_id, principal_id, expiration, audit_ref, req_principal) VALUES (?,?,?,?,?);";
     private static final String SQL_DELETE_GROUP_MEMBER = "DELETE FROM principal_group_member WHERE group_id=? AND principal_id=?;";
+    private static final String SQL_DELETE_EXPIRED_GROUP_MEMBER = "DELETE FROM principal_group_member WHERE group_id=? AND principal_id=? AND ? = expiration;";
```
havetisyan, line 480: let's update this to have "expiration=?" to be consistent with other where conditions

havetisyan (changes requested):

PR #1952, opened 2022-09-05
In athenz-auth-core skip tests for arm architecture

OferLevi85:
(libpam.os not available)

### DIFF
--- a/libs/java/auth_core/pom.xml
+++ b/libs/java/auth_core/pom.xml
@@ -105,4 +105,28 @@
     </dependency>
   </dependencies>
 
+  <profiles>
+    <profile>
+      <id>arm-os-build</id>
+      <activation>
+        <os>
+          <arch>aarch64</arch>
+        </os>
+      </activation>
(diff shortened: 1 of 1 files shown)


libs/java/auth_core/pom.xml
```diff
@@ -105,4 +105,26 @@
...
+        </os>
+      </activation>
+      <build>
+        <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-surefire-plugin</artifactId>
+            <version>${maven-surefire-plugin.version}</version>
+            <configuration>
+              <skipTests>true</skipTests>
```
dvirguttman, line 123: Is it possible to skip only UserAuthorityTest?
I know that it is possible to achieve it in code..

OferLevi85: Fixed it.

dvirguttman (approved):

PR #1955, opened 2022-09-07
Fix some issues in the zms client return obj.

royelbaum:
Add some unit test for zms client return obj.

Signed-off-by: Roy Elbaum <roy.elbaum@verizonmedia.com>

### DIFF
--- a/clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
+++ b/clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
@@ -1090,8 +1090,8 @@ public class ZMSClient implements Closeable {
      * @param role       role object to be added to the domain
      * @throws ZMSClientException in case of failure
      */
-    public void putRole(String domainName, String roleName, String auditRef, Role role) {
-     putRole(domainName, roleName, auditRef, false, role);
+    public Role putRole(String domainName, String roleName, String auditRef, Role role) {
+        return  putRole(domainName, roleName, auditRef, false, role);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -1090,8 +1090,8 @@ public Role putRole(String domainName, String roleName, String auditRef, Boolean
      * @param role       role object to be added to the domain
      * @throws ZMSClientException in case of failure
      */
-    public void putRole(String domainName, String roleName, String auditRef, Role role) {
-     putRole(domainName, roleName, auditRef, false, role);
+    public Role putRole(String domainName, String roleName, String auditRef, Role role) {
```
havetisyan, line 1093: we cannot change the existing interface since that will break all of our existing customers. There is no need for this change. We already have a putRole that returns the updated Role object.

same applies to all other changes.

havetisyan (changes requested):

clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -1195,6 +1209,21 @@ public void putMembership(String domainName, String roleName, String memberName,
...
 
 
+    /**
+     * Add a new member in the specified role.
+     *
+     * @param domainName name of the domain
+     * @param roleName   name of the role
+     * @param memberName name of the member to be added
+     * @param auditRef   string containing audit specification or ticket number
+     * @throws ZMSClientException in case of failure
```
havetisyan, line 1219: missing @returnObj from javadoc

clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -1179,6 +1179,20 @@ public void putMembership(String domainName, String roleName, String memberName,
...
 
+
+    /**
+     * Add a new member in the specified role.
+     *
+     * @param domainName name of the domain
+     * @param roleName   name of the role
+     * @param memberName name of the member to be added
+     * @param auditRef   string containing audit specification or ticket number
+     * @throws ZMSClientException in case of failure
```
havetisyan, line 1190: missing @returnObj from javadoc

havetisyan (changes requested):

havetisyan:
@royelbaum also review the jacoco output to make sure all new lines are fully covered since the SD job failed with:

10:12:36   [WARNING] Rule violated for bundle athenz-zms-java-client: lines covered ratio is 0.9514, but expected minimum is 0.9545

havetisyan:
included in #1957

PR #1972, opened 2022-09-28
Delete static workload expects resource name and not instanceId

OferLevi85:


havetisyan (commented):
@OferLevi85 I don't see any changes for the java msd client. Wouldn't the changes in the rdl impact the java client as well?

abvaidya:
@OferLevi85 Please revert the rdl changes since we need the name param to delete static workload.

OferLevi85:
> @OferLevi85 Please revert the rdl changes since we need the name param to delete static workload.

I've reverted the URL structure to be the same but renamed "instanceId" to "name" and changed its type to "string". It is now a mandatory field as it was before.

abvaidya (approved):

PR #1991, opened 2022-10-16
Change the api file in the ui

royelbaum:
Signed-off-by: Roy Elbaum <roy.elbaum@verizonmedia.com>


ui/src/redux/selectors/domains.js
```diff
@@ -44,15 +44,17 @@ export const selectAllDomainsList = (state) => {
     return state.domains.allDomainsList ? state.domains.allDomainsList : [];
 };
 
-export const selectPendingMembersList = (state, domainName) => {
```
OferLevi85, line 47: What happens if view === 'domains' but domainName is null?

OferLevi85 (approved):

PR #2006, opened 2022-11-27
fix review tab

mendi160:

### DIFF
--- a/ui/src/components/review/ReviewTable.js
+++ b/ui/src/components/review/ReviewTable.js
@@ -104,7 +104,6 @@ class ReviewTable extends React.Component {
         let members = props.members && props.members.map((m) => m.memberName);
         this.state = {
             roleObj: props.roleDetails,
-            list: props.members || [],
             submittedReview: false,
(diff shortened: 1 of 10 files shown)


mendi160:
The purpose of this PR is to fix the bug reported here: https://jira.ouryahoo.com/browse/ATHENS-6774

ui/src/components/review/ReviewTable.js
```diff
@@ -165,6 +164,10 @@ class ReviewTable extends React.Component {
                     this.props.onUpdateSuccess(
                         `Successfully submitted the review for role ${this.props.role}`
                     );
+                    this.setState({
```
chandrasekhar1996, line 167: does react componentDidUpdate method work with redux? 
Instead of setting the state again here we can use componentDidUpdate and if props has changed then update the state

chandrasekhar1996:
Line 308 in GroupReviewTable.js also needs to be updated to only show members who have already been approved in Review Tab

chandrasekhar1996 (approved):

PR #2011, opened 2022-12-01
simplify way to change user's expiry/reminder dates

mendi160:


chandrasekhar1996 (commented):
show the edit icon only if the member has not expired.

mendi160:
> show the edit icon only if the member has not expired.

 expiry edit must be available at all times.

PR #2019, opened 2022-12-12
1. Ask for confirmation when users request to delete members

mendi160:
2. Switch to the Members tab once the user clicks on the submit button

### DIFF
--- a/ui/src/components/group/GroupReviewTable.js
+++ b/ui/src/components/group/GroupReviewTable.js
@@ -24,6 +24,8 @@ import { selectReviewGroupMembers } from '../../redux/selectors/group';
 import { reviewGroup } from '../../redux/thunks/groups';
 import { connect } from 'react-redux';
 import produce from 'immer';
+import CollectionUtils from 'lodash';
+import DeleteModal from '../modal/DeleteModal';
 
 const TitleDiv = styled.div`
     font-size: 16px;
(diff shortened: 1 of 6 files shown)


ui/src/components/group/GroupReviewTable.js
```diff
@@ -115,52 +124,66 @@ class GroupReviewTable extends React.Component {
                 });
                 return;
             }
+            // show prompt for user to ask for confirmation once the user asked to delete member/s
+            let showDeleteConfirmation =
+                CollectionUtils.intersection(
```
chandrasekhar1996, line 129: why is taking intersection required here? Wouldn't checking the size of set `deletedMembers` suffice for this condition?

ui/src/components/group/GroupReviewTable.js
```diff
@@ -81,14 +83,24 @@ class GroupReviewTable extends React.Component {
...
         this.state = {
             submittedReview: false,
+            showDeleteConfirmation: false,
             extendedMembers: new Set(members),
             deletedMembers: new Set(),
         };
     }
+
+    componentDidUpdate = (prevProps) => {
+        if (prevProps.members.length > this.props.members.length) {
```
chandrasekhar1996, line 99: I don't think relying on the length is a good indication whether the data has changed or not

ui/src/components/group/GroupReviewTable.js
```diff
@@ -116,51 +132,61 @@ class GroupReviewTable extends React.Component {
...
+                this.setState({
+                    submittedReview: true,
+                    errorMessage: null,
+                    justification: '',
                 });
-        }
+                this.props.onUpdateSuccess(
+                    `Successfully submitted the review for group ${this.props.groupName}`
+                );
+                this.loadGroup();
```
chandrasekhar1996, line 183: Is calling loadGroup necessary since we are routing to the members page through onUpdateSuccess

ui/src/components/group/GroupReviewTable.js
mendi160: In my opinion, if there are more members in the prevProps.members, is a good indication that at least one member was deleted, so now, the deleted members set contains a member that has already been deleted. Anyway, as you mentioned in the comment below, we can get rid of this code since we are routing to the members page through onUpdateSuccess.

chandrasekhar1996 (approved):

PR #2028, opened 2023-01-03
setup functional test with webdriverio

jimmytsang:
Signed-off-by: jtsang01 [jimmy.tsang@yahooinc.com](mailto:jimmy.tsang@yahooinc.com)


ui/src/__tests__/spec/wdio.conf.js
```diff
@@ -1,73 +1,145 @@
...
+  WORK_DIR,
+  TUNNEL_IDENTIFIER = appConfig.tunnelIdentifier,
+  PARENT_TUNNEL = appConfig.parentTunnel,
 } = process.env;
+let athenzDomain;
+let athenzService;
+let sdv4FunctionalJob = SD_JOB_NAME && SD_JOB_NAME === appConfig.devSdJobName;
+
+if (process.env.INSTANCE) { // executing functional test pointing at dev environment
+  athenzDomain = 'athenz.k8s';
```
chandrasekhar1996, line 19: we have this information as part of default-config.js, lets use the value from there

ui/src/__tests__/spec/wdio.conf.js
```diff
@@ -1,73 +1,145 @@
...
+  ];
+}
+let config = {
+    ...localOrRemote.saucelabs,
+    //
+    // ====================
+    // Runner Configuration
+    // ====================
+    // WebdriverIO supports running e2e tests as well as unit and component tests.
+    runner: 'local',
```
chandrasekhar1996, line 104: wouldn't this set runner as local for saucelabs test too?

chandrasekhar1996:
General note: you need to add a copyright header for all new files 
sample: https://github.com/AthenZ/athenz/blob/65cbeafabf857e6981d97838a505847c48ef18bd/ui/src/__tests__/pages/index.test.js#L1

ui/src/__tests__/spec/wdio.conf.js
```diff
@@ -84,158 +156,322 @@ const config = {
...
+                    fulfill();
+                  });
+                });
+              });
+            });
+          });
+        });
+        let count = 0;
+        let Mocha = require('mocha');
+        if (!Mocha._pesExt) {
```
chandrasekhar1996, line 318: what is _pesExt ?

ui/src/__tests__/spec/wdio.conf.js
jimmytsang: it would be, but this doesn't affect saucelabs functional test execution

ui/src/__tests__/spec/wdio.conf.js
jimmytsang: it's just a flag to check if the hook is applied onto mocha. I actually don't think it's needed since saucelabs records and saves all test failures which means this screenshot logic is not needed. I'll remove this

ui/src/__tests__/spec/wdio.conf.js
jimmytsang: hmm I can't seem to find the values in default-config.js, could you give me a line reference?

Initially I saw that there's `athenzDomainService: 'athenz.k8s.athenz-ui-development'` in `extended-config.js` however, i want to split up the athenz and service to be able to execute functional test locally as `athenz.k8s` pointing to dev environment and `athenz.dev` pointing at local environment

chandrasekhar1996: I was referring to the athenzDomainService and then extract value from it or add 2 more config options, just looking to reduce the number of lines we need to change with a service change, not a blocker

chandrasekhar1996 (approved):

chandrasekhar1996 (changes requested):
instance url

chandrasekhar1996 (approved):

PR #2049, opened 2023-01-25
adding support for user access management profile option in sia eks

abvaidya:

### DIFF
--- a/provider/aws/sia-eks/authn.go
+++ b/provider/aws/sia-eks/authn.go
@@ -31,7 +31,7 @@ func GetEKSPodId() string {
 	return podId
 }
 
-func GetEKSConfig(configFile, metaEndpoint string, useRegionalSTS bool, region string) (*options.Config, *options.ConfigAccount, error) {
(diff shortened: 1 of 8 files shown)


provider/aws/sia-eks/authn.go
```diff
@@ -31,7 +31,7 @@ func GetEKSPodId() string {
 	return podId
 }
 
-func GetEKSConfig(configFile, metaEndpoint string, useRegionalSTS bool, region string) (*options.Config, *options.ConfigAccount, error) {
+func GetEKSConfig(profileConfigFile, configFile, metaEndpoint string, useRegionalSTS bool, region string) (*options.Config, *options.ConfigAccount, *options.AccessProfileConfig, error) {
```
havetisyan, line 34: in EC2 Config we have configFile, profileConfigFile - so let's keep the order consistent

havetisyan (changes requested):

havetisyan (approved):

PR #2053, opened 2023-01-29
Add a requirement for admin approval to delete members in review-enab…

mendi160:
…led roles


servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -2462,7 +2474,10 @@ boolean insertPendingRoleMember(int roleId, int principalId, RoleMember roleMemb
 
         int affectedRows;
         if (roleMemberExists) {
-
+            // check first that the current request doesn't contradict the existing one
+            if (conflictRoleMemberPendingRequest(roleId, roleMember.getMemberName(), roleMember.getPendingState())) {
```
havetisyan, line 2478: this actually generates another mysql call which I believe is unnecessary. before calling this method we already take a hit and call roleMemberExists method which checks if the member already exists or not. we can just update that query to return the state as well and if the request is a pending one we can check if the match and return exception there.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1884,8 +1906,13 @@ void executeDeleteMembership(ResourceContext ctx, String domainName, String role
 
                 // audit log the request
 
-                auditLogRequest(ctx, domainName, auditRef, caller, ZMSConsts.HTTP_DELETE,
-                        roleName, "{\"member\": \"" + normalizedMember + "\"}");
+                if (pending) {
```
havetisyan, line 1909: let's simplify the call and just use `... caller, pending ? ZMSConsts.HTTP_PUT : MSConsts.HTTP_DELETE, roleName ....`

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1870,7 +1880,19 @@ void executeDeleteMembership(ResourceContext ctx, String domainName, String role
 
                 // process our delete role member operation
 
-                if (!con.deleteRoleMember(domainName, roleName, normalizedMember, principal, auditRef)) {
+                Role role = con.getRole(domainName, roleName);
+                boolean pending = (role != null) && (role.getReviewEnabled() == Boolean.TRUE || role.getAuditEnabled() == Boolean.TRUE);
```
havetisyan, line 1884: I've been thinking about this and we should introduce a new setting otherwise some may not like a change in behavior. There is also another feature request to protect deletion of roles themselves so we might be able to use the same setting for that as well.

Let's introduce a new meta attribute called deleteProtection. so for audit and review enabled roles, we'll ask for delete confirmation only if this flag is also turned on.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -1049,8 +1058,9 @@ boolean shouldRetryOperation(ResourceException ex, int retryCount) {
         switch (ex.getCode()) {
 
             case ResourceException.CONFLICT:
-
-                retry = true;
+                if (!ex.getMessage().contains("already has a pending request in a different state")) {
```
havetisyan, line 1061: we should not make our logic based on message content. it is possible that the customer might write their own datastore plugin and may not be aware that they must return something specific in the error message. instead of returning CONFLICT, we should just return BAD_REQUEST and the error message can indicate what we already have.

havetisyan (changes requested):

havetisyan (approved):

PR #2056, opened 2023-02-02
1. pedning request for delete member from protected role

mendi160:
2. display switch for delete protection role


ui/src/redux/reducers/roles.js
```diff
@@ -90,26 +90,6 @@ export const roles = (state = {}, action) => {
             });
             return newState;
         }
-        case ADD_MEMBER_TO_STORE: {
```
mendi160, line 93: removed due to the duplication of code

chandrasekhar1996 (approved):

PR #2063, opened 2023-02-05
AssertionCondition - scope

OferLevi85:
Added the TransportPolicyScope type which is part of TransportPolicyRule and represents the cloud providers scope for this rule (onPrem, AWS, GCP, etc)


havetisyan:
@OferLevi85 shouldn't there be updates to the msd go client since the cdl was changed?

core/msd/src/main/rdl/TransportPolicyRule.tdl
```diff
@@ -38,6 +38,7 @@ type TransportPolicySubject Struct {
 type TransportPolicyCondition Struct {
     TransportPolicyEnforcementState enforcementState; // State of transport policy enforcement ( ENFORCE / REPORT )
     Array<String> instances (optional); // Acts as restrictions. If present, this transport policy should be restricted to only mentioned instances.
+    Int64 scope (optional); // Scope of transport policy
```
abvaidya, line 41: Having an int here, without knowing what it represents, doesn't sound right to me. We should use ENUM since that makes it clear about what values the field can hold.

abvaidya (changes requested):

core/msd/src/main/rdl/TransportPolicyRule.tdl
abvaidya: Also I would recommend to use a prescriptive name for the enum something like TransportPolicyScope

OferLevi85: Enum in RDL won't work here as I need the value to be a mask. However, following additional technical difficulties that Rajesh mentioned, I've decided to go a different route: a simple boolean per cloud environment (A TransportPolicyScope structure with isOnPrem and isAws members).

dvirguttman (approved):

abvaidya (approved):

PR #2064, opened 2023-02-06
UI: Add scope to MSD policies

OferLevi85:


chandrasekhar1996 (approved):
👍 LGTM, thank you for the quick changes

One small note: If user selects, on-prem and AWS then we should automatically check "All" to prevent displaying "OnPrem AWS" in enforcement state view.

PR #2067, opened 2023-02-07
extends group with deleteProtection attribute and pending group membe…

mendi160:
…r with pendingState attribute


libs/go/zmscli/cli.go
```diff
@@ -1017,6 +1017,14 @@ func (cli *Zms) EvalCommand(params []string) (*string, error) {
 				}
 				return cli.SetGroupReviewEnabled(dn, args[0], reviewEnabled)
 			}
+		case "set-group-delete-protection":
+			if argc == 2 {
+				deleteProtection, err := strconv.ParseBool(args[1])
+				if err != nil {
+					return nil, err
+				}
+				return cli.SetGroupReviewEnabled(dn, args[0], deleteProtection)
```
havetisyan, line 1026: wrong function is being called. it should be SetGroupDeleteProtection

havetisyan (changes requested):

libs/go/zmscli/cli.go
mendi160: copy&paste is a terrible thing...
fixed

havetisyan (approved):

PR #2081, opened 2023-02-21
use user domain of config value when making requests to zms

ysknkd:
I would like to be able to use the config value when making requests to zms. Because the user domain is not `user` in our environment.


ui/src/components/constants/constants.js
```diff
@@ -22,7 +22,7 @@ export const GROUP_MEMBER_NAME_REGEX =
 export const MICROSEGMENTATION_SERVICE_NAME_REGEX =
     '\\*|([a-zA-Z0-9_][a-zA-Z0-9_-]*\\.)*[a-zA-Z0-9_][a-zA-Z0-9_-]*';
 export const GROUP_ROLES_CATEGORY = 'group-roles';
-export const GROUP_MEMBER_PLACEHOLDER = 'user.<userid> or <domain>.<service>';
+export const GROUP_MEMBER_PLACEHOLDER = `${process.env.NEXT_PUBLIC_USER_DOMAIN}.<userid> or <domain>.<service>`;
```
chandrasekhar1996, line 25: To avoid breaking the existing builds of UI, can you export a constant which would be set to `user` if `NEXT_PUBLIC_USER_DOMAIN` is undefined else would use the value of environment variable.
That constant can be used in other client files as well.

ui/src/components/member/Member.js
```diff
@@ -79,7 +79,7 @@ export default class Member extends React.Component {
         }
 
         let fullName = '';
-        let shortId = this.props.item.memberName.startsWith('user.')
+        let shortId = this.props.item.memberName.startsWith(`${process.env.NEXT_PUBLIC_USER_DOMAIN}`)
             ? this.props.item.memberName.substr(
                   5,
```
chandrasekhar1996, line 84: `5` will need to be updated to length of new constant + 1.

chandrasekhar1996 (commented):
Thank you for the PR @ysknkd

ui/src/components/member/Member.js
ysknkd: I'm trying to fix it like this, but It appears that `shortId` is not being used, so can I remove the logic regarding `shortId`?

```js
        const userDomain = `${process.env.NEXT_PUBLIC_USER_DOMAIN}.`;
        let shortId = this.props.item.memberName.startsWith(userDomain)
            ? this.props.item.memberName.substr(
                  userDomain.length,
                  this.props.item.memberName.length
              )
            : this.props.item.memberName;
```

chandrasekhar1996: yes, it can be removed.

On another note, this file also needs to be updated: https://github.com/AthenZ/athenz/blob/06d9820526480c4887c7e40c541ab2cd581f51dc/ui/src/components/domain/CreateDomain.js#L149

ui/src/components/constants/constants.js
ysknkd: Would it be better to define `USER_DOMAIN` in constants.js and reference it from the required components as follows?

constants.js

```js
export const USER_DOMAIN = process.env.NEXT_PUBLIC_USER_DOMAIN || 'user';
export const GROUP_MEMBER_PLACEHOLDER = `${USER_DOMAIN}. <userid> or <domain>. <service>`;
```

e.g.) Apply USER_DOMAIN to AddMember.js

```js
import { GROUP_MEMBER_NAME_REGEX, USER_DOMAIN } from '. /constants/constants';

... snip...

                            placeholder={
                                this.props.category === 'role' &&
                                this.props.collection ! == 'admin' !
                                    ? `${USER_DOMAIN}. <shortid> or <domain>. <service> or unix.<group> or <domain>:group.<group>`
                                    : `${USER_DOMAIN}. <shortid> or <domain>. <service> or unix.<group>`
```

chandrasekhar1996: Yes, this looks good

chandrasekhar1996: Something similar will also be required for server side code

ui/src/components/member/Member.js
ysknkd: Thank you, I fixed it.

ui/src/components/constants/constants.js
ysknkd: The server-side code uses the `userDomain` values defined in `src/config/config.js`.  Does this mean I should refer to `process.env.NEXT_PUBLIC_USER_DOMAIN` in config.js as well?

chandrasekhar1996: I am sorry I didn't notice that userDomain is already defined. This shouldn't be an issue.

chandrasekhar1996 (approved):
👍

Issue #2093, opened 2023-03-09
server not handling case sensitive tag names correctly


havetisyan:
(oath_tools) ➜ ✗ zms-cli -d home.craman add-domain-tag ums.restrictto test3
tags:
      - key: ums.restrictto
        values:
          - test3
      - key: ums.restrictTo
        values:
          - test1
          - test2

(oath_tools) ➜ ✗ zms-cli -d home.craman add-domain-tag ums.restrictto test4
tags:
      - key: ums.restrictto
        values:
          - test3
          - test4

fter running the last command, the key ums.restictTo got deleted.
I was expecting two tag keys to be present
a) ums.restrictTo and b) ums.restrictto with two tag values each

PR #2094, opened 2023-03-09
update zms server schema to set tag key column with utf8-bin collation

havetisyan:


abvaidya (approved):

havetisyan:
addresses issue #2093

PR #2095, opened 2023-03-14
add related changes for gcp support

jimmytsang:


ui/src/redux/thunks/user.js
```diff
@@ -27,3 +34,23 @@ export const getUserPendingMembers = () => async (dispatch, getState) => {
         dispatch(loadUserPendingMembers(userPendingMembersList, expiry));
     }
 };
+
+export const getUserResourceAccessList = () => async (dispatch, getState) => {
+    if (getState().user.userResourceAccessList) {
```
royelbaum, line 39: I am not sure if userResourceAccessList can be updated but if it is it should come after the check of expiry part, because if you put it here it will never be expired and refreshed

jimmytsang: good point, updated logic to first check for expiry then fetch data if needed.

ui/src/pages/gcp/login.js
```diff
@@ -0,0 +1,338 @@
...
+                projectNameAndRole.length > 1 ? projectNameAndRole[0] : '';
+            let projectRoleName =
+                projectNameAndRole.length > 1
+                    ? projectNameAndRole[1].replace('role.', '')
+                    : '';
+            let projectObject = {
+                projectId,
+                projectRoleName,
+                projectName,
+            };
```
jimmytsang, line 190: use util function

jimmytsang:
```
projects = ['athenz.project2:role.gcp.fed.admin.user']

projectRoleMap = {
    athenz.project2: [{
        projectId: 'athenz.project2:role.gcp.fed.admin.user',
        projectRoleName: 'gcp.fed.admin.user',
        projectName: 'athenz.project2'
}]
}
```

ui/src/server/handlers/api.js
```diff
@@ -3012,6 +3012,33 @@ Fetchr.registerService({
     },
 });
 
+Fetchr.registerService({
+    name: 'resource-access',
+    read(req, resource, params, config, callback) {
+        req.clients.zms.getResourceAccessList(
+            { action: params.action, principal: 'user.' + req.session.shortId },
```
chandrasekhar1996, line 3019: Instead of using `user.`, can you use appConfig.userDomain
Similar to https://github.com/AthenZ/athenz/blob/004da941e7356f791bbd9cefde3f636357440359/ui/src/server/handlers/api.js#L461

ui/src/pages/gcp/login.js
```diff
@@ -0,0 +1,338 @@
...
+                                Error: There are no GCP project roles associated
+                                with your account.
+                            </h3>
+                            <p>
+                                Check to make sure that your <tt>gcp.*</tt>{' '}
+                                roles contain your user.
+                                <br />
+                                E.g., if your account is 'jdoe', then{' '}
+                                <tt>gcp.fed.admin.user</tt> should have{' '}
+                                <tt>user.jdoe</tt> as a member.
```
chandrasekhar1996, line 283: Instead of `user.jdoe`, can you use `consts.USER_DOMAIN + ".jdoe"`

https://github.com/AthenZ/athenz/blob/f7791d91ed7709110d7644a5cd9ed28cd7409c64/ui/src/components/constants/constants.js#L25

ui/src/server/handlers/api.js
```diff
@@ -3012,6 +3012,33 @@ Fetchr.registerService({
...
+            (err, list) => {
+                if (err) {
+                    debug(
+                        `principal: ${req.session.shortId} rid: ${
+                            req.headers.rid
+                        } Error from ZMS while calling getRole API: ${JSON.stringify(
```
chandrasekhar1996, line 3025: `getRole` instead of `getResourceAccessList` in error message

ui/src/pages/gcp/login.js
```diff
@@ -0,0 +1,338 @@
...
+
+    componentDidUpdate(prevProps) {
+        const { resourceAccessList } = this.props;
+        if (prevProps && prevProps.resourceAccessList !== resourceAccessList) {
+            let projects = this.populateProjects();
+            this.populateProjectRoleMap(projects);
+        }
+    }
+
+    showError(errorMessage) {
```
chandrasekhar1996, line 143: Is this being used somewhere?

jimmytsang: good catch. Not anymore, will remove.

jimmytsang: fixed.

ui/src/server/handlers/api.js
jimmytsang: fixed.

ui/src/pages/gcp/login.js
jimmytsang: fixed.

ui/src/server/handlers/api.js
jimmytsang: fixed

ui/src/pages/gcp/login.js
jimmytsang: fixed

ui/src/pages/gcp/login.js
```diff
@@ -0,0 +1,329 @@
...
+        let response = RequestUtils.errorCheckHelper(err);
+        reload = response.reload;
+        error = response.error;
+        return [{}];
+    });
+
+    let queryParams = context.query || {};
+    let isAdmin = queryParams.isAdmin === 'true';
+    let projectDomainName = queryParams.projectDomainName || '';
+    let serverSideError = queryParams.error || '';
```
chandrasekhar1996, line 107: I don't think we should set if an error has occurred in query parameter. Once case to consider: 
If an occurred in server side (let's say an intermittent issue with zms), user will be redirected to URL with `error=true`, now the user will try to refresh the page and even though the issue may have been fixed on the server side, we will still show the error page since error param is set to true.

ui/src/redux/thunks/user.js
```diff
@@ -27,3 +34,19 @@ export const getUserPendingMembers = () => async (dispatch, getState) => {
...
 };
+
+export const getUserResourceAccessList = () => async (dispatch, getState) => {
+    let userResourceAccessList = selectUserResourceAccessList(getState());
+    let isUserResourceAccessListEmpty =
+        Array.isArray(userResourceAccessList) &&
+        userResourceAccessList.length < 1;
+    if (isExpired(getState().user.expiry)) {
+        userResourceAccessList = await API().getResourceAccessList({
+            action: 'gcp.assume_role',
```
chandrasekhar1996, line 45: let's not hard code this action here, instead add it as a parameter and let login.js send `gcp.assume_role`, that way other pages can use this if required later on

ui/src/pages/gcp/login.js
jimmytsang: updated validation error logic

ui/src/redux/thunks/user.js
jimmytsang: fixed

chandrasekhar1996: I believe there are few more changes required since currently you are just storing an array. Now if another page makes the same call with a different action, that array may get overridden. So I believe you need to maintain a mapping of "action" to "resourceAccessList". 
Can you confirm @royelbaum

jimmytsang: yeah i'll need to create a map between the action -> resourceAccessList. Selectors will need changes as well to include action param

royelbaum: completely true i talked about it with Jimmy

jimmytsang: Will address this in a follow up PR.

ui/src/server/handlers/api.js
chandrasekhar1996: @jimmytsang this one isn't fixed yet

jimmytsang: Sorry about that. Weird, I must've fixed it in another branch or something.

jimmytsang: fixed

chandrasekhar1996 (approved):

PR #2097, opened 2023-03-15
Add support for loading unencrypted pkcs8 keys

hiragi-gkuth:
Crypto.java has a function to read a private key from a file, but it cannot read a private key that is not encrypted in PKCS8 format.

Recent openssl cli generates unencrypted pkcs8 private keys by default, so we want to support this.


ysknkd:
@havetisyan Everything passed when I built and tested it on my local macOS. Could you please restart the Screwdriver job?.
image

havetisyan:
@hiragi-gkuth are you planning more changes since the PR is marked as draft?

hiragi-gkuth:
@havetisyan 

Thank you for your response. I set it to draft because CI did not pass, but I have re-run the tests on my local environment and did not encounter any issues, so I removed draft.

havetisyan (approved):

PR #2112, opened 2023-03-27
Increased assertion condition default quota to 14

OferLevi85:


abvaidya:
This can be set as a global config.

abvaidya (changes requested):
This can be set as a global config and shouldn't require a code change.

OferLevi85:
> This can be set as a global config and shouldn't require a code change.

@abvaidya yes but as of today this property wasn't overridden and without updating this value, the UI is broken. So I figured it is best to change the default. But sure I'll update the property instead if you think it is best.

PR #2119, opened 2023-03-30
Update DB Dockerfile to fix deploy-dev

4xpl0r3r:
When I tried to use `make deploy-dev` to deploy the servers in my development environment, I got this error in the `server.log`

image

The command I used to deploy is

```bash
make remove-all && make build && make deploy-dev
```

I believe it's caused by the misconfiguration of mysql db, so I made some modification to make it work normally.

I have tested these new `Dockerfile` and now it works well.


abvaidya (approved):
Thank you for the PR.

PR #2120, opened 2023-03-31
relax sandns check for aws instances to require configured suffix with domain/service values

havetisyan:


libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceUtils.java
```diff
@@ -51,6 +49,43 @@ public static String getInstanceProperty(final Map<String, String> attributes,
...
+        // extract the prefix component of the dns name
+
+        final String prefix = hostname.substring(0, suffixIdx);
+        for (String comp : prefix.split("\\.")) {
+            if (service.equals(comp)) {
+                return true;
+            }
+        }
+
+        LOGGER.error("{} does not end include required service name {} component", hostname, service);
```
MartinTrojans, line 85: Is this a typo? "end include"

MartinTrojans (approved):

PR #2123, opened 2023-04-04
gcp sia

abvaidya:
High level changes in this PR -
libs/go/sia/agent is copied from libs/go/sia/aws/agent and added few changes
libs/go/sia/options is copied from libs/go/sia/aws/options and added few changes
libs/go/sia/sds is copied from libs/go/sia/aws/sds for re-use
libs/go/sia/gcp is the new folder supporting gcp specific meta and attestation implementation

libs/go/sia/host/provider has the Provider interface, new methods are added to it. The idea is to use the Provider implementation for cloud / environment specific logic ( think STS, Meta, Attestation data etc ) 

No changes are done to the existing SIAs, except the Provider implementation with default implementation for new methods.

provider/gcp/{sia-gce,sia-gke} are the new SIAs using the new code from libs/go/sia/{agent,options,sds}


libs/go/sia/gcp/attestation/attestation.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/AthenZ/athenz/libs/go/sia/gcp/meta"
+)
+
+type GoogleAttestationData struct {
+	IdentityToken string `json:"identityToken,omitempty"` //the instance identity token obtained from the metadata server
+}
+
+// New creates a new AttestationData by getting instance identity token
+// from the Google metadata server
+func New(base, service, ztsServerName string) (string, error) {
```
havetisyan, line 30: we should update the ztsServerName parameter to either be ztsUrl to make it clear as far what we're passing or just calling audience and that would let each provider agent call determine what is the audience value

libs/go/sia/gcp/meta/meta.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	if idx := strings.Index(string(serviceBytes), "@"); idx > 0 {
+		service := string(serviceBytes[:idx])
+		return service, nil
+	}
+	return "", fmt.Errorf("unable to derive service name from metadata")
+}
+
+func GetProfile(metaEndpoint string) (string, error) {
+	profileBytes, err := GetData(metaEndpoint, "/computeMetadata/v1/instance/attributes/calypsoProfile")
```
havetisyan, line 116: we had decided not to use calypso in any of the open source calls. it is referred as accessProfile in most other places so let's use that value here as well.

provider/gcp/sia-gce/authn.go
```diff
@@ -0,0 +1,73 @@
...
+	config, _, err := options.InitFileConfig(configFile, metaEndpoint, false, region, "", provider)
+	if err != nil {
+		log.Printf("Unable to process configuration file '%s': %v\n", configFile, err)
+		log.Println("Trying to determine service details from the environment variables...")
+		config, _, err = options.InitEnvConfig(config, provider)
+		if err != nil {
+			log.Printf("Unable to process environment settings: %v\n", err)
+			// if we do not have settings in our environment, we're going
+			// to use fallback to retrieve values from the context ( metadata etc )
+			config, profileConfig, err = options.InitGenericProfileConfig(metaEndpoint, "-service", "@", provider)
```
havetisyan, line 37: why are we passing "-service" and "@" arguments here? we have logic using them, so shouldn't they just be ""s to avoid any confusion?

provider/gcp/sia-gce/authn.go
```diff
@@ -0,0 +1,73 @@
...
+		profileConfig, err := GetGCEAccessProfile(profileConfigFile, metaEndpoint, false, region, provider)
+		if err != nil {
+			log.Printf("Unable to determine user access management profile information: %v\n", err)
+		}
+		return config, profileConfig, nil
+	}
+	return config, profileConfig, nil
+}
+
+func GetGCEAccessProfile(configFile, metaEndpoint string, useRegionalSTS bool, region string, provider provider.Provider) (*options.AccessProfileConfig, error) {
```
havetisyan, line 54: since these are provider specific functions we can drop the useRegioinalSTS flag since I don't think it makes sense for GCP

provider/gcp/sia-gce/authn.go
```diff
@@ -0,0 +1,73 @@
...
+	accessProfileConfig, err := options.InitAccessProfileFileConfig(configFile)
+	if err != nil {
+		log.Printf("Unable to process user access management configuration file '%s': %v\n", configFile, err)
+		log.Println("Trying to determine user access management profile details from the environment variables...")
+		accessProfileConfig, err = options.InitAccessProfileEnvConfig()
+		if err != nil {
+			log.Printf("Unable to process environment settings: %v\n", err)
+			// if we do not have settings in our environment, we're going
+			// to use fallback to retrieving access profile from meta
+			log.Println("Trying to determine user access management profile name from security credentials...")
```
havetisyan, line 64: let's also fix these messages since we're not dealing with instance profile ARNs, etc.

provider/gcp/sia-gce/authn.go
```diff
@@ -0,0 +1,73 @@
...
+			log.Println("Trying to determine user access management profile details from instance profile arn...")
+			_, accessProfileConfig, err = options.InitGenericProfileConfig(metaEndpoint, "-service", "@", provider)
```
havetisyan, line 66: same comment here as well. let's replace "-service" and "@" with ""s

provider/gcp/sia-gke/authn.go
```diff
@@ -0,0 +1,95 @@
...
+	config, _, err := options.InitFileConfig(configFile, metaEndpoint, false, region, "", provider)
+	if err != nil {
+		log.Printf("Unable to process configuration file '%s': %v\n", configFile, err)
+		log.Println("Trying to determine service details from the environment variables...")
+		config, _, err = options.InitEnvConfig(config, provider)
+		if err != nil {
+			log.Printf("Unable to process environment settings: %v\n", err)
+			// if we do not have settings in our environment, we're going
+			// to use fallback to retrieve values from the context ( metadata etc )
+			config, _, err = options.InitGenericProfileConfig(metaEndpoint, "-service", "@", provider)
```
havetisyan, line 51: same as with gce - let's replace "-service" and "@" with ""s

provider/gcp/sia-gke/authn.go
```diff
@@ -0,0 +1,95 @@
...
+		profileConfig, err := GetGKEAccessProfile(profileConfigFile, metaEndpoint, false, region)
+		if err != nil {
+			log.Printf("Unable to determine user access management profile information: %v\n", err)
+		}
+		return config, profileConfig, nil
+	}
+	return config, nil, nil
+}
+
+func GetGKEAccessProfile(configFile, metaEndpoint string, useRegionalSTS bool, region string) (*options.AccessProfileConfig, error) {
```
havetisyan, line 69: let's remove useRegioinalSTS argument from this method

provider/gcp/sia-gke/authn.go
```diff
@@ -0,0 +1,95 @@
...
+			// if we do not have settings in our environment, we're going
+			// to use fallback to <domain>.<service>-service@access-profile naming structure
+			log.Println("Trying to determine user access management profile name from security credentials...")
+			_, accessProfileConfig, err = options.InitCredsConfig("-service", "@", useRegionalSTS, region)
+			// if the profile is empty try to determine access profile info from instance profile
+			if accessProfileConfig == nil || accessProfileConfig.Profile == "" {
+				if err != nil {
+					log.Printf("Unable to obtain user access management profile info from security credentials, err: %v\n", err)
+				}
+				log.Println("Trying to determine user access management profile details from instance profile arn...")
```
havetisyan, line 86: let's update the error messages in this method to reflect gcp resources

provider/gcp/sia-gke/authn.go
```diff
@@ -0,0 +1,95 @@
...
+				_, accessProfileConfig, err = options.InitProfileConfig(metaEndpoint, "-service", "@")
```
havetisyan, line 87: same as with gce - let's replace "-service" and "@" with ""s

havetisyan (changes requested):

havetisyan (approved):

PR #2131, opened 2023-04-16
for eks/gke generate san dns entries based on k8s dns spec

havetisyan:
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods

### DIFF
--- a/libs/go/sia/host/utils/utils.go
+++ b/libs/go/sia/host/utils/utils.go
@@ -17,11 +17,15 @@
 package utils
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"os/exec"
 	"strings"
 )
+
+const PodClusterLocalDNSSuffix = "pod.cluster.local"
+const ServiceClusterLocalDNSSuffix = "svc.cluster.local"
 
 // GetHostname returns the hostname
 func GetHostname(fqdn bool) string {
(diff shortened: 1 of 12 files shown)


libs/go/sia/host/utils/utils.go
```diff
@@ -44,3 +48,47 @@ func GetHostname(fqdn bool) string {
...
+	}
+	podNamespace := os.Getenv("ATHENZ_SIA_POD_NAMESPACE")
+	if podNamespace == "" {
+		// use namespace associated with the service account
+		if ns, err := os.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace"); err == nil {
+			podNamespace = strings.TrimSpace(string(ns))
+		}
+	}
+	podIP := os.Getenv("ATHENZ_SIA_POD_IP")
+	podService := os.Getenv("ATHENZ_SIA_POD_SERVICE")
```
abvaidya, line 76: Should we also set the $service.$namespace.svc.cluster.local sandns entry?

havetisyan: good idea. I only added what's in the pod spec but that's probably as useful as the pod based ones.

libs/go/sia/host/utils/utils.go
```diff
@@ -44,3 +48,55 @@ func GetHostname(fqdn bool) string {
...
+	}
+	// in all of our components we need our namespace so if we don't
+	// have one configured there is no need to fetch other settings
+	if podNamespace == "" {
+		return k8sDnsEntries
+	}
+
+	podName := os.Getenv("ATHENZ_SIA_POD_NAME")
+	if podName == "" {
+		podName = os.Getenv("HOSTNAME")
```
abvaidya, line 78: is utils.GetHostName() a better fit here?

libs/go/sia/host/utils/utils.go
```diff
@@ -44,3 +48,55 @@ func GetHostname(fqdn bool) string {
...
+	}
+	podIP := os.Getenv("ATHENZ_SIA_POD_IP")
+	podService := os.Getenv("ATHENZ_SIA_POD_SERVICE")
+	podSubdomain := os.Getenv("ATHENZ_SIA_POD_SUBDOMAIN")
+
+	if podIP != "" {
+		podIPWithDashes := strings.ReplaceAll(podIP, ".", "-")
```
abvaidya, line 85: do we have to worry about IPv6?

libs/go/sia/host/utils/utils.go
havetisyan: No, we don't want to use fqdn as the pod name which GetHostName will return

libs/go/sia/host/utils/utils.go
havetisyan: I haven't seen any references about using IPv6 in the pod dns name. So the expectation is that you'll set the ATHENZ_SIA_POD_IP only if you have ipv4 assigned to your pod otherwise, you won't get the ip based pod dns name in your sandns list.

abvaidya (approved):

PR #2143, opened 2023-04-26
Ensure snow error does not interfere with UI usability

jimmytsang:


ui/src/api.js
```diff
@@ -1643,8 +1643,9 @@ const Api = (req) => {
                     .read('meta')
                     .params(params)
                     .end((err, data) => {
+                        // this error can be ignored
                         if (err) {
-                            reject(err);
+                            resolve({});
```
jimmytsang, line 1648: Discussed with @havetisyan that errors from this API endpoint can be ignored from UI. As a result, users will be able to continue using athenz ui but be unable to add Business services to their domain if this api is in error state

chandrasekhar1996: I don't think we should ignore the error here, Redux will think that the API call was successful and won't try to fetch the data again until time expiry. 
If we handle it as part of Redux reducer (or more appropriate place) then when user will navigate between pages, we can make Redux to try to fetch the data with each navigation.

MartinTrojans (approved):

ui/src/api.js
jimmytsang: good point, fixed!

chandrasekhar1996 (approved):

PR #2149, opened 2023-04-28
gce fix for host cert principals and x509 cert expiry

abvaidya:


libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceGCPProvider.java
```diff
@@ -45,8 +45,7 @@ public class InstanceGCPProvider implements InstanceProvider {
     static final String GCP_PROP_BOOT_TIME_OFFSET = "athenz.zts.gcp_boot_time_offset";
     static final String GCP_PROP_DNS_SUFFIX       = "athenz.zts.gcp_dns_suffix";
     static final String GCP_PROP_REGION_NAME      = "athenz.zts.gcp_region_name";
-
-    static final String GCP_PROP_CERT_VALIDITY_NO_ADDITIONAL_INFO = "athenz.zts.gcp_cert_validity_no_additional_info";
+    static final String GCP_PROP_CERT_VALIDITY = "athenz.zts.gcp_cert_validity";
```
MartinTrojans, line 48: Should we keep the indent to make sure the "=" in the same row

MartinTrojans (approved):

Issue #2171, opened 2023-05-13
Want to overwrite roles with zms-cli


ysknkd:
# Want to overwrite roles with zms-cli

The `add-regular-role` command in zms-cli performs a role duplication check, which prevents from overwriting a role with specified members.
https://github.com/AthenZ/athenz/blob/cc277fca14df6093d41df9252e32eefda7e3b4e8/libs/go/zmscli/role.go#L122-L125

Here is the current behavior:

```shell
$ zms-cli -d test-domain add-regular-role test-role user.alice user.bob

role:
    - name: test-role
      members:
        - name: user.alice
        - name: user.bob

$ zms-cli -d test-domain add-regular-role test-role user.carol user.charlie
*** role already exists: test-domain:role.test-role
```

I would like to have a feature that overwrites an existing role when a certain option is specified.

```shell
$ zms-cli -d test-domain add-regular-role test-role user.alice user.bob

role:
    - name: test-role
      members:
        - name: user.alice
        - name: user.bob

$ zms-cli -d test-domain -force=true add-regular-role test-role user.carol user.charlie

role:
    - name: test-role
      members:
        - name: user.carol 
        - name: user.charlie
```

Just out of curiosity, is the absence of such a feature due to the current policy of zms-cli?

havetisyan:
If you're not careful, having such functionality could have un-intended results where admins may accidentally overwrite their roles with only a few members and thus cause some kind of outage. This is why we never added such a flag and nobody has requested it.

Additionally, you cannot overwrite roles that are either review-enabled or audit-enabled. If you have these flags set on a role, then you must use the the add/delete role member api commands.

But you're welcome to submit a PR to add such a flag - though I would use `-overwrite=true` instead of `-force=true` since overwrite clearly specifies the intention of the call. As I mentioned above, it will only be supported for roles without review/audit flags.

ysknkd:
Thank you for your response.

I understand your explanation regarding the potential for unintended results and the restrictions on overwriting roles that are review-enabled or audit-enabled.

With your points in mind, I will proceed to create a PR using the -overwrite=true flag, taking care to ensure it only applies to roles without review/audit flags.

PR #2179, opened 2023-05-19
add overwrite option for zms-cli

TakuyaMatsu:
https://github.com/AthenZ/athenz/issues/2171

### DIFF
--- a/utils/zms-cli/zms-cli.go
+++ b/utils/zms-cli/zms-cli.go
@@ -154,7 +154,7 @@ func usage() string {
 	buf.WriteString("   -k                  Disable peer verification of SSL certificates.\n")
 	buf.WriteString("   -key x509_key       Athenz X.509 Key file for authentication\n")
 	buf.WriteString("   -o output_format    Output format - json or yaml (default=yaml)\n")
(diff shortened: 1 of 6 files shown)


utils/zms-cli/zms-cli.go
```diff
@@ -154,6 +154,7 @@ func usage() string {
 	buf.WriteString("   -k                  Disable peer verification of SSL certificates.\n")
 	buf.WriteString("   -key x509_key       Athenz X.509 Key file for authentication\n")
 	buf.WriteString("   -o output_format    Output format - json or yaml (default=yaml)\n")
+	buf.WriteString("   -overwrite           Overwrites without checking for existence\n")
```
havetisyan, line 157: not to be super picky but there is an extra space before Overwrites so let's make sure our online help is all aligned

havetisyan (changes requested):

utils/zms-cli/zms-cli.go
TakuyaMatsu: thank you. ill check it

havetisyan (approved):

PR #2180, opened 2023-05-19
replace deprecated request pkg

jimmytsang:
replace the `request` package with `axios`


ui/package.json
```diff
@@ -97,7 +96,7 @@
         "redux-devtools-extension": "2.13.9",
         "redux-mock-store": "1.5.4",
         "redux-thunk": "2.4.1",
-        "request": "2.88.2",
+        "axios": "1.4.0",
```
jimmytsang, line 99: Considered using native nodejs functions but ran into connectivity errors that were not easily resolvable.
I believe this current solution using `axios` is appropriate for the time being as it involves very little code change and is a popular [npm package](https://www.npmjs.com/package/axios) that has a very active community supporting it.

ui/src/rdl-rest.js
```diff
@@ -17,6 +17,7 @@
 
 var url = require('url');
 var path = require('path');
+var axios = require('axios');
 var request = require('request');
```
chandrasekhar1996, line 21: need to remove this

chandrasekhar1996:
package-lock.json will also need to be updated

ui/src/rdl-rest.js
jimmytsang: removed

chandrasekhar1996 (approved):

PR #2185, opened 2023-05-24
fix role update on expiration and review dates

noy93845:
add filtering option to role members by updates to expiration or review dates

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/util/AuthzHelper.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/util/AuthzHelper.java
@@ -33,6 +33,7 @@ import org.slf4j.LoggerFactory;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Objects;
 
 public class AuthzHelper {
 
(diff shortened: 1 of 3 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/util/AuthzHelper.java
```diff
@@ -47,12 +48,19 @@ static ObjectMapper initJsonMapper() {
         return objectMapper;
     }
 
-    public static void removeRoleMembers(List<RoleMember> originalRoleMembers, List<RoleMember> removeRoleMembers) {
+    private static boolean isUpdateRequired(RoleMember member1, RoleMember member2) {
+        return !Objects.equals(member1.getExpiration(), member2.getExpiration()) ||
+                !Objects.equals(member1.getReviewReminder(), member2.getReviewReminder());
+    }
```
dvirguttman, line 54: Why not use the RoleMember::equals instead?
Don't we want to compare all RoleMember fields?

noy93845: I've also considered that, but the only fields that can be updated by the end user (like through the UI) are only these two fields. I don't want to check other things that can't be changed or are not allowed to be changed.

havetisyan:
@noy93845 most likely same issue exists for group members- processUpdateGroupMembers method in DBService. Can you please update the logic there as well.

noy93845:
> @noy93845 most likely same issue exists for group members- processUpdateGroupMembers method in DBService. Can you please update the logic there as well.

I've pushed a fix for groups as well

havetisyan (approved):

PR #2195, opened 2023-05-31
Add support in service tags

royelbaum:


servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -6700,8 +6729,112 @@ public boolean deleteRoleTags(String roleName, String domainName, Set<String> ta
...
+            throw sqlError(ex, caller);
+        }
+        return serviceTag;
+    }
+
+
+
     private void addTagsToGroups(Map<String, Group> groupMap, String domainName) {
 
+        //TODO change to the generic one
```
dvirguttman, line 6837: delete todos from code

clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSClient.java
```diff
@@ -1980,13 +1978,15 @@ public void deleteServiceIdentity(String domainName, String serviceName, String
      * @param domainName name of the domain
      * @param publicKeys include all public keys for services as well
      * @param hosts      include all configured hosts for services as well
+     * @param tagKey     query all groups with given tag name
+     * @param tagValue   query all groups with given tag key and value
```
dvirguttman, line 1982: it should be services

core/zms/src/main/java/com/yahoo/athenz/zms/ZMSSchema.java
```diff
@@ -2370,6 +2371,8 @@ private static Schema build() {
             .pathParam("domainName", "DomainName", "name of the domain")
             .queryParam("publickeys", "publickeys", "Bool", false, "return list of public keys in the service")
             .queryParam("hosts", "hosts", "Bool", false, "return list of hosts in the service")
+            .queryParam("tagKey", "tagKey", "CompoundName", null, "flag to query all groups that have a given tagName")
+            .queryParam("tagValue", "tagValue", "CompoundName", null, "flag to query all groups that have a given tag name and value")
```
dvirguttman, line 2375: same - it should be services .
probably root cause is the rdl definition

core/zms/src/main/rdl/ServiceIdentity.rdli
```diff
@@ -56,10 +56,12 @@ resource ServiceIdentity DELETE "/domain/{domain}/service/{service}" {
 }
 
 //Retrieve list of service identities
-resource ServiceIdentities GET "/domain/{domainName}/services?publickeys={publickeys}&hosts={hosts}" {
+resource ServiceIdentities GET "/domain/{domainName}/services?publickeys={publickeys}&hosts={hosts}&tagKey={tagKey}&tagValue={tagValue}" {
     DomainName domainName; //name of the domain
     Bool publickeys (optional, default=false); // return list of public keys in the service
     Bool hosts (optional, default=false); // return list of hosts in the service
+    CompoundName tagKey (optional); //flag to query all groups that have a given tagName
+    CompoundName tagValue (optional); //flag to query all groups that have a given tag name and value
```
dvirguttman, line 64: services..

libs/go/zmscli/service.go
```diff
@@ -69,6 +69,100 @@ func (cli Zms) ShowUpdatedService(service *zms.ServiceIdentity) (*string, error)
 	return cli.dumpByFormat(service, oldYamlConverter)
 }
 
+func (cli Zms) DeleteServiceTags(dn string, sn, tagKey string, tagValue string) (*string, error) {
+	service, err := cli.Zms.GetServiceIdentity(zms.DomainName(dn), zms.SimpleName(sn))
+	if err != nil {
+		return nil, err
+	}
+	//meta := getServiceMeta(service)
```
dvirguttman, line 77: delete unused code

libs/go/zmscli/cli.go
```diff
@@ -677,6 +677,24 @@ func (cli Zms) EvalCommand(params []string) (*string, error) {
...
+				return cli.DeleteServiceTags(dn, args[0], args[1], args[2])
+			}
+		case "show-services":
+			if argc == 0 {
+				return cli.ShowServices(dn, "", "")
+			} else if argc == 1 {
+				return cli.ShowServices(dn, args[0], "")
+			} else if argc == 2 {
+				return cli.ShowServices(dn, args[0], args[1])
+			}
```
dvirguttman, line 697: you must add helpers in `HelpSpecificCommand`

libs/go/zmscli/dump.go
```diff
@@ -447,10 +447,12 @@ func (cli Zms) dumpObjectList(buf *bytes.Buffer, list []string, dn string, objec
 	}
 }
 
-func (cli Zms) dumpServices(buf *bytes.Buffer, dn string) {
+func (cli Zms) dumpServices(buf *bytes.Buffer, dn string, tagKey string, tagValue string) {
+	buf.WriteString(indentLevel1)
+	buf.WriteString("services:\n")
```
dvirguttman, line 452: I am pretty sure we cannot add here texts, since there are teams that are parsing zms-cli output.
@havetisyan is it correct?

anyway, why did you add it?

libs/go/zmscli/service.go
```diff
@@ -69,6 +69,100 @@ func (cli Zms) ShowUpdatedService(service *zms.ServiceIdentity) (*string, error)
...
+	if tagValue != "" && service.Tags != nil {
+		currentTagValues := service.Tags[zms.CompoundName(tagKey)]
+		if currentTagValues != nil {
+			for _, curTagValue := range currentTagValues.List {
+				if tagValue != string(curTagValue) {
+					tagValueArr = append(tagValueArr, curTagValue)
+				}
+			}
+		}
+	}
```
dvirguttman, line 94: if this piece of code is relevant for all tag functions - please make it generic

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -6563,6 +6578,7 @@ Integer nullIfDefaultValue(int value, int defaultValue) {
 
     private void addTagsToRoles(Map<String, Role> roleMap, String domainName) {
 
+        // TODO change to the generic one
```
dvirguttman, line 6581: delete unused code

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -6713,6 +6845,31 @@ private void addTagsToGroups(Map<String, Group> groupMap, String domainName) {
         }
     }
 
+    Map<String, Map<String, TagValueList>> getDomainResourceTags(String domainName, String caller, String sqlStatement) {
+        final String funcCaller = "getDomain" + caller + "Tags";
```
dvirguttman, line 6849: you should not edit the caller..

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -6713,6 +6845,31 @@ private void addTagsToGroups(Map<String, Group> groupMap, String domainName) {
         }
     }
 
+    Map<String, Map<String, TagValueList>> getDomainResourceTags(String domainName, String caller, String sqlStatement) {
```
dvirguttman, line 6848: If I understand correctly - this function comes to replace the specific function per resource.
you can address this change as part of this pr or once it merged in different pr.

and, I think, `getResourceTags` is better naming

servers/zms/src/test/java/com/yahoo/athenz/zms/ZMSImplTest.java
```diff
@@ -33993,4 +33992,202 @@ public void testValidateGcpProjectDetails() {
...
+        final String auditRef = zmsTestInitializer.getAuditRef();
+
+        final String domainName = "sys.auth";
+        // put service with multiple tags
+        final String serviceWithTags = "swt-serviceWithTags";
+        final String tagKey = "tag-key";
+        List<String> multipleTagValues = Arrays.asList("val1", "val2");
+        ServiceIdentity service = zmsTestInitializer.createServiceObject(domainName, serviceWithTags,
+                "http://localhost", "/usr/bin/java", "root", "users", "host1");
+//        ServiceIdentity service = zmsTestInitializer.createServiceObject(domainName, serviceWithTags);
```
dvirguttman, line 34040: please go through your code and delete unused code

servers/zms/src/test/java/com/yahoo/athenz/zms/ZMSServiceTagsTest.java
```diff
@@ -0,0 +1,273 @@
...
+            hasServiceWithTags(serviceList, serviceWithTags, tagKey, tagValues, 2);
+            hasServiceWithTags(serviceList, serviceSingleTag, tagKey, singleTagValue, 1);
+            // ensure there are no more services
+            assertEquals(serviceList.getList().size(), 2);
+        }
+
+        public void testProcessUpdateServiceTags() {
+            ObjectStoreConnection conn = Mockito.mock(ObjectStoreConnection.class);
+
+        }
```
dvirguttman, line 129: ?

libs/go/zmscli/dump.go
royelbaum: show services use this function in order to show all the services return from the query.

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
royelbaum: it was part of the attempt to make it more generic

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
royelbaum: do we want to make it more generic? we said that this file should be simple as possible

servers/zms/src/test/java/com/yahoo/athenz/zms/ZMSServiceTagsTest.java
royelbaum: by mistake create another test function that "testQueryUpdateServiceWithTags"

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5961,11 +5962,17 @@ List<ServiceIdentity> setupServiceIdentityList(AthenzDomain domain, Boolean publ
...
+                    .filter(service -> filterByTag(tagKey, tagValue, service, ServiceIdentity::getTags))
+                    .collect(Collectors.toList());
+        }
+
         return services;
     }
 
     public ServiceIdentities getServiceIdentities(ResourceContext ctx, String domainName,
-            Boolean publicKeys, Boolean hosts) {
+            Boolean publicKeys, Boolean hosts, String tagKey, String tagValue) {
```
havetisyan, line 5975: We should have validate calls for tagKey and tagValue values to make sure they're CompoundNames if they're specified.

havetisyan (changes requested):

servers/zms/schema/updates/update-20230530.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS `zms_server`.`service_tags` (
+   `service_id` INT UNSIGNED NOT NULL,
+   `key` VARCHAR(64) CHARACTER SET 'utf8' COLLATE 'utf8_bin' NOT NULL,
+    `value` VARCHAR(256) NOT NULL,
+    PRIMARY KEY (`service_id`, `key`, `value`),
+    INDEX `tag_key_val_sec_idx` (`key` ASC, `value` ASC),
+    CONSTRAINT `fk_serivce_tag_service`
+    FOREIGN KEY (`service_id`)
```
havetisyan, line 8: can we indent the lines 8-11 to be consistent with the definition in the zms_server.sql

libs/go/zmscli/utils.go
```diff
@@ -157,3 +157,23 @@ func (cli Zms) RemoveAll(fullList []string, removeList []string) []string {
 	}
 	return newList
 }
+
+func (cli Zms) GetTagsAfterDeletion(resourceTags *zms.TagValueList, valuesToDelete []string) []zms.TagCompoundValue {
+	tagValueArr := make([]zms.TagCompoundValue, 0)
+	if resourceTags == nil || len(valuesToDelete) == 0 {
+		return tagValueArr
```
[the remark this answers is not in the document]
royelbaum: The only way to send an empty valuesToDelete array is not to send values at all which means that the user want to delete the tag key with all its values and not a specific value and this is why we want to return an empty array

servers/zms/schema/updates/update-20230530.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS `zms_server`.`service_tags` (
```
havetisyan, line 1: we seem to have duplicate update files now for the same table.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
royelbaum: I didn't understand this comment can you please explain? 
this is the same way that we do it for roles and groups.

havetisyan: looks like we missed in other places as well. as part of any input data, we must always validate the data. we can't allow any data unchecked from the user to follow through the system if we have specific constrains set in our rdl. So in this case both are defined as CompoundName but could also be null. So we need to validate the data with:

```
if (!StringUtils.isEmpty(tagKey) {
          validate(tagKey, TYPE_COMPOUND_NAME, caller);
}
if (!StringUtils.isEmpty(tagValue) {
          validate(tagValue, TYPE_COMPOUND_NAME, caller);
}
```

havetisyan (approved):

Issue #2201, opened 2023-06-06
request for wildcard-based skip domain determination


ysknkd:
In our environment, there are cases where some products dynamically create subdomains, but don't register the Athenz service with Athenz. Adding a domain to the ZMS property athenz.zms.validate_service_members_skip_domains each time a subdomain is created is operationally costly. Hence, we would like to determine the skip domain using a wildcard.

https://github.com/AthenZ/athenz/blob/ffa5cd11659d3eef5d2e2563fd8f2e13046b5872/servers/zms/conf/zms.properties#L434

We will be submitting a PR with our proposed implementation soon.

PR #2202, opened 2023-06-07
add wildcard support for ServiceMemberSkipDomains

hiragi-gkuth:
resolve https://github.com/AthenZ/athenz/issues/2201

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -3989,10 +3989,22 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         // are typically domains (like for ci/cd) where services
         // are dynamic and do not need to be registered in Athenz
 
-        if (!validateServiceMemberSkipDomains.contains(domainName)) {
-            if (dbService.getServiceIdentity(domainName, serviceName, true) == null) {
-                throw ZMSUtils.requestError("Principal " + memberName + " is not a valid service", caller);
+        for (String skipDomain : validateServiceMemberSkipDomains) {
+            // first, we perform validation using wildcards
+            if (skipDomain.endsWith("*")) {
+                String skipDomainPrefix = skipDomain.substring(0, skipDomain.length()-1);
+                if (domainName.startsWith(skipDomainPrefix)) {
+                    return;
+                }
             }
+            // then, we conduct a perfect match search
+            if (skipDomain.equals(domainName)) {
+                return;
+            }
+        }
+        // If it reaches here, check if the service exists
+        if (dbService.getServiceIdentity(domainName, serviceName, true) == null) {
+            throw ZMSUtils.requestError("Principal " + memberName + " is not a valid service", caller);
         }
     }
 


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3989,10 +3989,22 @@ void validateServicePrincipal(final String memberName, final String caller) {
...
+        for (String skipDomain : validateServiceMemberSkipDomains) {
+            // first, we perform validation using wildcards
+            if (skipDomain.endsWith("*")) {
+                String skipDomainPrefix = skipDomain.substring(0, skipDomain.length() - 1);
+                if (domainName.startsWith(skipDomainPrefix)) {
+                    return;
+                }
             }
+            // then, we conduct a perfect match search
+            if (skipDomain.equals(domainName)) {
```
havetisyan, line 4001: this should be "else if" otherwise we're wasting operations comparing the same value which already contains * against a domain name that will never include *

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -3989,10 +3989,22 @@ void validateServicePrincipal(final String memberName, final String caller) {
         // are typically domains (like for ci/cd) where services
         // are dynamic and do not need to be registered in Athenz
 
-        if (!validateServiceMemberSkipDomains.contains(domainName)) {
-            if (dbService.getServiceIdentity(domainName, serviceName, true) == null) {
-                throw ZMSUtils.requestError("Principal " + memberName + " is not a valid service", caller);
+        for (String skipDomain : validateServiceMemberSkipDomains) {
```
havetisyan, line 3992: validateServiceMemberSkipDomains was created as a Set to have simple contains check. if we're treating as a regular list instead, you should change the definition to be an ArrayList instead of a Set.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
hiragi-gkuth: thank you. I've fixed it.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
hiragi-gkuth: thank you. I've fixed it.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -830,7 +830,7 @@ void loadConfigurationSettings() {
 
         final String skipDomains = System.getProperty(
                 ZMSConsts.ZMS_PROP_VALIDATE_SERVICE_MEMBERS_SKIP_DOMAINS, "");
-        validateServiceMemberSkipDomains = new HashSet<>(Arrays.asList(skipDomains.split(",")));
+        validateServiceMemberSkipDomains = new ArrayList<>(Arrays.asList(skipDomains.split(",")));
```
havetisyan, line 833: No need to create a new ArrayList<> here. Arrays.asList already does the job and returns an ArrayList. So just call it:

validateServiceMemberSkipDomains = Arrays.asList(skipDomains.split(","));

hiragi-gkuth: I considered writing it that way, but Arrays.asList() and new ArrayList return different types. ([ref](https://docs.oracle.com/javase/8/docs/api/java/util/Arrays.html#asList-T...-))
As a result, there may be some confusion about this behavior in the future.
However, I am not too concerned about it, so I think Arrays.asList() is fine.
Thanks for pointing this out.

havetisyan (commented):
@hiragi-gkuth one last request.

https://github.com/AthenZ/athenz/blob/master/servers/zms/conf/zms.properties#L434

can you update the documentation to indicate that the domain name can now include * wildcard at the end of the domain name to providing subdomain matching.

hiragi-gkuth:
@havetisyan 

Thank you for your review. I've updated the documentation

hiragi-gkuth:
I forgot to commit with signed-off. I will rebase later.

PR #2206, opened 2023-06-11
Create support for policy tags

royelbaum:


servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -3982,9 +3995,153 @@ void getAthenzDomainPolicies(String domainName, int domainId, AthenzDomain athen
...
+                }
+            }
+        }
+    }
+
+    Map<String, Map<String, TagValueList>> getDomainPolicyTags(String domainName) {
+        final String funcCaller = "getDomain Policy Tags";
+        Map<String, Map<String, TagValueList>> domainResourceTags = null;
+
+        try (PreparedStatement ps = con.prepareStatement(SQL_GET_DOMAIN_POLICY_TAGS)) {
```
dvirguttman, line 4021: I think your Join query should ignore inactive policies.
In addition - please add multi-version policies tests to validate we are returning tags only for the active policy

servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
```diff
@@ -7155,6 +7162,17 @@ public void testUpdateDomainMembersExpirationObjectStoreFailure() {
...
+        Group group1 = createGroupObject(domainName, "group1", "user.john", "user.jane");
+        Group group2 = createGroupObject(domainName, "group1", "user.john", "user.jane");
+        group2.setMemberExpiryDays(3);
+        Group group3 = createGroupObject(domainName, "group1", "user.john", "user.jane");
+        group3.setGroupMembers(null);
+        List<Group> groupList = new ArrayList<>();
+        groupList.add(group1);
+        groupList.add(group2);
+        groupList.add(group3);
+        athenzDomain.setGroups(groupList);
```
dvirguttman, line 7174: why the groups here?

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
royelbaum: this query shouldn't ignore inactive policy version because in the end it uses for adding the tags for all the policies in a specific domain.

servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
royelbaum: needed to add coverage

royelbaum: to add coverage

PR #2212, opened 2023-06-23
Prevent buffer overflow

tokle:
The `notExpired` method currently always returns `false` under "normal" circumstances e.g when using zts defaults.


havetisyan:
@tokle are you using this class in production?

PR #2218, opened 2023-06-29
Fix miss usage of processTags in domain, group and role.

royelbaum:
There was i logic problem with processTags it always returned true which cause problems in processPolicyTags so i fix the logic of the code to return false when it didn't update any thing related to tags.

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
@@ -609,6 +609,7 @@ public class DBService implements RolesProvider {
             requestSuccess = con.updateRole(domainName, role);
         }
 
+        requestSuccess = processRoleTags(role, roleName, domainName, originalRole, con) || requestSuccess;
         // if we didn't update any roles then we need to return failure
 
         if (!requestSuccess) {
@@ -645,9 +646,7 @@ public class DBService implements RolesProvider {
             }
         }
 
-        if (!processRoleTags(role, roleName, domainName, originalRole, con)) {
-            return false;
-        }
+
 
         auditDetails.append('}');
         return true;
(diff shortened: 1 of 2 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -609,6 +609,7 @@ boolean processRole(ObjectStoreConnection con, Role originalRole, String domainN
             requestSuccess = con.updateRole(domainName, role);
         }
 
+        requestSuccess = processRoleTags(role, roleName, domainName, originalRole, con) || requestSuccess;
```
havetisyan, line 612: this is actually a wrong move. we cannot go ahead and process our role tags if the role update or insert operation failed. the original order is correct.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -677,6 +676,7 @@ boolean processGroup(ObjectStoreConnection con, Group originalGroup, final Strin
             requestSuccess = con.updateGroup(domainName, group);
         }
 
+        requestSuccess = processGroupTags(group, groupName, domainName, originalGroup, con) || requestSuccess;
```
havetisyan, line 679: this is actually a wrong move. we cannot go ahead and process our group tags if the group update or insert operation failed. the original order is correct.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -753,7 +750,7 @@ boolean processUpdateTags(Map<String, TagValueList> currentTags, Map<String, Tag
         if (originalTags == null || originalTags.isEmpty()) {
             if (currentTags == null || currentTags.isEmpty()) {
                 // no tags to process..
-                return true;
+                return false;
```
havetisyan, line 753: we can't just make this change. with the change, if there is nothing to process, it's a failure.

havetisyan (changes requested):
@royelbaum I don't think this PR is correct. 

I'm not still not clear what the original issue is so let's talk. We should not change the behavior of our functions.

PR #2219, opened 2023-06-29
Adding policy tags

royelbaum:
Finish the changes needed to support policy tags


clients/go/zms/model.go
```diff
@@ -2510,9 +2510,14 @@ type Policy struct {
 	Active *bool `json:"active,omitempty" rdl:"optional" yaml:",omitempty"`
 
 	//
-	// a description of the policy
+	// a description of the policy\
```
dvirguttman, line 2513: invalid character

clients/go/zms/zms_schema.go
```diff
@@ -358,7 +358,8 @@ func init() {
 	tPolicy.Field("caseSensitive", "Bool", true, nil, "If true, we should store action and resource in their original case")
 	tPolicy.Field("version", "SimpleName", true, nil, "optional version string, defaults to 0")
 	tPolicy.Field("active", "Bool", true, nil, "if multi-version policy then indicates active version")
-	tPolicy.Field("description", "String", true, nil, "a description of the policy")
+	tPolicy.Field("description", "String", true, nil, "a description of the policy\\")
```
dvirguttman, line 361: invalid character

clients/go/zms/zms_schema.go
```diff
@@ -1827,6 +1828,8 @@ func init() {
 	mGetPolicies.Input("domainName", "DomainName", true, "", "", false, nil, "name of the domain")
 	mGetPolicies.Input("assertions", "Bool", false, "assertions", "", true, false, "return list of assertions in the policy")
 	mGetPolicies.Input("includeNonActive", "Bool", false, "includeNonActive", "", true, false, "include non-active policy versions")
+	mGetPolicies.Input("tagKey", "CompoundName", false, "tagKey", "", true, nil, "flag to query all groups that have a given tagName")
+	mGetPolicies.Input("tagValue", "CompoundName", false, "tagValue", "", true, nil, "flag to query all groups that have a given tag name and value")
```
dvirguttman, line 1832: should not be groups

core/zms/src/main/java/com/yahoo/athenz/zms/ZMSSchema.java
```diff
@@ -311,7 +313,8 @@ private static Schema build() {
             .field("caseSensitive", "Bool", true, "If true, we should store action and resource in their original case")
             .field("version", "SimpleName", true, "optional version string, defaults to 0")
             .field("active", "Bool", true, "if multi-version policy then indicates active version")
-            .field("description", "String", true, "a description of the policy");
+            .field("description", "String", true, "a description of the policy\\")
```
dvirguttman, line 316: invalid character

core/zms/src/main/rdl/Policy.rdli
```diff
@@ -27,10 +27,12 @@ resource PolicyList GET "/domain/{domainName}/policy?limit={limit}&skip={skip}"
 }
 
 //List policies provisioned in this namespace.
-resource Policies GET "/domain/{domainName}/policies?assertions={assertions}&includeNonActive={includeNonActive}" {
+resource Policies GET "/domain/{domainName}/policies?assertions={assertions}&includeNonActive={includeNonActive}&tagKey={tagKey}&tagValue={tagValue}" {
     DomainName domainName; //name of the domain
     Bool assertions (optional, default=false); // return list of assertions in the policy
     Bool includeNonActive (optional, default=false); // include non-active policy versions
+    CompoundName tagKey (optional); //flag to query all groups that have a given tagName
+    CompoundName tagValue (optional); //flag to query all groups that have a given tag name and value
```
dvirguttman, line 35: should not be group

core/zms/src/main/rdl/Policy.tdl
```diff
@@ -47,7 +47,8 @@ type Policy Struct {
     Bool caseSensitive (optional); // If true, we should store action and resource in their original case
     SimpleName version (optional); // optional version string, defaults to 0
     Bool active (optional); // if multi-version policy then indicates active version
-    String description (optional); //a description of the policy
+    String description (optional); //a description of the policy\
```
dvirguttman, line 50: invalid character

libs/go/zmscli/cli.go
```diff
@@ -466,6 +466,24 @@ func (cli Zms) EvalCommand(params []string) (*string, error) {
...
+				return cli.DeletePolicyTags(dn, args[0], args[1], args[2])
+			}
+		case "show-policies":
+			if argc == 0 {
+				return cli.ShowPolicies(dn, "", "")
+			} else if argc == 1 {
+				return cli.ShowPolicies(dn, args[0], "")
+			} else if argc == 2 {
+				return cli.ShowPolicies(dn, args[0], args[1])
+			}
```
dvirguttman, line 486: make sure to document your new commands

libs/go/zmscli/policy.go
```diff
@@ -417,3 +417,97 @@ func (cli Zms) SetActivePolicyVersion(dn string, pn string, version string) (*st
...
+	if tagValue != "" && Policy.Tags != nil {
+		currentTagValues := Policy.Tags[zms.CompoundName(tagKey)]
+		if currentTagValues != nil {
+			for _, curTagValue := range currentTagValues.List {
+				if tagValue != string(curTagValue) {
+					tagValueArr = append(tagValueArr, curTagValue)
+				}
+			}
+		}
+	}
```
dvirguttman, line 443: please delete unused code - and the same as your service pull request - if code can be generic - do it

servers/zms/pom.xml
```diff
@@ -196,6 +196,12 @@
       <groupId>software.amazon.awssdk</groupId>
       <artifactId>dynamodb-enhanced</artifactId>
     </dependency>
+    <dependency>
+      <groupId>ch.qos.logback</groupId>
+      <artifactId>logback-core</artifactId>
+      <version>1.4.7</version>
+      <scope>test</scope>
+    </dependency>
```
dvirguttman, line 204: why this dependency is needed?
additionally, if it is needed - the version should not be here

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5035,17 +5035,25 @@ List<Policy> setupPolicyList(AthenzDomain domain, Boolean assertions, Boolean ve
...
+                            .setTags(policy.getTags());
                     policies.add(newPolicy);
                 }
             }
         }
 
+        if (tagKey != null) {
+            policies = policies.stream()
+                    .filter(policy -> filterByTag(tagKey, tagValue, policy, Policy::getTags))
+                    .filter(policy -> (versions != null && versions) || policy.active)
```
dvirguttman, line 5048: few things:
1. this one is better : `versions == Boolean.TRUE`
2. do not access directly to attributes, but `policy.getActive()`
3. I think in this case multiple filter is incorrect, but single one:
```
policy -> filterByTag(tagKey, tagValue, policy, Policy::getTags) && (versions == Boolean.TRUE || policy.getActive() == Boolean.TRUE )
```

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSResources.java
```diff
@@ -2073,13 +2073,15 @@ public PolicyList getPolicyList(
     public Policies getPolicies(
         @Parameter(description = "name of the domain", required = true) @PathParam("domainName") String domainName,
         @Parameter(description = "return list of assertions in the policy", required = false) @QueryParam("assertions") @DefaultValue("false") Boolean assertions,
-        @Parameter(description = "include non-active policy versions", required = false) @QueryParam("includeNonActive") @DefaultValue("false") Boolean includeNonActive) {
+        @Parameter(description = "include non-active policy versions", required = false) @QueryParam("includeNonActive") @DefaultValue("false") Boolean includeNonActive,
+        @Parameter(description = "flag to query all groups that have a given tagName", required = false) @QueryParam("tagKey") String tagKey,
+        @Parameter(description = "flag to query all groups that have a given tag name and value", required = false) @QueryParam("tagValue") String tagValue) {
```
dvirguttman, line 2078: should not be group

utils/zpe-updater/util/canonicalstring_test.go
```diff
@@ -8,8 +8,8 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
 	"github.com/AthenZ/athenz/clients/go/zts"
+	"github.com/stretchr/testify/assert"
```
dvirguttman, line 12: ?

utils/hostdoc/hostdoc_test.go
```diff
@@ -37,37 +37,37 @@ func TestProcess(t *testing.T) {
 	}{
 		{
```
dvirguttman, line 38: why do you have changes in this file?

royelbaum: I think I have some auto-formatting that caused it

servers/zms/schema/updates/update-20230606.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE IF NOT EXISTS `zms_server`.`policy_tags` (
+    `policy_id` INT UNSIGNED NOT NULL,
+    `key` VARCHAR(64) CHARACTER SET 'utf8' COLLATE 'utf8_bin' NOT NULL,
+    `value` VARCHAR(256) NOT NULL,
+    PRIMARY KEY (`policy_id`, `key`, `value`),
+    INDEX `tag_key_val_sec_idx` (`key` ASC, `value` ASC),
+    CONSTRAINT `fk_policy_tag_policy0`
```
havetisyan, line 7: can we get rid of 0 at the end of policy0

servers/zms/schema/updates/update-20230606.sql
```diff
@@ -0,0 +1,12 @@
...
+    FOREIGN KEY (`policy_id`)
```
havetisyan, line 8: can we indent lines 8-11 to be consistent with definition in zms_server.sql

servers/zms/schema/zms_server.sql
```diff
@@ -561,6 +561,23 @@ CREATE TABLE IF NOT EXISTS `zms_server`.`service_domain_dependency` (
...
+-- -----------------------------------------------------
+-- Table `zms_server`.`policy_tags`
+-- -----------------------------------------------------
+CREATE TABLE IF NOT EXISTS `zms_server`.`policy_tags` (
+  `policy_id` INT UNSIGNED NOT NULL,
+  `key` VARCHAR(64) CHARACTER SET 'utf8' COLLATE 'utf8_bin' NOT NULL,
+  `value` VARCHAR(256) NOT NULL,
+  PRIMARY KEY (`policy_id`, `key`, `value`),
+  INDEX `tag_key_val_sec_idx` (`key` ASC, `value` ASC),
+  CONSTRAINT `fk_policy_tag_policy0`
```
havetisyan, line 573: let's update the zms_server.mwb to remove the 0 from policy0

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -3760,6 +3776,12 @@ Policy getPolicy(ObjectStoreConnection con, String domainName, String policyName
         if (policy != null) {
             policy.setAssertions(con.listAssertions(domainName, policyName, version));
         }
+
+        Map<String, TagValueList> policyTags = con.getPolicyTags(domainName, policyName, version);
```
havetisyan, line 3780: there is no reason to get the policyTags if the policy is null. so we should have:

if (policy != null) {
   Map<String, TagValueList> policyTags = con.getPolicyTags(domainName, policyName, version);
...

and not check if policy != null after we get the tags

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -3892,56 +3914,17 @@ void executePutDomainMeta(ResourceContext ctx, Domain domain, DomainMeta meta,
...
 
-        if (originalDomain == null || originalDomain.getTags() == null || originalDomain.getTags().isEmpty()) {
-            if (domainTags == null || domainTags.isEmpty()) {
-                // no tags to process
-                return true;
-            }
-            if (!con.insertDomainTags(domainName, domainTags)) {
-                return false;
-            }
+        BiFunction<ObjectStoreConnection, Map<String, TagValueList>, Boolean> insertOp = (ObjectStoreConnection c, Map<String, TagValueList> tags) -> c.insertDomainTags(domainName, tags);
```
havetisyan, line 3917: wasn't this already merged?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5099,17 +5099,24 @@ List<Policy> setupPolicyList(AthenzDomain domain, Boolean assertions, Boolean ve
...
+                            .setActive(policy.getActive())
+                            .setTags(policy.getTags());
                     policies.add(newPolicy);
                 }
             }
         }
 
+        if (tagKey != null) {
+            policies = policies.stream()
+                    .filter(policy -> (filterByTag(tagKey, tagValue, policy, Policy::getTags) && ((versions == Boolean.TRUE) || policy.getActive() == Boolean.TRUE)))
```
havetisyan, line 5111: why are we including 

((versions == Boolean.TRUE) || policy.getActive() == Boolean.TRUE)

in the filter. Shouldn't we be only looking at the tag key?

servers/zms/src/main/java/com/yahoo/athenz/zms/store/ObjectStore.java
```diff
@@ -41,8 +41,9 @@ public interface ObjectStore {
      * @param domainLimit domain tags limit
      * @param roleLimit role tags limit
      * @param groupLimit group tags limit
+     * @param policyLimit policy tags limit
      */
-    void setTagLimit(int domainLimit, int roleLimit, int groupLimit);
+    void setTagLimit(int domainLimit, int roleLimit, int groupLimit, int policyLimit);
```
havetisyan, line 46: this api is also modified in the service tag so it's not going to cause a failure with a merge

servers/zms/src/main/java/com/yahoo/athenz/zms/store/impl/jdbc/JDBCConnection.java
```diff
@@ -4006,9 +4019,154 @@ void getAthenzDomainPolicies(String domainName, int domainId, AthenzDomain athen
...
+                Map<String, TagValueList> policyTag = domainPolicyTags.get(ZMSUtils.extractPolicyName(domainName, policyEntry.getValue().name) + ":" + policyEntry.getValue().getVersion());
+                if (policyTag != null) {
+                    policyEntry.getValue().setTags(policyTag);
+                }
+            }
+        }
+    }
+
+    Map<String, Map<String, TagValueList>> getDomainPolicyTags(String domainName) {
+        final String funcCaller = "getDomain Policy Tags";
```
havetisyan, line 4042: caller  is the func name without any spaces - this allows lookups in Yamas.

utils/zpe-updater/util/canonicalstring_test.go
```diff
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"testing"
 
-	"github.com/stretchr/testify/assert"
```
havetisyan, line 11: why sure why this was removed and more interestingly how did the code compile??

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
royelbaum: it was only for role and group i added this changed for domain as well

utils/zpe-updater/util/canonicalstring_test.go
royelbaum: more interesting is that i didn't touched this file in my changes i will check it

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
dvirguttman: The previous condition was:
```
if (versions != Boolean.TRUE && policy.getActive() == Boolean.FALSE) {
                    continue;
}
```
That means we would like to collect only active policies or if the request wants to get all policies.
If tag key/value exists, we need to add on top of that (only if the previous condition accepted) tag key-value filtering.

dvirguttman: sorry you are right. I reviewed it locally and missed that..

> in the filter. Shouldn't we be only looking at the tag key?
We probably want to filter by tag value also .. correct?

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
havetisyan: this was not addressed - this is a wasted DB call if the policy is null. use the if (policy != null) block from 3776 to avoid unnecessary calls if policy is null

havetisyan (changes requested):

libs/go/zmscli/utils.go
```diff
@@ -157,3 +157,18 @@ func (cli Zms) RemoveAll(fullList []string, removeList []string) []string {
 	}
 	return newList
 }
+
+func (cli Zms) GetTagsAfterDeletion(resourceTags *zms.TagValueList, valuesToDelete []string) []zms.TagCompoundValue {
+	tagValueArr := make([]zms.TagCompoundValue, 0)
+	if resourceTags == nil || len(valuesToDelete) == 0 {
```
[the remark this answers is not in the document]
royelbaum: I answer this comment in the service pr, the only case that this function gets an empty array is when the user enters only key value which means he wants to delete all the values in the list and therefore we return an empty array.

havetisyan (approved):

PR #2220, opened 2023-06-29
Go code to get SIA certs from a CGF (Google Cloud Function)

gilad-bendor:


libs/go/sia/gcp/functions/gcf_get_sia_cert.go
```diff
@@ -0,0 +1,263 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+)
+
+// GcfGetSiaCerts this method can be called from within a GCF (Google Cloud Function) -
+//
+//	to get an Athenz certificate from ZTS.
+//
+// This file should usually be copied without changes into the GCF source-code.
```
havetisyan, line 25: not sure what this is referring to. nobody should copy the source file. instead, they'll just import the module and call the methods, so let's remove this comment

libs/go/sia/gcp/functions/gcf_get_sia_cert.go
```diff
@@ -0,0 +1,263 @@
+package functions
```
havetisyan, line 1: need to add Athenz License header block

libs/go/sia/gcp/functions/gcf_get_sia_cert.go
```diff
@@ -0,0 +1,263 @@
...
+	athenzDomain = strings.ToLower(athenzDomain)
+	athenzService = strings.ToLower(athenzService)
+	athenzProvider = strings.ToLower(athenzProvider)
+
+	// Get an identity-document for this GCF from GCP.
+	attestationData, err := getGcpFunctionAttestationData(ztsUrl)
+	if err != nil {
+		return nil, err
+	}
+	//log.Printf("GCP Attestation Data: %s", attestationData) // commented out - sensitive info
```
havetisyan, line 47: let's remove all the commented log entries

libs/go/sia/gcp/functions/gcf_get_sia_cert.go
```diff
@@ -0,0 +1,263 @@
...
+		return nil, err
+	}
+
+	siaCertData.privateKey = privateKey
+	siaCertData.privateKeyPem = string(privateKeyPem)
+	return siaCertData, nil
+}
+
+// SiaCertData response of GcfGetSiaCerts()
+type SiaCertData struct {
```
havetisyan, line 104: No need to define a new struct. This is already available as part of our RDL

import 	"github.com/AthenZ/athenz/clients/go/zts"

and generate and return zts.InstanceIdentity object

havetisyan (changes requested):

libs/go/sia/gcp/functions/gcf_get_sia_cert.go
gilad-bendor: Archeological remnant removed...

libs/go/sia/gcp/functions/gcf_get_sia_cert.go
gilad-bendor: :+1:

libs/go/sia/gcp/functions/gcf_get_sia_cert.go
gilad-bendor: 👍

libs/go/sia/gcp/functions/gcf_get_sia_cert.go
gilad-bendor: Thing is, zts.InstanceIdentity is JSON-able: it only contain PEM stuff.
My SiaCertData also contains rsa.PrivateKey and x509.Certificate objects - so usage is simpler.
Also, zts.InstanceIdentitycontains many properties that I will not fill-in, and it would be confusing to users.
I would not make this change, but will yield to your judgment...
Are you sure you want to go that way?

havetisyan (approved):

PR #2237, opened 2023-07-13
Add support for service tag

royelbaum:
Screenshot 2023-07-18 at 16 32 10
Screenshot 2023-07-18 at 16 33 12


ui/src/components/header/CollectionDetails.js
```diff
@@ -16,8 +16,8 @@
 import styled from '@emotion/styled';
 import DateUtils from '../utils/DateUtils';
 import React from 'react';
-import { selectTimeZone } from '../../redux/selectors/domains';
-import { connect } from 'react-redux';
+import {connect} from "react-redux";
+import {selectTimeZone} from "../../redux/selectors/domains";
```
jimmytsang, line 20: are we using double quotes now?

ui/src/pages/domain/[domain]/service/[service]/tags.js
```diff
@@ -0,0 +1,222 @@
...
+            state,
+            props.domainName,
+            props.serviceName
+        ),
+        serviceTags: selectServiceTags(
+            state,
+            props.domainName,
+            props.serviceName
+        ),
+    };
```
Comment (jimmytsang) [inline ui/src/pages/domain/[domain]/service/[service]/tags.js:213]: can we get a screen recording of the new page linked to this pr? Would like to have a visual for better review.

jimmytsang:
changes look good so far, hoping to see a visual from screen recording 👍

ui/src/components/header/CollectionDetails.js
royelbaum: no it was my bad prettier wasn't on

ui/src/pages/domain/[domain]/service/[service]/tags.js
Comment (royelbaum) [reply ui/src/pages/domain/[domain]/service/[service]/tags.js:213]: added 2 screenshots to the description.

ui/src/components/history/CollectionHistoryList.js
```diff
@@ -23,9 +23,9 @@ import Alert from '../denali/Alert';
 import { MODAL_TIME_OUT } from '../constants/constants';
 import DateUtils from '../utils/DateUtils';
 import { selectIsLoading } from '../../redux/selectors/loading';
-import { selectTimeZone } from '../../redux/selectors/domains';
 import { connect } from 'react-redux';
 import { ReduxPageLoader } from '../denali/ReduxPageLoader';
+import {selectTimeZone} from "../../redux/selectors/domains";
```
jimmytsang, line 28: double quotes shouldn't be here right?

ui/src/components/tag/TagList.js
```diff
@@ -217,6 +217,8 @@ class TagList extends React.Component {
                 tags: collectionDetails.tags,
             };
         }
+        // For Service and Policy just return the collection because we don't build a new metaData object.
+        return collectionDetails;
```
jimmytsang, line 221: 👍  thanks roy

ui/src/pages/domain/[domain]/service/[service]/tags.js
```diff
@@ -0,0 +1,222 @@
...
+            window.location.reload();
+            return <div />;
+        }
+        const err = this.props.error || this.state.error;
+        if (err) {
+            return <Error err={err} />;
+        }
+
+        return isLoading.includes('getDomainData') ? (
+            <ReduxPageLoader message={'Loading domain data'} />
```
Comment (jimmytsang) [inline ui/src/pages/domain/[domain]/service/[service]/tags.js:143]: should we check for `isLoading.includes('getServices')` too?

ui/src/pages/domain/[domain]/service/[service]/tags.js
```diff
@@ -0,0 +1,222 @@
...
+            serviceDetails,
+            serviceTags,
+            _csrf,
+            isLoading,
+        } = this.props;
+        if (reload || this.state.reload) {
+            window.location.reload();
+            return <div />;
+        }
+        const err = this.props.error || this.state.error;
```
Comment (jimmytsang) [inline ui/src/pages/domain/[domain]/service/[service]/tags.js:137]: Minor: you can do this here for cleaner code
```
serviceDetails = serviceDetails || {};
```
then replace lines 161, 168,  181... with just
```
collectionDetails={serviceDetails}
```

ui/src/redux/selectors/services.js
```diff
@@ -41,6 +41,20 @@ export const selectService = (state, domainName, serviceName) => {
...
+            getFullName(domainName, serviceDelimiter, serviceName)
+        ] &&
+        state.services.services[
+            getFullName(domainName, serviceDelimiter, serviceName)
+        ].tags
+        ? state.services.services[
+              getFullName(domainName, serviceDelimiter, serviceName)
+          ].tags
+        : {};
+};
```
jimmytsang, line 56: we should add a helper function 

```
function getSafe(fn, defaultVal) {
    try {
      return fn();
    } catch (e) {
      return defaultVal;
    }
  };
```
then we can just do 
```
return getSafe(state.services.services[
    getFullName(domainName, serviceDelimiter, serviceName)].tags, {});
```

ui/src/redux/thunks/collections.js
```diff
@@ -328,7 +328,7 @@ export const updateTags =
             dispatch(
                 updateTagsToStore(
                     getFullCollectionName(domain, collectionName, category),
-                    detail.tags,
+                    detail,
```
jimmytsang, line 331: can you help me understand this change?

royelbaum: Previously, we had a conditional check for handling role and group tags in the redux reducers files. If the tags of a collection were empty, we used to create a collection(role\group) with only the tags as an attribute. However, I made a decision to change this approach. Now, regardless of whether the tags are empty or not, we will always create the collection and add it to the store so we need to send the collection data and not only its tags.

ui/src/pages/domain/[domain]/service/[service]/tags.js
Comment (royelbaum) [reply ui/src/pages/domain/[domain]/service/[service]/tags.js:137]: serviceDetails is a const because it does not suppose to be changed on the page, but I took your approach and change the lines to serviceDetails || {} which is more elegant.

ui/src/pages/domain/[domain]/service/[service]/tags.js
Comment (royelbaum) [reply ui/src/pages/domain/[domain]/service/[service]/tags.js:143]: no, as you will see we do it on every page because we divide the domain data part to the main page in the sub-components as tagsList we check if isLoading.length !== 0 and if so we show loading in the component itself.

ui/src/redux/selectors/services.js
royelbaum: this is a great idea I think we can implement it on all the selectors.

jimmytsang (approved):
👍 nice work roy!

chandrasekhar1996 (approved):

chandrasekhar1996 (approved):

PR #2253, opened 2023-07-25
Enable to get policy with tags.

royelbaum:
The function getPolicy was passing a null value for the "version" variable, which resulted in the policy being returned without the tags. This is because each policy version is associated with its own set of tags.

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
@@ -3794,7 +3794,7 @@ public class DBService implements RolesProvider {
         Policy policy = con.getPolicy(domainName, policyName, version);
         if (policy != null) {
             policy.setAssertions(con.listAssertions(domainName, policyName, version));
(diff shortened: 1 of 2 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -3794,7 +3794,7 @@ Policy getPolicy(ObjectStoreConnection con, String domainName, String policyName
         Policy policy = con.getPolicy(domainName, policyName, version);
         if (policy != null) {
             policy.setAssertions(con.listAssertions(domainName, policyName, version));
-            policy.setTags(con.getPolicyTags(domainName, policyName, version));
+            policy.setTags(con.getPolicyTags(domainName, policyName, version != null ? version : policy.getVersion()));
```
havetisyan, line 3797: This change should not be necessary. In the last PR I fixed the logic to handle versions correctly. So please verify that your test case works without this change.

havetisyan (changes requested):

PR #2265, opened 2023-08-03
Generic Kubernetes provider for Athenz

abvaidya:


libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceK8SProvider.java
```diff
@@ -0,0 +1,274 @@
...
+import javax.net.ssl.SSLContext;
+import java.util.*;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.TimeUnit;
+
+import static com.yahoo.athenz.common.server.util.config.ConfigManagerSingleton.CONFIG_MANAGER;
+import static com.yahoo.athenz.instance.provider.impl.InstanceAWSProvider.*;
+import static com.yahoo.athenz.instance.provider.impl.InstanceGCPProvider.*;
+
+public class InstanceK8SProvider implements InstanceProvider {
```
havetisyan, line 42: so we can't support both aws and gcp at the same time since the provider will only create a single oidcissuervalidator?

abvaidya: oh yeah. Great catch! will fix that.

abvaidya: actually, we are not creating a single OIDCIssuerValidator but a single OIDCIssuerValidatorFactory which can hand out instance of DefaultAWSOIDCIssuerValidator OR DefaultGCPOIDCIssuerValidator.

havetisyan: let's discuss this Monday since I'm not sure this is as generic as we want it to be.

havetisyan (approved):

PR #2267, opened 2023-08-07
add wildcard support for ServiceSkipDomains

ricordanza:
## description
We have enabled the use of wildcards in the ServiceSkipDomains.

## reference
I implemented it based on the reference from https://github.com/AthenZ/athenz/pull/2202.

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -3534,7 +3534,7 @@ public class ZTSImpl implements KeyStore, ZTSHandler {
                     if (domainName.startsWith(serviceSkipDomainPrefix)) {
                         return;
                     }
-                } else if (validateServiceSkipDomains.contains(domainName)) {
+                } else if (serviceSkipDomain.equals(domainName)) {
                     // if skipDomain doesn't have wildcard, we conduct a perfect match search
                     return;
                 }
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -3526,18 +3526,16 @@ public class ZTSImpl implements KeyStore, ZTSHandler {
         // need to check anything
 
         final String domainName = domainData.getName();
-        if (domainName != null) {
-            for (String serviceSkipDomain : validateServiceSkipDomains) {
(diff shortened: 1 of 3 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3526,8 +3526,19 @@ void validateInstanceServiceIdentity(DomainData domainData, final String service
...
-            return;
+        if (domainName != null) {
+            for (String serviceSkipDomain : validateServiceSkipDomains) {
+                // first, we perform validation using wildcards
+                if (serviceSkipDomain.endsWith("*")) {
+                    String serviceSkipDomainPrefix = serviceSkipDomain.substring(0, serviceSkipDomain.length() - 1);
+                    if (domainName.startsWith(serviceSkipDomainPrefix)) {
+                        return;
+                    }
+                } else if (validateServiceSkipDomains.contains(domainName)) {
```
havetisyan, line 3537: wrong check here to verify against the whole list every time. The correct line would be 

} else if (skipDomain.equals(domainName)) {

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
ricordanza: Thank you. I've fixed it.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3526,8 +3526,19 @@ void validateInstanceServiceIdentity(DomainData domainData, final String service
         // need to check anything
 
         final String domainName = domainData.getName();
-        if (validateServiceSkipDomains.contains(domainName)) {
-            return;
+        if (domainName != null) {
```
havetisyan, line 3529: one more comment - looking at the server code, this can never be null so we should remove It, otherwise it's missing unit tests.

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
ricordanza: Thank you. 
Since it has been determined that domainName will not be null, I have removed the branching.
Additionally, I've made modifications to the test code to specify the domainName.

havetisyan (approved):

PR #2272, opened 2023-08-11
ability to add members by searching for their names

dvirguttman:


ui/src/redux/thunks/user.js
```diff
@@ -60,3 +60,12 @@ export const getUserResourceAccessList =
             dispatch(returnUserResourceAccessList());
         }
     };
+
+export const getUsers = async (prefix, dispatch) => {
+    try {
+        const userList = await API().getUsers(prefix);
+        return Promise.resolve(userList);
+    } catch (e) {
+        throw e;
```
abvaidya, line 69: might be a good idea to do `return Promise.reject(e);` instead of just `throe e;`

ui/src/components/member/AddMember.js
```diff
@@ -182,7 +183,23 @@ class AddMember extends React.Component {
     }
 
     inputChanged(key, evt) {
-        this.setState({ [key]: evt.target.value });
+        this.setState({ [key]: evt ? evt.value : '' });
+    }
+
+    userSearch(part) {
+        if (part.startsWith('user.')) {
```
abvaidya, line 190: - I believe `'user.'` needs to be configurable coming from constants.USER_DOMAIN
- we might want to keep an option to extend this feature for other types of principals ( headless users initially may be services down the road if its possible )

dvirguttman: basically, it is just backward compatibility. 
currently there is no need to type the prefix (user.) , since it is autocomplete by the user short id.
extend this piece of code might be very easy and I don't see a reason to make it more complex now ...

chandrasekhar1996 (commented):
Instead of making API calls for each keystroke, would it be better to store the userdata in redux layer and then return from there?
I think it will increase performance.

ui/src/redux/thunks/user.js
```diff
@@ -60,3 +60,12 @@ export const getUserResourceAccessList =
             dispatch(returnUserResourceAccessList());
         }
     };
+
+export const getUsers = async (prefix, dispatch) => {
+    try {
+        const userList = await API().getUsers(prefix);
```
royelbaum, line 66: This usage of thunk is not aligned with the recommended approach. By examining other thunk calls, a clear pattern emerges with two distinct mechanisms:

Defining a maximum validity duration for the retained data.
Storing the data within the Redux store, facilitating subsequent retrieval without further API requests.
To bring this implementation in line with the established pattern, a modification is necessary:

Validate if the data is present in the store and if its validity period has not elapsed. If so, return the stored data. If not, proceed to the next step.
In case the data is absent or its validity period has expired, initiate the API request. Upon receiving the response, save the obtained value in the Redux store for future use.

ui/src/components/utils/MemberUtils.js
```diff
@@ -38,4 +40,20 @@ export default class MemberUtils {
             validUsers,
         };
     }
+
+    static userSearch(part, userList) {
+        if (part.startsWith(USER_DOMAIN)) {
+            part = part.substring(5);
```
chandrasekhar1996, line 46: length of USER_DOMAIN+1

ui/src/components/utils/MemberUtils.js
```diff
@@ -38,4 +40,20 @@ export default class MemberUtils {
...
+        }
+        return getUsers(part, userList).then((r) => {
+            let usersArr = [];
+            r.forEach((u) =>
+                usersArr.push({
+                    name: u.name + ' [user.' + u.login + ']',
```
chandrasekhar1996, line 52: instead of "user.", should be "USER_DOMAIN", same in next line

ui/src/server/services/userService.js
```diff
@@ -77,5 +81,26 @@ function refreshUserData(config, extServiceClient) {
         });
 }
 
+async function readUsersFileFromDisk(fileName) {
```
chandrasekhar1996, line 84: couldn't find use of this function, should it be removed?

dvirguttman: it allows alternative way to load user file from disk, to be used by vzm-ui for example

Issue #2269, opened 2023-08-09
occasional ConcurrentModificationException encountered in ZTS's getResourceAccess method


ysknkd:
The getResourceAccess method in ZTS may occasionally encounter a ConcurrentModificationException.

Upon investigating this issue, it was discovered that there's a high likelihood of this exception occurring after the domain targeted by getResourceAccess is updated in ZMS. When there are changes to the domain, ZTS attempts to cache the domain, but we suspect that the following collection might be getting modified while being iterated upon in ZTSAuthorizer.

https://github.com/AthenZ/athenz/blob/036db67d36bb33af52f6927185679abf7e160c5b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSAuthorizer.java#L208

```
14:39:27.184 [qtp720770771-2230] WARN  o.eclipse.jetty.server.HttpChannel - /zts/v1/access/{action}/{resource}
jakarta.servlet.ServletException: jakarta.servlet.ServletException: java.util.ConcurrentModificationException
	at org.eclipse.jetty.server.handler.HandlerCollection.handle(HandlerCollection.java:157)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:122)
	at org.eclipse.jetty.server.Server.handle(Server.java:563)
	at org.eclipse.jetty.server.HttpChannel.lambda$handle$0(HttpChannel.java:505)
	at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:762)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:497)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:282)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:314)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
	at org.eclipse.jetty.io.ssl.SslConnection$DecryptedEndPoint.onFillable(SslConnection.java:558)
	at org.eclipse.jetty.io.ssl.SslConnection.onFillable(SslConnection.java:379)
	at org.eclipse.jetty.io.ssl.SslConnection$2.succeeded(SslConnection.java:146)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:100)
	at org.eclipse.jetty.io.SelectableChannelEndPoint$1.run(SelectableChannelEndPoint.java:53)
[...]

PR #2277, opened 2023-08-18
avoid ConcurrentModificationExceptions when processing domain updates

havetisyan:
the PR addresses https://github.com/AthenZ/athenz/issues/2269

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
@@ -897,22 +897,27 @@ public class DataStore implements DataCacheProvider, RolesProvider {
         }
 
         // Determine which roles have been deleted and should be removed from cache
-        final Set<String> validRoles = (roles != null) ? roles.stream().map(Role::getName).collect(Collectors.toSet()) : new HashSet<>();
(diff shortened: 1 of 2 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
```diff
@@ -897,22 +897,27 @@ void processDomainRoles(DomainData domainData, DataCache domainCache) {
...
-        final Set<String> validRoles = (roles != null) ? roles.stream().map(Role::getName).collect(Collectors.toSet()) : new HashSet<>();
+        // first we're going to extract our original roles. if we don't have
+        // any, then there is nothing to process, so we can return away
 
-        List<Role> deletedRoles = null;
         DataCache dataCache = getCacheStore().getIfPresent(domainData.getName());
-        if (dataCache != null) {
-            deletedRoles = dataCache.getDomainData().getRoles();
-            if (deletedRoles != null && !validRoles.isEmpty()) {
-                deletedRoles.removeIf(item -> validRoles.contains(item.getName()));
```
havetisyan, line 907: here is what the issue is. deleteRoles is pointing to our cache so we cannot modify that list

abvaidya (approved):

PR #2283, opened 2023-08-25
libs/go/sia/agent/devel/ztsmock: fix dropped error

alrs:
This fixes a dropped `err` variable in `libs/go/sia/agent/devel/ztsmock`.


havetisyan (approved):

havetisyan:
Thanks!

PR #2284, opened 2023-08-28
add constructor to allow proxy configuration

ysknkd:
## Description

Adds a constructor to the ZTSClient for configuring a proxy.

### DIFF
--- a/clients/java/zts/src/test/java/com/yahoo/athenz/zts/ZTSClientTest.java
+++ b/clients/java/zts/src/test/java/com/yahoo/athenz/zts/ZTSClientTest.java
@@ -4065,4 +4065,29 @@ public class ZTSClientTest {
             assertEquals(400, ex.getCode());
         }
     }
+
+    @Test
+    public void testCanSetProxyConfiguration() {
+        ServiceIdentityProvider siaProvider = Mockito.mock(ServiceIdentityProvider.class);
+        ZTSClient client = new ZTSClient("http://localhost:4080/",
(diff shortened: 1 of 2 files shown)


clients/java/zts/src/test/java/com/yahoo/athenz/zts/ZTSClientTest.java
```diff
@@ -4065,4 +4065,29 @@ public void testPostExternalCredentialsRequest() throws IOException, URISyntaxEx
...
+
+    @Test
+    public void testInvalidProxyException() {
+        ServiceIdentityProvider siaProvider = Mockito.mock(ServiceIdentityProvider.class);
+
+        ZTSClient client = null;
+        try {
+            client = new ZTSClient("http://localhost:4080/",
+                    "invalid-proxy", "iaas.athenz", "ci", siaProvider);
+            fail();
```
havetisyan, line 4085: Can you please add test cases where

- domainName
- serviceName
- siaProvider

are null and IllegalArgumentException is thrown.

havetisyan (changes requested):

clients/java/zts/src/test/java/com/yahoo/athenz/zts/ZTSClientTest.java
ysknkd: Thank you for the feedback. I have added the tests.

havetisyan (approved):
Thanks

PR #2289, opened 2023-08-30
fix review group page error with expiry settings

jimmytsang:
Seems like we forgot to include `timeZone={this.props.timeZone}` in one of the components which was causing the issue in group page -> review tab when users have some defined expiry data

https://github.com/AthenZ/athenz/blob/master/ui/src/components/review/ReviewRow.js#L88
```
// this fn was getting passed undefined

this.localDate.getLocalDate(
    member.expiration,
    this.props.timeZone,
    this.props.timeZone
)
```


MartinTrojans (approved):

jimmytsang:
ideally we have functional test to catch these kinds of errors, need to prioritize fixing/adding functional tests in the future.

PR #2312, opened 2023-09-13
utils/zts-svccert: fix dropped errors

alrs:
This fixes two `err` variables that were being re-declared inside of an `if` block with `:=`, causing them to be lost before they could be checked.


havetisyan (approved):
Thanks!

PR #2316, opened 2023-09-18
fix - show add msd policy err

jimmytsang:
Problem: when user attempts to add policy via the microsegmentation tab in athenz ui ([example link](https://dev-ui.athenz.ouryahoo.com/domain/home.jtsang01/microsegmentation)), the add msd modal does not display the error caught.

Solution: make sure parent component only renders child component once
image


abvaidya:
taken care by #2317

PR #2317, opened 2023-09-19
Show msd error

mendi160:


jimmytsang (commented):
Awesome, I tested this and it indeed solves the bug on my local. I am curious however 

> How come in the add role experience from AddRole.js which calls redux getRoles  -> dispatch(loadingInProcess('getRoles')) does not have the same defect triggering the parent component’s ReduxLoader?

jimmytsang:
This PR looks good but can you please fix the pr build failure?

PR #2323, opened 2023-09-26
fetch verification key from server during access token validation

ricordanza:
# description
Currently, AuthZpeClient is limited to dynamically fetching public keys exclusively during RoleToken verifications.
We aim to expand this functionality to also dynamically retrieve public keys when verifying AccessToken.

### DIFF
--- a/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/jwts/JwtsSigningKeyResolver.java
+++ b/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/jwts/JwtsSigningKeyResolver.java
@@ -35,6 +35,8 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.concurrent.ConcurrentHashMap;
 
+import static com.yahoo.athenz.auth.AuthorityConsts.AUTH_PROP_MILLIS_BETWEEN_ZTS_CALLS;
+
 public class JwtsSigningKeyResolver implements SigningKeyResolver {
 
     public static final String ZTS_PROP_ATHENZ_CONF           = "athenz.athenz_conf";
@@ -49,6 +51,10 @@ public class JwtsSigningKeyResolver implements SigningKeyResolver {
     private static long millisBetweenZtsCalls;
 
     ConcurrentHashMap<String, PublicKey> publicKeys;
+
+    static {
+        setMillisBetweenZtsCalls(Long.parseLong(System.getProperty(AUTH_PROP_MILLIS_BETWEEN_ZTS_CALLS, "86400000")));
+    }
 
     static ObjectMapper initJsonMapper() {
         ObjectMapper mapper = new ObjectMapper();
--- a/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/jwts/JwtsSigningKeyResolver.java
+++ b/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/jwts/JwtsSigningKeyResolver.java
@@ -45,6 +45,9 @@ public class JwtsSigningKeyResolver implements SigningKeyResolver {
     private static final ObjectMapper JSON_MAPPER = initJsonMapper();
     private final SSLContext sslContext;
     private final String jwksUri;
+    private final boolean skipConfig;
+    private static long lastZtsJwkFetchTime;
+    private static long millisBetweenZtsCalls;
 
     ConcurrentHashMap<String, PublicKey> publicKeys;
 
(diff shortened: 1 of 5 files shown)


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/jwts/JwtsSigningKeyResolver.java
```diff
@@ -45,6 +45,9 @@ public class JwtsSigningKeyResolver implements SigningKeyResolver {
     private static final ObjectMapper JSON_MAPPER = initJsonMapper();
     private final SSLContext sslContext;
     private final String jwksUri;
+    private final boolean skipConfig;
+    private static long lastZtsJwkFetchTime;
+    private static long millisBetweenZtsCalls;
```
havetisyan, line 50: what is the default value for this setting? this change would require every user of the library to set the value otherwise with unknown key ids the library will be making a call to the server every time. we can probably start with a conserve once a day value.

havetisyan (changes requested):

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/jwts/JwtsSigningKeyResolver.java
```diff
@@ -45,6 +45,9 @@ public class JwtsSigningKeyResolver implements SigningKeyResolver {
     private static final ObjectMapper JSON_MAPPER = initJsonMapper();
     private final SSLContext sslContext;
     private final String jwksUri;
+    private final boolean skipConfig;
```
havetisyan, line 48: any reason why we're not storing this field if we're not using it anywhere else?

ricordanza: Thank you, as you mentioned, since we're no longer using this field from anywhere else, we decided to remove it.

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/token/jwts/JwtsSigningKeyResolver.java
ricordanza: The default value is set to 30 minutes. This setting can be modified in the property file.

https://github.com/AthenZ/athenz/blob/e1e5e5ff152422b1a1b8bcd2249f23c2761ddccf/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java#L182

In order to avoid increasing the server load by sending requests to ZTS every time an unknown key ID is specified, we have set it to make a request every 30 minutes.

havetisyan: you're incorrectly assuming that it's the only place that the library is used. it's not a server library. it's a common client shared library so there are other applications that are using that functionality.

ricordanza: I apologize for the mistake. There was an error in the comment. It's not a property file, but a system property.
When need to modify the settings while using the client library, can make use of system properties.

havetisyan: Doesn't matter if it can be setup using a system property. We cannot make changes in shared client libraries where we change its current behavior and technically could result in DOS attacks. You must set the default value for the millisBetweenZtsCalls field to one day and then you can change that value where it's needed (e.g. with zpu)

havetisyan (approved):

ricordanza:
@havetisyan 

By merging this PR, the main Screwdriver job has stopped.
https://cd.screwdriver.cd/pipelines/6606/builds/902642/steps/build

This issue is caused by insufficient coverage in athenz-zpe-java-client.
```
11:45:15 [WARNING] Rule violated for bundle athenz-zpe-java-client: lines covered ratio is 0.8702, but expected minimum is 0.8711
```

The problem occurs intermittently.
What we know is that depending on the timing, there may be cases where the absence of caching results in the code not being executed.
Running the job multiple times may lead to successful execution.
If you could provide commentary on https://github.com/AthenZ/athenz/issues/2340, as it requires attention.
If it is necessary to share the investigation results regarding this matter, it will be conducted at https://github.com/AthenZ/athenz/issues/2340.

The jobs that reacted to the PR have been successful, as shown below.
https://cd.screwdriver.cd/pipelines/6606/builds/902059/steps/build

havetisyan:
interesting since the last PR build shows that it passed successfully. I'm somewhat surprised about the intermittent comment unless there is some bug in the jococo plugin when calculating the percentages.

PR #2324, opened 2023-09-26
Extract and reuse duplicate helper function ExtractSignerInfo.

zhaoyonghe:
Extract a reused function and export it
```
	block, _ := pem.Decode(privateKeyPEM)
	if block == nil {
		return nil, x509.UnknownSignatureAlgorithm, fmt.Errorf("unable to load private key")
	}

	switch block.Type {
	case "EC PRIVATE KEY":
		key, err := x509.ParseECPrivateKey(block.Bytes)
		if err != nil {
			return nil, x509.UnknownSignatureAlgorithm, err
		}
		return key, x509.ECDSAWithSHA256, nil
	case "RSA PRIVATE KEY":
		key, err := x509.ParsePKCS1PrivateKey(block.Bytes)
		if err != nil {
			return nil, x509.UnknownSignatureAlgorithm, err
		}
		return key, x509.SHA256WithRSA, nil
	case "PRIVATE KEY":
		key, err := x509.ParsePKCS8PrivateKey(block.Bytes)
		if err != nil {
			return nil, x509.UnknownSignatureAlgorithm, err
		}
		switch k := key.(type) {
		case *ecdsa.PrivateKey:
			return k, x509.ECDSAWithSHA256, nil
		case *rsa.PrivateKey:
			return k, x509.SHA256WithRSA, nil
		default:
			// PKCS#8 format may contain multiple key types other than RSA / EC, but current ZMS / ZTS server implementation only supports RSA / EC private keys
			return nil, x509.UnknownSignatureAlgorithm, fmt.Errorf("unsupported private key type: %s", reflect.TypeOf(k).Name())
		}
	default:
		return nil, x509.UnknownSignatureAlgorithm, fmt.Errorf("unsupported private key type: %s", block.Type)
	}
```


libs/go/zmssvctoken/signer.go
```diff
@@ -40,43 +40,52 @@ func hashString(input string) ([]byte, error) {
 	return h.Sum(nil), nil
 }
 
-// NewSigner creates an instance of Signer using the given private key (ECDSA or RSA).
-func NewSigner(privateKeyPEM []byte) (Signer, error) {
+// ExtractSignerInfo extract crypto.Signer and x509.SignatureAlgorithm from the given private key (ECDSA or RSA).
```
havetisyan, line 43: is it possible to move this method to the athenzutils go library. it's somewhat confusing to use a function from zms named module in zts utils. that would provide better visibility and more common usage

havetisyan (changes requested):

libs/go/zmssvctoken/signer.go
zhaoyonghe: updated

havetisyan (approved):
Thanks!

havetisyan (approved):

PR #2330, opened 2023-09-28
Fix gcp login initially loading error no roles incorrectly

jimmytsang:
Problem: gcp sso page was not processing the resourceAccessList fetch correctly thus initially showing "Error you don't have any gcp roles..."

Solution: dispatch loading, loading success, and loading failure in the redux `getResourceAccessList` and render the `ReduxLoader`

### DIFF
--- a/ui/src/pages/gcp/login.js
+++ b/ui/src/pages/gcp/login.js
@@ -166,7 +166,6 @@ class GCPLoginPage extends React.Component {
             errorMessage: '',
             projectRoleMap: {},
             roleName: '',
-            isFetching: true,
         };
         this.getAssertionList = this.getAssertionList.bind(this);
         this.populateProjectRoleMap = this.populateProjectRoleMap.bind(this);
@@ -196,7 +195,6 @@ class GCPLoginPage extends React.Component {
         this.setState((prevState) => ({
             ...prevState,
             errorMessage: errorMessage,
-            isFetching: false,
         }));
     }
 
@@ -239,7 +237,6 @@ class GCPLoginPage extends React.Component {
         this.setState((prevState) => ({
             ...prevState,
             projectRoleMap,
-            isFetching: false,
         }));
     }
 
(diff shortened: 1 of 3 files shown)


ui/src/pages/gcp/login.js
```diff
@@ -303,62 +298,61 @@ class GCPLoginPage extends React.Component {
...
+                    </p>
+                    <p>
+                        If that does not work, please ask your Athenz domain
+                        admin for assistance.
+                    </p>
+                </ParentWrapperDiv>
+            </div>
+        );
+        return this.props.isLoading.length !== 0 ? (
+            <ReduxPageLoader message={'Loading microsegmentation data'} />
```
abvaidya, line 352: This error message doesnt look correct. Shouldnt it be something like "Loading GCP projects"?

abvaidya (changes requested):

ui/src/pages/gcp/login.js
```diff
@@ -303,62 +298,61 @@ class GCPLoginPage extends React.Component {
...
+                    <h3>
+                        Error: There are no GCP project roles associated with
+                        your account.
+                    </h3>
+                    <p>
+                        Check to make sure that your <tt>gcp.*</tt> roles
+                        contain your user.
+                        <br />
+                        E.g., if your account is 'jdoe', then{' '}
+                        <tt>gcp.fed.admin.user</tt> should have{' '}
```
havetisyan, line 341: this is a yahoo specific role so we should not included it here. Let's keep it generic and just refer to that your account should be in the configured federated roles that have access to the project and not list any specific role names.

jimmytsang: how's this? 
image
```
Error: There are no GCP project roles associated with your account.
Check to make sure that your gcp.* roles contain your user. Your account should be in the configured federated roles that have access to the project. If that does not work, please ask your Athenz domain admin for assistance.
```

ui/src/pages/gcp/login.js
jimmytsang: great catch, fixed.

abvaidya (approved):

PR #2346, opened 2023-10-12
Make KeyRefresher's background thread more efficient

jeffreytolar:
InputStream::read() reads only a single byte at a time; this results in many read() syscalls and inefficient single-byte updates to the MessageDigest.

It'd be most efficient to first do a `stat` call and compare file size and timestamps; that's a bigger change, however. (Using a [WatchService](https://docs.oracle.com/javase/8/docs/api/java/nio/file/WatchService.html) would be another option)

### DIFF
--- a/libs/java/cert_refresher/src/main/java/com/oath/auth/KeyRefresher.java
+++ b/libs/java/cert_refresher/src/main/java/com/oath/auth/KeyRefresher.java
@@ -20,7 +20,6 @@ import java.io.InputStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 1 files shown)


jeffreytolar:
This was found because of a key refresher leak.

With the leak still present, this changed resulted in a significant drop in CPU: 
image

In the top graph, the blue is system CPU and the green user CPU, both measured as `cpuTime / realTime`. This was on a 24-hyperthread machine, so when the total levels off around 24, the various key refreshers are saturating the CPU.

The bottom graph is showing that the thread leak is still present. (That's not an athenz bug)

havetisyan (approved):

PR #2399, opened 2023-11-08
update service provider manager test case

havetisyan:
# Description
the test case sets up provider list update every second, but it's possible if the build system is slow, the 10 threads verifying the list don't finish within a second before the list is refreshed. now, it's set to 5 secs so plenty of time for threads to complete their tasks.
# Contribution Checklist:

### DIFF
--- a/servers/zms/src/test/java/com/yahoo/athenz/zms/provider/ServiceProviderManagerTest.java
+++ b/servers/zms/src/test/java/com/yahoo/athenz/zms/provider/ServiceProviderManagerTest.java
@@ -39,8 +39,6 @@ import java.util.concurrent.*;
 
 import static com.yahoo.athenz.zms.ZMSConsts.*;
 import static java.util.stream.Collectors.toList;
-import static org.mockito.ArgumentMatchers.eq;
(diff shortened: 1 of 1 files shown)


servers/zms/src/test/java/com/yahoo/athenz/zms/provider/ServiceProviderManagerTest.java
```diff
@@ -64,7 +62,7 @@ public void clean() {
     public void testIsServiceProvider() throws InterruptedException, ExecutionException {
         String testDomainName = "test.domain";
         String testRoleName = "test_role";
-        long fetchFrequency = 1L; // For the test, fetch every second
+        long fetchFrequency = 5L; // For the test, fetch every second
```
MartinTrojans, line 65: Should we update the comment accordingly?

havetisyan: Yes, thanks.updated the comment

abvaidya (approved):

PR #2423, opened 2023-11-21
libs/go/sia/access/tokens: fix dropped test errors

alrs:
This fixes three dropped test `err` variables in `libs/go/sia/access/tokens`.


havetisyan (approved):
Thanks

PR #2424, opened 2023-11-21
support max member limit on roles and groups

havetisyan:
# Description
support max member limit on roles and groups

# Contribution Checklist:


servers/zms/schema/zms_server.sql
```diff
@@ -90,6 +90,7 @@ CREATE TABLE IF NOT EXISTS `zms_server`.`role` (
   `group_expiry_days` INT NOT NULL DEFAULT 0,
   `description` VARCHAR(4096) NOT NULL DEFAULT '',
   `delete_protection` TINYINT(1) NOT NULL DEFAULT 0,
+  `max_members` INT NOT NULL DEFAULT 0,
```
MartinTrojans, line 93: Should we have some other numbers as the default value instead of 0?

havetisyan: No, we can't introduce backward incompatible changes. Value of 0 indicates there is no limit which is the current behavior.

servers/zms/src/main/java/com/yahoo/athenz/zms/QuotaChecker.java
```diff
@@ -178,49 +194,93 @@ void checkGroupQuota(ObjectStoreConnection con, String domainName, Group group,
...
+
+        Membership membership = con.getRoleMember(domainName, roleName, memberName, 0, false);
+        if (membership.getIsMember()) {
+            return;
+        }
+
+        // so at this point we know that we'll be adding a new member to the
+        // role. so first let's count the number of role members
+
+        int roleMemberCount = con.countRoleMembers(domainName, roleName);
```
chandrasekhar1996, line 220: Is it meant to count the expired members also as valid role/group member in terms of count?

havetisyan: yes, the count includes everyone regardless if they're expired or not

chandrasekhar1996 (approved):

PR #2425, opened 2023-11-22
show notification bell icon more obviously with role/group to review

jimmytsang:
Essentially this PR featureFlag wraps the following logic ...

- fetchAllRolesToReview for user (new)

- fetchAllGroupsToReview for user (new)

- fetchPendingMembersToReview (existing)

if there are any results from the above fetches , show notification icon like the following 
image

Important files to review are

- HeaderMenu.*js
- redux/*.js
- */api.js


jimmytsang:
I'm planning on adding functional tests in future pr after the feature works end to end

ui/src/components/header/HeaderMenu.js
```diff
@@ -87,68 +92,103 @@ const HeaderMenuUserDiv = styled.div`
...
+
+    render() {
+        let props = this.props;
+        let reviewNotificationRedDot = '';
+        let icon = 'notification';
+        let clickIcon = this.onClickFunction.bind(
+            this,
+            PageUtils.workflowAdminPage()
+        );
+        if (props.pending || props.reviewRoles || props.reviewGroups) {
```
MartinTrojans, line 137: Will that be better to put all these conditions together? And if necessary, we can have a function to check the status of pending members, or review roles or groups, and you can just pass these values to that function

MartinTrojans (approved):

ui/src/components/header/HeaderMenu.js
jimmytsang: ok good suggestion, i can definitely write a helper function to simplify this to 
```
if (shouldAlertUser(...))
```

ui/src/components/header/HeaderMenu.js
```diff
@@ -87,68 +92,103 @@ const HeaderMenuUserDiv = styled.div`
...
+        this.props.router.push(route, route);
+    }
+
+    componentDidMount() {
+        this.props.getUserPendingMembers();
+        this.props.getReviewRoles();
+        this.props.getReviewGroups();
+        this.api.getPageFeatureFlag('roleGroupReview').then((data) => {
+            this.setState({
+                roleGroupReviewFeatureFlag: data['roleGroupReviewFeatureFlag'],
```
chandrasekhar1996, line 124: should we add this roleGroupReviewFeatureFlag to default-config.js

ui/src/components/header/HeaderMenu.js
```diff
@@ -87,68 +92,103 @@ const HeaderMenuUserDiv = styled.div`
...
-                </Menu>
-            </HeaderMenuUserDiv>
-            {props.headerDetails && props.headerDetails.userId && (
+    }
+
+    render() {
+        let props = this.props;
+        let reviewNotificationRedDot = '';
+        let icon = 'notification';
+        let clickIcon = this.onClickFunction.bind(
```
chandrasekhar1996, line 133: future note: when you add the new tabs for role/group review, update this to go to the tab which has pending approval work.

chandrasekhar1996 (changes requested):

ui/src/components/header/HeaderMenu.js
jimmytsang: oh interesting good suggestion

ui/src/components/header/HeaderMenu.js
jimmytsang: done.

ui/src/components/header/HeaderMenu.js
jimmytsang: done

chandrasekhar1996 (approved):

PR #2446, opened 2023-12-01
add role and group review tabs

jimmytsang:
This PR introduces the role and group review tabs which show all roles and groups that are pending the user's review.

### DIFF
--- a/ui/src/__tests__/components/review/__snapshots__/ReviewTable.test.js.snap
+++ b/ui/src/__tests__/components/review/__snapshots__/ReviewTable.test.js.snap
@@ -407,13 +407,13 @@ exports[`ReviewTable should render review table 1`] = `
           >
             <input
               checked=""
-              id="radiobutton-undefinedrole-review-user1-extend"
-              name="undefinedrole-review-user1"
+              id="radiobutton-undefinedrole-review-roleNameuser1-extend"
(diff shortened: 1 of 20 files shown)


chandrasekhar1996:
can the card be centered or the width expanded to take up more space in the page

chandrasekhar1996 (changes requested):
Also, I think once the user clicks on the notification icon and reaches the workflow page, the user may have actions required in 2 tabs which we are not communicating. So User will have to click through all 4 tabs to check for pending actions. Instead if we can highlight the tabs in which action is pending, it would be useful.

jimmytsang:
> Also, I think once the user clicks on the notification icon and reaches the workflow page, the user may have actions required in 2 tabs which we are not communicating. So User will have to click through all 4 tabs to check for pending actions. Instead if we can highlight the tabs in which action is pending, it would be useful.

I fully agree with you that would make the user's life a bit easier. If possible I'd like to address this as a follow up item after I get the feature working.

MartinTrojans (approved):

jimmytsang:
Previous PRs have implemented the Feature flag that determines whether or not the role review and group review tab shows up or not thus this PR does not need to feature flag protect any of the code changed.

chandrasekhar1996 (approved):

PR #2463, opened 2023-12-19
fix functional tests pt1

jimmytsang:
Problem: The `got` package, which is a dependency of saucelabs, switched to ES modules in version 12.0.0. This is causing the error.
```
00:00:03 var _got = _interopRequireDefault(require("got"));
00:00:03                                   ^
00:00:03 
00:00:03 Error [ERR_REQUIRE_ESM]: require() of ES Module..
```

Solution: Downgrade the got package to a version that uses CommonJS modules.

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -140,9 +140,5 @@
         "wdio-chromedriver-service": "8.0.1",
         "webdriver-manager": "12.1.8",
         "webdriverio": "8.0.13"
-    },
-    "overrides": {
-        "got": "^11.8.2",
-        "qs": "^6.10.3"
     }
 }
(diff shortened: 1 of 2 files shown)


ui/package.json
```diff
@@ -140,9 +140,5 @@
         "wdio-chromedriver-service": "8.0.1",
         "webdriver-manager": "12.1.8",
         "webdriverio": "8.0.13"
-    },
-    "overrides": {
-        "got": "^12.1.0",
-        "qs": "^6.10.3"
```
jimmytsang, line 146: Ideally we shouldn't have any package overrides which interrupt with the nested package dependencies. I've tested on my local with these packages removed, everything works as expected. Will double check and test on our other environments once merged

chandrasekhar1996 (approved):

PR #2471, opened 2024-01-03
add scope field to microsegmentation table

jimmytsang:
This PR adds the following 

- add scope field to microsegmentation table

image

- add description field to role header

image

### DIFF
--- a/ui/src/components/microsegmentation/RuleRow.js
+++ b/ui/src/components/microsegmentation/RuleRow.js
@@ -30,6 +30,7 @@ import { connect } from 'react-redux';
 import { deleteTransportRule } from '../../redux/thunks/microsegmentation';
 import { deleteAssertionCondition } from '../../redux/thunks/policies';
 import { selectDomainAuditEnabled } from '../../redux/selectors/domainData';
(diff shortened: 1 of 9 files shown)


ui/src/components/microsegmentation/RuleRow.js
```diff
@@ -317,6 +319,13 @@ export class RuleRow extends React.Component {
             ''
         );
 
+        let scope = '';
+        if (data && data['conditionsList']) {
+            data['conditionsList'].forEach((item) => {
+                scope = this.stringUtils.getScopeString(item);
+            });
+        }
+
```
jimmytsang, line 328: @chandrasekhar1996 , from what I've noticed all the msd rules only have one item in the `data['conditionsList']`. Are there examples such that there are multiple items in the conditionsList?

jimmytsang: discussed with chandu on this and found an example with multiple items, will fix code with next commit.

jimmytsang: fixed.

dantewuuuuu (approved):

PR #2504, opened 2024-02-07
upgrade saucelabs and webdriver pkgs

jimmytsang:


ui/package.json
```diff
@@ -137,8 +137,8 @@
         "react-test-renderer": "18.2.0",
         "sinon": "14.0.0",
         "supertest": "6.2.4",
-        "wdio-chromedriver-service": "8.0.1",
-        "webdriver-manager": "12.1.8",
-        "webdriverio": "8.0.13"
+        "wdio-chromedriver-service": "^8.1.1",
+        "webdriver-manager": "^12.1.9",
+        "webdriverio": "^8.31.0"
```
jimmytsang, line 142: safe to automatically upgrade minor releases as these are devDependencies that we only use when executing functional tests.

abvaidya: we use explicit versions without any markers to have reproducible builds between open source and internal.

chandrasekhar1996 (approved):

Issue #2182, opened 2023-05-23
Cannot use PrincipalGroups in the Athenz Template


hiragi-gkuth:
In the current Athenz implementation, [PrincipalGroups](https://athenz.github.io/athenz/principal_groups/) cannot be added as RoleMember to an Athenz Template's Role. This requires multiple calls to putRole / putPolicy etc. when complex domain definitions are needed. To resolve this, I would like to enable the use of PrincipalGroups in Athenz Templates.

I think the issue is due to specifying the CompoundName type as the request body for the putDomainTemplate API in the RDL as follows:

The putDomainTemplate API accepts a DomainTemplate as the body.
https://github.com/AthenZ/athenz/blob/498e14e0a1d19bcb15b027839f0587cab79a9d68/core/zms/src/main/rdl/Domain.rdli#L232-L236

The value field of TemplateParam is a `CompoundName` .
https://github.com/AthenZ/athenz/blob/498e14e0a1d19bcb15b027839f0587cab79a9d68/core/zms/src/main/rdl/Template.tdl#L35-L43

However, the type of PrincipalGroups is `GroupName` .
https://github.com/AthenZ/athenz/blob/498e14e0a1d19bcb15b027839f0587cab79a9d68/core/zms/src/main/rdl/Names.tdl#L59-L60

I'm planning to fix this by accepting PrincipalGroups in putDomainTemplate and adding them to the Role. Are there any implementation concerns or issues that would conflict with the design philosophy of Athenz with this change?

havetisyan:
You're describing a use case where a role member is a variable defined in a template and you want the user to specify the role member as a group as part of the template arguments?

hiragi-gkuth:
Thank you for your response.

> You're describing a use case where a role member is a variable defined in a template and you want the user to specify the role member as a group as part of the template arguments?

That's true.

havetisyan:
it looks like when we added support for Groups as role members the template rdl was not updated accordingly. What was your proposed fix for the issue?

hiragi-gkuth:
I'm considering modifying the `putDomainTemplate()` to accept not only `CompoundName` but also `GroupName` in the value field of the `TemplateParam` in the `DomainTemplate` type.
[...]

PR #2511, opened 2024-02-16
allow param values to include any string

havetisyan:
# Description

when using templates to create policies, the actions and resources can include any string. having template param value defined as compound value blocks supports for such a use case. Now, we allow any string to be set for the param and then make sure the objects we generate (roles/services/policies) are validated before being updated in the db.

# Contribution Checklist:


havetisyan:
this would address #2182

dantewuuuuu (approved):

PR #2527, opened 2024-02-26
fetch verification key from server via proxy during accesstoken and roletoken verification

Bigdrea6:
# Description
In this pull request, we have enabled the use of a proxy URL when retrieving the verification key during the validation of accesstoken and roletoken.

# Contribution Checklist:

### DIFF
--- a/clients/java/zpe/pom.xml
+++ b/clients/java/zpe/pom.xml
@@ -31,7 +31,7 @@
     <mock.server.version>5.15.0</mock.server.version>
     <commons.io.version>2.15.1</commons.io.version>
     <uberjar.name>benchmarks</uberjar.name>
-    <code.coverage.min>0.8682</code.coverage.min>
+    <code.coverage.min>0.8702</code.coverage.min>
   </properties>
 
   <dependencies>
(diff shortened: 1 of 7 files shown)


clients/java/zpe/pom.xml
```diff
@@ -31,7 +31,7 @@
     <mock.server.version>5.15.0</mock.server.version>
     <commons.io.version>2.15.1</commons.io.version>
     <uberjar.name>benchmarks</uberjar.name>
-    <code.coverage.min>0.8702</code.coverage.min>
+    <code.coverage.min>0.8682</code.coverage.min>
```
havetisyan, line 34: any reason why we're reducing our code coverage? please review your additions to make sure all the new code is included in the tests.
we only allow reducing the code coverage if you're removing functions from the module which doesn't seem to be the case here.

havetisyan (changes requested):

clients/java/zpe/pom.xml
Bigdrea6: We modified the implementation of setAccessTokenSignKeyResolver in AuthZpeClient.java to restore coverage.

Bigdrea6:
@havetisyan 
The reason this screwdriver is failing is not due to our implementation, but rather an issue with screwdriver itself.
Please review.

havetisyan (approved):

PR #2547, opened 2024-03-09
use spiffe namespace of default

havetisyan:
# Description
 when not running within K8S, the spiffe uri is generated using namespace of default. On the server side instead of requiring the client to submit the namespace as part of the instance object, we'll default to value "default" if the field is empty or null.
# Contribution Checklist:


zetaj:
👍

PR #2553, opened 2024-03-13
Adding provider interface for fetching public keys of a service

psasidhar:
# Description
Adding a provider interface for fetching public keys of a service
Extending Go provider interface with GetSuffixes method

# Contribution Checklist:

### DIFF
--- a/libs/go/sia/host/provider/provider.go
+++ b/libs/go/sia/host/provider/provider.go
@@ -61,8 +61,8 @@ type Provider interface {
 	// GetSanIp returns an array of IPs that can be included in San IPs from the list of IPs found on the box
 	GetSanIp(map[string]bool, []net.IP, ip.Opts) []net.IP
 
-	// GetSuffix returns the suffix for the current provider
-	GetSuffix() string
+	// GetSuffixes returns a list of suffixes for the current provider
+	GetSuffixes() []string
 
 	// CloudAttestationData gets the attestation data to prove the identity from metadata of the respective cloud
 	CloudAttestationData(string, string, string) (string, error)
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/key/PubKeysProvider.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/key/PubKeysProvider.java
@@ -1,0 +1,35 @@
+/*
+ * Copyright The Athenz Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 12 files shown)


libs/go/sia/host/provider/provider.go
```diff
@@ -64,6 +64,9 @@ type Provider interface {
 	// GetSuffix returns the suffix for the current provider
 	GetSuffix() string
 
+	// GetSuffixes returns a list of suffixes for the current provider
+	GetSuffixes() []string
```
havetisyan, line 68: why keep the GetSuffix() api? since this is an internal interface for now and only used within the server, we can just update all the provider implementations to use GetSuffixes() instead of maintaining a deprecated api

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/key/PubKeysProvider.java
```diff
@@ -0,0 +1,19 @@
+package com.yahoo.athenz.common.server.key;
```
havetisyan, line 1: need the copyright header block

libs/java/server_common/src/test/java/com/yahoo/athenz/common/server/key/PubKeysProviderTest.java
```diff
@@ -0,0 +1,28 @@
+package com.yahoo.athenz.common.server.key;
```
havetisyan, line 1: need the copyright header block

servers/zts/src/main/java/com/yahoo/athenz/zts/store/DataStore.java
```diff
@@ -1936,10 +1939,28 @@ public List<Role> getRolesByDomain(String domain) {
...
+    @Override
+    public List<PublicKeyEntry> getPubKeysByService(String domain, String service) {
+        DomainData domainData = getDomainData(domain);
+        if (domainData == null) {
+            LOGGER.error("domainData can not be null, domain: {}, service: {}", domain, service);
+            return new ArrayList<>();
+        }
+
+        for (ServiceIdentity serviceIdentity: domainData.getServices()) {
+            if (serviceIdentity.getName().equalsIgnoreCase(domain + "." + service)) {
```
havetisyan, line 1951: we should move the "domain + "." + service" outside of the for block so we're not executing for every call in the iteration since the value does not change

havetisyan (changes requested):

havetisyan (approved):

PR #2556, opened 2024-03-14
Fixed ZMSUtils to correctly determine PrincipalType

hiragi-gkuth:
# Description
Background is similar to https://github.com/AthenZ/athenz/pull/2532.
When home domain and user domain are identical, the `ZMSUtils.principalType()` returns `Principal.Type.USER` regardless a provided principal string was `Principal.Type.GROUP`.

e.g.
home domain = user domain = `personal`

ZMSUtils.principalType("personal.hiragi-gkuth") // returns `USER`
ZMSUtils.principalType("personal.hiragi-gkuth.service") // returns `USER` but `SERVICE` is expected
ZMSUtils.principalType("personal.hiragi-gkuth:group.mygroup") // returns `USER` but `GROUP` is expected
ZMSUtils.principalType("personal.hiragi-gkuth.subdomain:group.subgroup") // returns `USER` but `GROUP` is expected

So we add more conditions.

# Contribution Checklist:
  - There might be unintended consequences that I haven't anticipated.


hiragi-gkuth:
I'll fix failed test.

hiragi-gkuth:
I have a few questions regarding [this test code](https://github.com/AthenZ/athenz/blob/master/servers/zms/src/test/java/com/yahoo/athenz/zms/ZMSDeleteUserTest.java).

After adding some conditions to determine the principal type,[these lines](https://github.com/AthenZ/athenz/blob/f4e68c58edc74ee0d2d3f16308831cccb18e6c25/servers/zms/src/test/java/com/yahoo/athenz/zms/ZMSDeleteUserTest.java#L97-L98) started to fail. I had thought that `user.jack.sub1.api` was not a valid principal name for the Athenz Service (but `home.jack.sub1.api` is valid I think). Is this correct? And what is the purpose of adding user.jack.sub1.api to the Athenz Group in this test case?

havetisyan:
If you’re using user as the prefix for your personal domains then yes, user.jack.sub1.api is a valid service principal.

havetisyan:
We don’t recommend using the same prefix for users and home domains - for backwards compatibility they’re the same value but from experience it’s confusing so we have switched to using user to identify users and home to identify personal domains. So when we get user.joe it’s always a user identity and when we get home.joe then it’s user’s personal domain.

havetisyan:
add the following two lines on line 84 in the test case and that should fix the failure. 

`        ServiceIdentity service2 = new ServiceIdentity().setName(ResourceUtils.serviceResourceName("user.jack.sub1", "api"));
        zmsImpl.putServiceIdentity(ctx, "user.jack.sub1", "api", auditRef, false, service2);

        ServiceIdentity service3 = new ServiceIdentity().setName(ResourceUtils.serviceResourceName("user.jack.sub1", "service"));
        zmsImpl.putServiceIdentity(ctx, "user.jack.sub1", "service", auditRef, false, service3);`

servers/zms/src/main/java/com/yahoo/athenz/zms/utils/ZMSUtils.java
```diff
@@ -271,20 +271,27 @@ public static Principal.Type principalType(final String memberName, final String
...
                     return true;
                 }
             }
         }
 
         return false;
     }
+    
+    public static boolean isHeadlessUserDomainPrincipal(final String memberName, final String headlessUserDomainPrefix) {
+        if (memberName.startsWith(headlessUserDomainPrefix) && StringUtils.countMatches(memberName, '.') == 1) {
```
havetisyan, line 290: we should simplify this with just a single return statement:

```
return memberName.startsWith(headlessUserDomainPrefix) && StringUtils.countMatches(memberName, '.') == 1;
```

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/utils/ZMSUtils.java
hiragi-gkuth: fixed!

hiragi-gkuth:
> add the following two blocks on line 84 in the test case and that should fix the failure.

Thank you. I realized that someone had just not created the service because the tests were being conducted with the user and home domains being the same.

havetisyan (approved):

PR #2589, opened 2024-04-18
allow attribute validator for K8SProvider issuer validation

abvaidya:
# Description

# Contribution Checklist:


libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/DefaultAWSElasticKubernetesServiceValidator.java
```diff
@@ -96,10 +135,20 @@ public String validateIssuer(InstanceConfirmation confirmation, IdTokenAttestati
...
-        if (!verifyIssuerPresenceInDomainAWSAccount(issuer,
-                confirmation.getAttributes().get(ZTS_INSTANCE_AWS_ACCOUNT))) {
-            return null;
+        if (useIamRoleForIssuerValidation()) {
+            if (!verifyIssuerPresenceInDomainAWSAccount(issuer,
+                    confirmation.getAttributes().get(ZTS_INSTANCE_AWS_ACCOUNT))) {
+                return null;
+            }
+        } else {
+            confirmation.getAttributes().put(ZTS_INSTANCE_UNATTESTED_ISSUER, issuer);
```
havetisyan, line 144: if the attribute is only intended for the attrValidator then the usage would be more clear if we update the block to have:

```
if (attrValidator != null) {

    confirmation.getAttributes().put(ZTS_INSTANCE_UNATTESTED_ISSUER, issuer);

    if (!attrValidator.confirm(confirmation)) {

      return null;

    }

}
```

havetisyan (approved):

PR #2632, opened 2024-05-30
server k8s common module

abvaidya:
# Description

# Contribution Checklist:


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/PrivateKeyStore.java
```diff
@@ -68,15 +68,31 @@ default String getApplicationSecret(String appName, String keyName) {
...
+     * The application name specifies what component is this secret for;
+     * for example, jdbc for accessing the secret for the jdbc user.
+     * The default implementation assumes the key name is the secret.
+     * @param appName application name for the secret
+     * @param keygroupName key group name for the secret
+     * @param keyName name of the secret
+     * @return secret for the given key and application as char[]
+     */
+    default char[] getSecret(String appName, String keygroupName, String keyName) {
+        return keyName.toCharArray();
```
havetisyan, line 96: this should be set to getSecret(appName, null, keyName); to main backward compatibility and not break anyone before they get a chance to update their code.

havetisyan (changes requested):
Also I don't see the changes in the server side utilizing the new api

havetisyan (approved):

Issue #2638, opened 2024-06-11
How should the Athenz jwk config be distributed?


ysknkd:
The AuthZpeClient supports both the "traditional athenz.conf" and the "jwk format of athenz.conf". While the traditional athenz.conf can be obtained using the athenz-conf CLI, there seems to be no CLI available for obtaining the jwk format of athenz.conf.

In light of this, I have two questions:

* How are Athenz users expected to obtain the jwk format of athenz.conf?
* We are considering implementing `HttpPublicKeyStore.java` to retrieve jwk from ZTS, but we are concerned that this implementation is not currently provided. Is there any reason why either the traditional athenz.conf or the jwk format of athenz.conf must be placed as a file before the application starts?

ysknkd:
Is it assumed that the jwt format of athenz conf will be distributed by sia?
https://github.com/AthenZ/athenz/blob/f9f1240a770df867206aa94311e091d7ac49115b/libs/go/sia/agent/agent.go#L342-L347

ysknkd:
I understand that sia acquires the information via [postInstanceRegisterInformation](https://github.com/AthenZ/athenz/blob/f9f1240a770df867206aa94311e091d7ac49115b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java#L3899) and [postInstanceRefreshInformation](https://github.com/AthenZ/athenz/blob/f9f1240a770df867206aa94311e091d7ac49115b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java#L4227).

Would it be feasible to expose the jwkConfig utilized by the two methods mentioned above as an API of ZTS? https://github.com/AthenZ/athenz/blob/f9f1240a770df867206aa94311e091d7ac49115b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java#L411-L414

havetisyan:
I'm slowly moving away from athenz.conf since that was a proprietary format and instead I prefer to use the jwk format as defined in rfc7517.

the initial implementation relies on SIA to automatically fetch the conf file and keep it updated periodically. This eliminates the need for the client to worry about how to refresh and whenever it sees an unknown key, all it needs to do is to reload the latest file from disk which should be updated.

So I would strongly recommend if you're implement your own SIA, then make this part of the agent so it automatically gets updated daily. The server logic only returns the data if there are changes so you don't waste network resources by sending the data every time.

[...]

PR #2642, opened 2024-06-25
/oauth2/keys Specify the service to obtain the public key

TakuyaMatsu:
# Description
#2638
`/oauth2/keys` query parameter to specify a service so that the public key of the specified service can be obtained.
e.g. `/oauth2/keys?rfc=true&service=zms`

# Contribution Checklist:

### DIFF
--- a/servers/zts/pom.xml
+++ b/servers/zts/pom.xml
@@ -29,7 +29,7 @@
   <packaging>war</packaging>
 
   <properties>
-    <code.coverage.min>0.9954</code.coverage.min>
+    <code.coverage.min>0.9952</code.coverage.min>
   </properties>
 
   <dependencyManagement>
--- a/servers/zts/pom.xml
+++ b/servers/zts/pom.xml
@@ -29,7 +29,7 @@
   <packaging>war</packaging>
 
   <properties>
-    <code.coverage.min>0.9952</code.coverage.min>
+    <code.coverage.min>0.9954</code.coverage.min>
   </properties>
 
   <dependencyManagement>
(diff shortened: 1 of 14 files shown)


servers/zts/pom.xml
```diff
@@ -29,7 +29,7 @@
   <packaging>war</packaging>
 
   <properties>
-    <code.coverage.min>0.9954</code.coverage.min>
+    <code.coverage.min>0.9952</code.coverage.min>
```
havetisyan, line 32: we do not allow reducing the code coverage. please review the jacoco output and add test cases to have full 100% coverage for any new code.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -5020,13 +5020,18 @@ public ExternalCredentialsResponse postExternalCredentialsRequest(ResourceContex
...
-    public JWKList getJWKList(ResourceContext ctx, Boolean rfc) {
+    public JWKList getJWKList(ResourceContext ctx, Boolean rfc, String service) {
 
         final String caller = ctx.getApiName();
         final String principalDomain = logPrincipalAndGetDomain(ctx);
 
         validateOIDCRequest(ctx.request(), principalDomain, caller);
-        return dataStore.getZtsJWKList(rfc);
+        switch (service) {
+          case "zms":
```
havetisyan, line 5030: let's replace "zms" with ServerCommonConsts.ZMS_SERVICE

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
TakuyaMatsu: fixed

servers/zts/pom.xml
TakuyaMatsu: fixed

havetisyan (approved):

PR #2660, opened 2024-07-19
DIscover additional instances to MSD, with dynamic/static workloads

rajeshal:
# Description

# Contribution Checklist:


core/msd/src/main/rdl/Workload.tdl
```diff
@@ -87,4 +87,15 @@ type BulkWorkloadRequest Struct {
 type BulkWorkloadResponse Struct {
     Array<DomainServices> unmodifiedServices;   // list of services grouped by domain, those are not changed since time stamp in matchingTag
     Workloads workloads;                        // matching workloads
+}
+
+// generic instance
+type CompositeInstance Struct {
+    DomainName domainName;                  // name of the domain
+    EntityName serviceName;                 // name of the service
```
havetisyan, line 95: if this only refers to the actual service name (e.g. api) and not the full service name (e.g. athenz.api) since we have a domainName field, then it must be defined as SimpleName.

core/msd/src/test/java/com/yahoo/athenz/msd/CompositeInstanceTest.java
```diff
@@ -0,0 +1,156 @@
...
+
+        assertNotEquals(instance, null);
+        // for code coverage
+        assertFalse(instance.equals("mystring"));
+        assertNotEquals(instance, "mystring");
+
+        assertEquals(instance, instance);
+
+    }
+//
```
havetisyan, line 119: any reason why we have all this commented code?

core/msd/src/main/rdl/Workload.rdli
```diff
@@ -145,4 +145,36 @@ resource BulkWorkloadResponse POST "/workloads" (name=getWorkloadsByDomainAndSer
         ResourceError UNAUTHORIZED;
         ResourceError TOO_MANY_REQUESTS;
     }
-}
\ No newline at end of file
+}
+
+// Api to discover an additional instance which can have static or dynamic or both IPs
+resource Workloads PUT "/domain/{domainName}/service/{serviceName}/workload/discover/instance" (name=postCompositeInstance) {
```
havetisyan, line 151: since this is a PUT API, the name should not refer to it as postCompositeInstance and instead should be called putCompositeInstance

core/msd/src/main/rdl/Workload.rdli
```diff
@@ -145,4 +145,36 @@ resource BulkWorkloadResponse POST "/workloads" (name=getWorkloadsByDomainAndSer
...
+        ResourceError UNAUTHORIZED;
+        ResourceError TOO_MANY_REQUESTS;
+    }
+}
+
+// Api to delete an additional instance which can have static or dynamic or both IPs
+resource Workloads DELETE "/domain/{domainName}/service/{serviceName}/workload/discover/instance" (name=deleteCompositeInstance) {
+    DomainName domainName;      // name of the domain
+    EntityName serviceName;     // name of the service
+    CompositeInstance instance; // Generic instance
```
havetisyan, line 170: I'm not sure if we really want a DELETE api with a body. Even if the server supports it, there is no guarantee if the request going through some other proxy service before hitting the server will maintain the body or just simply drop it. So my recommendation would be to redesign the API to not have any body.

havetisyan (changes requested):

core/msd/src/main/rdl/Workload.tdl
rajeshal: It will have only the service name without domain name, shouldn't it be inconsistent with other endpoints 
For example
Static Workload https://github.com/AthenZ/athenz/blob/master/core/msd/src/main/rdl/Workload.rdli#L75
Dynamic Workload  https://github.com/AthenZ/athenz/blob/master/core/msd/src/main/rdl/Workload.rdli#L43
both accepts service name with out domain name as prefix

core/msd/src/test/java/com/yahoo/athenz/msd/CompositeInstanceTest.java
rajeshal: my bad, forgot to remove that.

core/msd/src/main/rdl/Workload.rdli
```diff
@@ -145,4 +145,36 @@ resource BulkWorkloadResponse POST "/workloads" (name=getWorkloadsByDomainAndSer
...
+        ResourceError BAD_REQUEST;
+        ResourceError NOT_FOUND;
+        ResourceError FORBIDDEN;
+        ResourceError UNAUTHORIZED;
+        ResourceError TOO_MANY_REQUESTS;
+    }
+}
+
+// Api to delete an additional instance which can have static or dynamic or both IPs
+resource Workloads DELETE "/domain/{domainName}/service/{serviceName}/workload/discover/instance/${instance}" (name=deleteCompositeInstance) {
```
havetisyan, line 167: there is an extra $ ${instance}

havetisyan (changes requested):

havetisyan (approved):

PR #2770, opened 2024-10-17
defer accesstoken error logging

abvaidya:
# Description

# Contribution Checklist:


havetisyan:
when we're appending multiple errors, there is no separator between the errors so we're going to get the last and the first words together - we should add some separator type at the end of each string (e.g. ;) to indicate the end of the error message before the next one is appended.

havetisyan (approved):

PR #2772, opened 2024-10-18
Store SIA user agent information in x509 certificate request table

rajeshal:
# Description

# Contribution Checklist:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/cert/impl/JDBCCertRecordStoreConnection.java
```diff
@@ -208,6 +210,7 @@ public boolean updateX509CertRecord(X509CertRecord certRecord) throws ServerReso
             ps.setString(10, certRecord.getProvider());
             ps.setString(11, certRecord.getInstanceId());
             ps.setString(12, certRecord.getService());
+            ps.setString(13, certRecord.getSiaProvider());
```
havetisyan, line 213: the sia provider is not the last component in the UPDATE command. It should be field 10 and then others are pushed back by 1.

can you also please review the test case for this method since this should have been caught by the unit tests.

servers/zts/schema/zts_server.sql
```diff
@@ -35,6 +35,7 @@ CREATE TABLE IF NOT EXISTS `zts_store`.`certificates` (
   `lastNotifiedServer` VARCHAR(512) NULL,
   `expiryTime` DATETIME(3) NULL,
   `hostName` VARCHAR(512) NULL,
+  `siaProvider` VARCHAR(256) NOT NULL,
```
havetisyan, line 38: These are the instructions for the ZMS schema but they apply to ZTS as well:

https://github.com/AthenZ/athenz/blob/master/servers/zms/schema/schema_update.md

havetisyan (changes requested):

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/cert/impl/JDBCCertRecordStoreConnection.java
rajeshal: yeah, good catch.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/util/ServletRequestUtil.java
```diff
@@ -45,5 +46,20 @@ public static String getRemoteAddress(final HttpServletRequest request) {
         }
         return addr;
     }
+
+    /**
+     * Return the SIA user agent name, which is set by sia agent as request header.
+     * @param request http servlet request
+     * @return SIA agent along with version, like 'SIA-FARGATE 1.32.0'
+     **/
+    public static String getSiaAgent(final HttpServletRequest request) {
```
havetisyan, line 55: is there any reason why we want to maintain the full value of the agent including the version number? Why not just extract the value after SIA- and then our provider will be EKS, AWS, Fargate, etc so it's easier to index and just run queries. Otherwise you'll have SIA-EKS 1.52.0, SIA-EKS 1.53.0, etc.

servers/zts/schema/zts_server.sql
```diff
@@ -35,8 +35,9 @@ CREATE TABLE IF NOT EXISTS `zts_store`.`certificates` (
   `lastNotifiedServer` VARCHAR(512) NULL,
   `expiryTime` DATETIME(3) NULL,
   `hostName` VARCHAR(512) NULL,
+  `siaProvider` VARCHAR(256) NULL,
   PRIMARY KEY (`provider`, `instanceId`, `service`),
-  INDEX `idx_hostName` (`hostName` ASC))
+  INDEX `idx_hostName` (`hostName` ASC) VISIBLE)
```
havetisyan, line 40: how did we get the VISIBLE flag here?

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/util/ServletRequestUtil.java
rajeshal: With the provider agent version it will give more flexibility in general to filter based on certain agent version. Since the version is at the end queries can still use fuzzy search in indexes.
Also will be scalable if another type of Service Identity Agent come into picture in addition to SIA

servers/zts/schema/zts_server.sql
rajeshal: It is mandatory for current version of `MySQL Workbench`

rajeshal: So who ever touch a table via `MySQL Workbench`, it change the index with default value `VISIBLE`. 
Since that is the default value, it should be fine as I don't think Athenz uses any indexes not to be used by optimizer.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/util/ServletRequestUtil.java
havetisyan: If that is the case, then the field should be called userAgent and not siaProvider since the user agent is not provider. Not sure exactly what use cases you're referring to regarding the version number - that information is already in the access log file so no need to maintain it here. In fact, for a lot of lambda identities, we're gong to get something like 'Jersey/2.35 (Apache HttpClient 4.5.13)' so that's not useful at all. 

We should keep our data clean and only save what we absolutely need. So we should get the agent, parse and only take agents with the format "SIA-<provider> <some other data>", extract the provider and save that value only. For all other cases, the provider will be empty and that should be excluded from any processing.

libs/java/server_common/src/test/java/com/yahoo/athenz/common/server/cert/X509CertRecordTest.java
```diff
@@ -45,6 +45,7 @@ public void testX509CertRecord() {
         certRecord.setLastNotifiedTime(now);
         certRecord.setExpiryTime(now);
         certRecord.setSvcDataUpdateTime(now);
+        certRecord.setSiaProvider("SIA-EKS 1.52.0");
```
havetisyan, line 48: let's update this test case to use cert.setSiaProvider("EKS") to be consistent with the implementation.

havetisyan (approved):

Issue #2818, opened 2024-12-04
ZTS Docker Image does not bundle the athenz-instance-provider library


ysknkd:
The docker image does not bundle the instance-provider library `athenz-instance-provider` , which causes the instanceZTSProvider to not function.

As far as I have confirmed, it seems that the instance-provider is no longer bundled from version 1.12.x. Instead of the instance-provider, the athenz-zts-java-client is now bundled.

```bash
$ docker run --rm --entrypoint="" athenz/athenz-zts-server:1.11.66 ls -la /opt/athenz/zts/lib/jars | grep athenz
-rw-r--r-- 1 root root   151173 Sep 28 22:05 athenz-auth-core-1.11.66.jar
-rw-r--r-- 1 root root    21232 Sep 28 22:07 athenz-cert-refresher-1.11.66.jar
-rw-r--r-- 1 root root    26681 Sep 28 22:06 athenz-client-common-1.11.66.jar
-rw-r--r-- 1 root root   103002 Sep 28 22:12 athenz-instance-provider-1.11.66.jar
-rw-r--r-- 1 root root    27800 Sep 28 22:20 athenz-jetty-container-1.11.66.jar
-rw-r--r-- 1 root root    74662 Sep 28 22:04 athenz-msd-core-1.11.66.jar
-rw-r--r-- 1 root root   259566 Sep 28 22:10 athenz-server-common-1.11.66.jar
-rw-r--r-- 1 root root   169566 Sep 28 22:02 athenz-zms-core-1.11.66.jar
-rw-r--r-- 1 root root    63772 Sep 28 22:08 athenz-zms-java-client-1.11.66.jar
-rw-r--r-- 1 root root    85595 Sep 28 22:03 athenz-zts-core-1.11.66.jar

$  docker run --rm --entrypoint="" athenz/athenz-zts-server:1.12.2 ls -la /opt/athenz/zts/lib/jars | grep athenz
-rw-r--r-- 1 root root   153565 Oct 18 02:27 athenz-auth-core-1.12.2.jar
-rw-r--r-- 1 root root    21486 Oct 18 02:28 athenz-cert-refresher-1.12.2.jar
-rw-r--r-- 1 root root    27031 Oct 18 02:28 athenz-client-common-1.12.2.jar
-rw-r--r-- 1 root root    27399 Oct 18 02:42 athenz-jetty-container-1.12.2.jar
[...]

PR #2820, opened 2024-12-04
include instance profile module with zts server binary image

havetisyan:
# Description
addresses #2818 
# Contribution Checklist:

### DIFF
--- a/assembly/zts/pom.xml
+++ b/assembly/zts/pom.xml
@@ -38,6 +38,11 @@
       <version>${project.parent.version}</version>
     </dependency>
     <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>athenz-instance-provider</artifactId>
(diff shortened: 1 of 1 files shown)


ysknkd:
@havetisyan This is just out of curiosity, I think the issue is with the configuration when building the docker image. If we fix this, will it also resolve the problem with the docker image?

havetisyan:
@ysknkd Yes, there is nothing wrong with the docker image configuration. If you look at how the docker image is built, it just takes the contents of the zts server tar file that is built as part of the assembly job.

https://github.com/AthenZ/athenz/blob/master/docker/util/athenz-builder/Dockerfile#L35

ysknkd:
@havetisyan I understand now. Thank you for your explanation.

psasidhar (approved):

PR #2832, opened 2024-12-17
ui search capability to include services

ArtjomsPorss:
# Description
UI search capability to include services
- added dropdown to pick Domain or Service search
- added call to new search services endpoint
- added display formatting of found services
- added functional tests
- updated jest snapshots

# Contribution Checklist:

### DIFF
--- a/ui/src/components/constants/constants.js
+++ b/ui/src/components/constants/constants.js
@@ -281,3 +281,6 @@ export const MEMBER_AUTHORITY_SYSTEM_SUSPENDED = 2;
 export const MEMBER_ATHENZ_SYSTEM_DISABLED = 4;
 
 export const PERCENTAGE_OF_DAYS_TILL_NEXT_REVIEW = 0.2;
+
+export const SEARCH_MAX_100_RESULTS_CAN_BE_DISPLAYED =
+    'Maximum 100 results can be searched for at a time. Specify more characters in the search criteria to get better results.';
--- a/ui/src/pages/search/[type]/[searchterm].js
+++ b/ui/src/pages/search/[type]/[searchterm].js
@@ -36,6 +36,9 @@ import { connect } from 'react-redux';
 import createCache from '@emotion/cache';
 import { CacheProvider } from '@emotion/react';
 import RequestUtils from '../../../components/utils/RequestUtils';
+import { searchServices } from '../../../redux/thunks/services';
+import Menu from '../../../components/denali/Menu/Menu';
+import { SEARCH_MAX_100_RESULTS_CAN_BE_DISPLAYED } from '../../../components/constants/constants';
 
 const AppContainerDiv = styled.div`
     align-items: stretch;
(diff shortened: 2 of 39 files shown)


ui/src/server/handlers/api.js
```diff
@@ -1798,6 +1798,32 @@ Fetchr.registerService({
     },
 });
 
+Fetchr.registerService({
+    name: 'search-services',
+    read(req, resource, params, config, callback) {
+        req.clients.zms.searchServiceIdentities(params, function (err, data) {
+            if (!err && Array.isArray(data.list)) {
+                data.list.sort((a, b) => {
+                    debug(a.name, b.name);
```
chandrasekhar1996, line 1807: remove in subsequent PR

chandrasekhar1996 (changes requested):
1) We should indicate to the the user that the search results being shown is partial. (The search results is partial when [serviceMatchCount](https://github.com/AthenZ/athenz/blob/c2c94e09c7e5b297b651e6e818600080ef000a01/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java#L8318) field is set.) One way to do that can be to show it in Search results header below
Screenshot 2025-01-13 at 6 57 55 PM

2) The backend API has the capability to filter results based on domain name as well. Can you confirm if this story/PR was also meant to add support for that in UI? IIRC users can give "athenz:zms" as search term then UI can make search call with domain name set to athenz and service to zms

ui/src/server/handlers/api.js
ArtjomsPorss: regarding point 2 - services search accepts partial name of the service, without the domain name

ArtjomsPorss: will make change regarding point 1

chandrasekhar1996: for point 2, the scenario I was thinking of is if searching for a service returns more than 100(default limit) results then user will never be able to use UI for search if the result is after top 100 even though the user may know the domain name which will help further narrow it down.

ArtjomsPorss: added 
image

ui/src/pages/search/[type]/[searchterm].js
```diff
@@ -174,15 +279,65 @@ class PageSearchDetails extends React.Component {
...
+                            {this.state.services?.serviceMatchCount > 0 && (
+                                <Menu
+                                    placement='bottom-start'
+                                    trigger={
+                                        <span>
+                                            <Icon
+                                                icon={'help-circle'}
+                                                isLink
+                                                size={'15px'}
+                                                color={colors.graphYellow}
```
Comment (ArtjomsPorss) [inline ui/src/pages/search/[type]/[searchterm].js:317]: questionmark in yellow colour

ui/src/components/constants/constants.js
```diff
@@ -281,3 +281,6 @@ export const MEMBER_AUTHORITY_SYSTEM_SUSPENDED = 2;
 export const MEMBER_ATHENZ_SYSTEM_DISABLED = 4;
 
 export const PERCENTAGE_OF_DAYS_TILL_NEXT_REVIEW = 0.2;
+
+export const SEARCH_MAX_100_RESULTS_CAN_BE_DISPLAYED =
+    'Maximum 100 results can be searched for at a time. Specify more characters in the search criteria to get better results.';
```
ArtjomsPorss, line 286: updated wording for question mark popup

chandrasekhar1996 (approved):

PR #2835, opened 2024-12-19
skip calling notify method if there are no recipients

havetisyan:
# Description

# Contribution Checklist:

### DIFF
--- a/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
+++ b/libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
@@ -97,7 +97,16 @@ public class EmailNotificationService implements NotificationService {
         final String subject = notificationAsEmail.getSubject();
         final String body = notificationAsEmail.getBody();
         Set<String> recipients = notificationAsEmail.getFullyQualifiedRecipientsEmail();
-            if (sendEmail(recipients, subject, body)) {
+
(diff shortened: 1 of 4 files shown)


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
```diff
@@ -97,7 +97,16 @@ public boolean notify(Notification notification) {
         final String subject = notificationAsEmail.getSubject();
         final String body = notificationAsEmail.getBody();
         Set<String> recipients = notificationAsEmail.getFullyQualifiedRecipientsEmail();
-            if (sendEmail(recipients, subject, body)) {
+
+        // if our list of recipients is empty then we have nothing to do,
+        // but we want to log it for debugging purposes
+
+        if (recipients.isEmpty()) {
+            LOGGER.error("No recipients specified in the notification. Subject={}", subject);
```
MartinTrojans, line 105: Should this be warn instead of error?

MartinTrojans (approved):

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/notification/impl/EmailNotificationService.java
havetisyan: we should treat it as an error since we're trying to send a notification but we don't have recipients.

PR #2840, opened 2024-12-26
Allow setting of server default values for member expiry days

TakuyaMatsu:
# Description
This PR adds parameters to configure the maximum expiry duration for Members in Athenz ZMS.
The maximum expiry duration can be set individually for Users, Services, and Groups using the following properties:
* `athenz.zms.default_user_expiry_days` : Applies the maximum expiry duration for all user members.
* `athenz.zms.default_service_expiry_days` : Applies the maximum expiry duration for all service members.
* `athenz.zms.default_group_expiry_days` : Applies the maximum expiry duration for all group members.

The default value is 0, which means no expiry is enforced unless explicitly configured.

# Contribution Checklist:

### DIFF
--- a/servers/zms/conf/zms.properties
+++ b/servers/zms/conf/zms.properties
@@ -577,3 +577,10 @@ athenz.zms.no_auth_uri_list=/zms/v1/schema
 # of results returned to the specified value. The default value is 100. This prevents
 # the server from returning a large number of results when the search criteria is too broad.
 #athenz.zms.search_service_limit=100
+
+# This property specifies the maximum expiry duration in days for user/service/group.
+# The value must be an integer, and the default value is 0.
+# If set to 0, it indicates that there is no expiry limit.
+#athenz.zms.default_user_expiry_days=0
+#athenz.zms.default_service_expiry_days=0
+#athenz.zms.default_group_expiry_days=0
--- a/servers/zms/conf/zms.properties
+++ b/servers/zms/conf/zms.properties
@@ -581,6 +581,6 @@ athenz.zms.no_auth_uri_list=/zms/v1/schema
 # This property specifies the maximum expiry duration in days for user/service/group.
 # The value must be an integer, and the default value is 0.
 # If set to 0, it indicates that there is no expiry limit.
-#athenz.zms.default_user_expiry_days=0
-#athenz.zms.default_service_expiry_days=0
-#athenz.zms.default_group_expiry_days=0
+#athenz.zms.default_max_user_expiry_days=0
+#athenz.zms.default_max_service_expiry_days=0
+#athenz.zms.default_max_group_expiry_days=0
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -4820,6 +4820,9 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
     }
 
     Timestamp getMemberDueDate(long cfgDueDateMillis, Timestamp memberDueDate) {
+        if (cfgDueDateMillis == 0) {
+            return memberDueDate;
+        }
         if (memberDueDate == null) {
             return Timestamp.fromMillis(cfgDueDateMillis);
         } else if (memberDueDate.millis() > cfgDueDateMillis) {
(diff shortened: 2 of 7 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/config/MemberDueDays.java
```diff
@@ -59,23 +64,23 @@ public MemberDueDays(Domain domain, Role role, Type type) {
...
     }
 
     public MemberDueDays(Domain domain, Group group) {
 
         // for groups we only have user and service members
         // groups cannot include other groups
 
-        Integer domainUserDays = domain.getMemberExpiryDays();
-        Integer domainServiceDays = domain.getServiceExpiryDays();
+        Integer domainUserDays = null;
```
havetisyan, line 77: Why did we remove the domainUserDays and domainServiceDays value?

TakuyaMatsu: Sorry, this is a mistake. It has been corrected. I have restored it.

servers/zms/conf/zms.properties
```diff
@@ -577,3 +577,10 @@ athenz.zms.no_auth_uri_list=/zms/v1/schema
 # of results returned to the specified value. The default value is 100. This prevents
 # the server from returning a large number of results when the search criteria is too broad.
 #athenz.zms.search_service_limit=100
+
+# This property specifies the maximum expiry duration in days for user/service/group.
+# The value must be an integer, and the default value is 0.
+# If set to 0, it indicates that there is no expiry limit.
+#athenz.zms.default_user_expiry_days=0
```
havetisyan, line 584: let's rename these to be athenz.zms.max_default_... (add the max_ component). default value indicates that we'll set this if no value is specified but here we're doing 2 things: first if no value is set, we default to this value and second, if a value is specified then we do not allow the value to be bigger than this (which is max part and not necessarily default).

servers/zms/src/main/java/com/yahoo/athenz/zms/config/MemberDueDays.java
```diff
@@ -19,9 +19,14 @@
 import com.yahoo.athenz.zms.Group;
 import com.yahoo.athenz.zms.Role;
 import com.yahoo.athenz.zms.utils.ZMSUtils;
+import com.yahoo.athenz.zms.ZMSConsts;
 
 public class MemberDueDays {
 
+    private static final int DEFAULT_USER_EXPIRY = Integer.parseInt(System.getProperty(ZMSConsts.ZMS_PROP_DEFAULT_USER_EXPIRY, "0"));
```
havetisyan, line 26: same here. let's name these DEFAULT_MAX_USER_EXPIRY, etc to correctly indicate the max aspect of the value.

servers/zms/src/main/java/com/yahoo/athenz/zms/utils/ZMSUtils.java
```diff
@@ -449,7 +449,7 @@ public static boolean metaValueChanged(Object domainValue, Object metaValue) {
         return metaValue != null && !metaValue.equals(domainValue);
     }
 
-    public static long configuredDueDateMillis(Integer domainDueDateDays, Integer roleDueDateDays) {
+    public static long configuredDueDateMillis(Integer serverDefaultDueDateDays, Integer domainDueDateDays, Integer roleDueDateDays) {
```
havetisyan, line 452: why do we have serverDefaultDueDateDays set as Integer. It's int so we should define the function as such and it also eliminates the boxing and unnecessary null check later.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4820,6 +4820,9 @@ public DomainRoleMembers getOverdueReview(ResourceContext ctx, String domainName
     }
 
     Timestamp getMemberDueDate(long cfgDueDateMillis, Timestamp memberDueDate) {
```
havetisyan, line 4822: please update the testGetMemberDueDate unit test case to include the cfgDueDateMillis == 0 case.

havetisyan (changes requested):

TakuyaMatsu:
thank you. Addressed your feedback.

havetisyan (approved):

havetisyan:
Thanks @TakuyaMatsu

PR #2849, opened 2025-01-07
ui - improve self review member approval

ArtjomsPorss:
# Description
use member comment to auto populate justification textarea when approving self review roles and groups

# Contribution Checklist:


chandrasekhar1996 (changes requested):
adding comment from offline conversation:

1) instead of using getRoles and getGroups to get details of all roles/groups in a domain, it would be better to do a getRole/Group call for specific collection which have a pending member

2) It will also be useful to check if the role/group data is already present in store, if not then make API call to get role/group and then store it in redux store.

ArtjomsPorss:
> adding comment from offline conversation:
> 
> 1. instead of using getRoles and getGroups to get details of all roles/groups in a domain, it would be better to do a getRole/Group call for specific collection which have a pending member
> 2. It will also be useful to check if the role/group data is already present in store, if not then make API call to get role/group and then store it in redux store.

will make changes to cover both points

ArtjomsPorss:
@chandrasekhar1996 
https://github.com/AthenZ/athenz/pull/2849/files#diff-cb36fec9399cc92922646c454ed2ea4e6def1c54aff8931657551510f2b1a5eeR95-R102
double checked, this part handles point 2 - checks if pending members are in redux storage and not expired (5min), then returns them. 
if expired or not present -> get from backend, enrich with selfServe and justification, store in redux storage.

ui/src/redux/thunks/user.js
```diff
@@ -34,6 +38,93 @@ import {
...
+    state,
+    domainName,
+    groupName
+) => {
+    let group = selectPendingMemberGroup(state, domainName, groupName);
+    if (!group || isExpired(group.expiry)) {
+        group = await API().getGroup(domainName, groupName, false, true);
+        group.expiry = getExpiryTime();
+        dispatch(storePendingGroup(group, domainName, groupName));
+    }
```
ArtjomsPorss, line 64: @chandrasekhar1996 this part covers selecting role/group data from state, if not present or expired - get role/group and store in state for later

ui/src/redux/thunks/user.js
```diff
@@ -34,6 +38,93 @@ import {
...
+    roleOrGroupSet.forEach((entity) => {
+        let [domain, role] = entity.split(':');
+        if (category === ROLE) {
+            promises.push(getPendingMemberRole(dispatch, state, domain, role));
+        } else if (category === GROUP) {
+            promises.push(getPendingMemberGroup(dispatch, state, domain, role));
+        }
+    });
+
+    let data = await Promise.all(promises);
```
ArtjomsPorss, line 98: getting unique roles/groups for selfReview check is done in parallel

chandrasekhar1996 (approved):

PR #2857, opened 2025-01-16
Resolved the issue where a NullPointer occurs when the admin role is a trust role

TakuyaMatsu:
# Description
It might be a rare case, but when the trust of the admin role is set, the admin role member cannot be retrieved, resulting in a NullPointerException. 
I have modified it so that member checks are not performed when it is a role with trust set.

# Contribution Checklist:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -2233,8 +2233,9 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         if (adminRole == null) {
             throw ZMSUtils.notFoundError("Invalid domain name specified", caller);
         }
-        if (StringUtils.isEmpty(adminRole.trust)) {
-            List<RoleMember> members = adminRole.getRoleMembers();
+        List<RoleMember> members = adminRole.getRoleMembers();
+        // If adminRole is a Trust Role, the member will be null.
+        if (members != null) {
             if (members.size() == 1 && members.get(0).getMemberName().equals(memberName)) {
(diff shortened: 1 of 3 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2233,9 +2233,11 @@ public void deleteDomainRoleMember(ResourceContext ctx, String domainName, Strin
         if (adminRole == null) {
             throw ZMSUtils.notFoundError("Invalid domain name specified", caller);
         }
-        List<RoleMember> members = adminRole.getRoleMembers();
-        if (members.size() == 1 && members.get(0).getMemberName().equals(memberName)) {
-            throw ZMSUtils.forbiddenError("deleteDomainRoleMember: Cannot delete last member of 'admin' role", caller);
+        if (StringUtils.isEmpty(adminRole.trust)) {
```
havetisyan, line 2236: 1. Rather than checking the trust attribute, why not just add a members != null condition to the if line. It will all other edge cases when somehow we get a null for the member list.

2. more importantly, we also have the same issue in the deleteMembership API which is probably more commonly used than deleteDomainRoleMember api. So can you please apply the same fix in that function as well.

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
TakuyaMatsu: I've made the changes. Please check

havetisyan: Thanks @TakuyaMatsu

havetisyan (approved):

PR #2858, opened 2025-01-17
Encode the Issuer using the same encoding type as the CA Subject

TakuyaMatsu:
# Description
As stated in [RFC 4630](https://datatracker.ietf.org/doc/html/rfc4630), "Since name comparison assumes that attribute values encoded in different types (e.g., PrintableString and UTF8String) are assumed to represent different strings, any name components that appear in both the subject field and the issuer field SHOULD use the same encoding throughout the certification path."

In the previous [PR](https://github.com/AthenZ/athenz/pull/2495), the method was modified to use DERUTF8String, which may cause issues in languages like Golang.

Therefore, I have made changes to ensure that the Issuer string is created using the same encoding type as the Subject.

# Contribution Checklist:


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
```diff
@@ -1400,8 +1402,13 @@ public static X500Name utf8DEREncodedIssuer(final String issuer) {
     public static X509Certificate generateX509Certificate(PKCS10CertificationRequest certReq,
             PrivateKey caPrivateKey, X509Certificate caCertificate, int validityTimeout,
             boolean basicConstraints) {
-
-        X500Name issuer = utf8DEREncodedIssuer(caCertificate.getSubjectX500Principal().getName());
+        X500Name issuer;
+        try {
+            issuer = new JcaX509CertificateHolder(caCertificate).getSubject();
```
havetisyan, line 1407: @TakuyaMatsu what was the reason for this change? Were there any interoperability issues? If you refer to the comments in the utf8DEREncodedIssuer method - we initially did that to work with openssl 1.x. Were you seeing any issues with newer openssl or something else?

ysknkd: The Subject of the CA we are using is encoded entirely in PrintableString. Therefore, if anything other than the country code is encoded in UTF8String, Go lang fails to verify the trust chain. As you know, Go lang compares EE/Issuer and CA/Subject in binary form.

We believe this PR will ensure that the encoding method of the CA/Subject matches that of the EE/Issuer. Do you have any concerns or points of interest regarding this?

havetisyan: No, I don't have any concerns about the change. Just want to understand the use case where this was causing an issue. So was the CA where the entire subject is PrintableString - was this generated by openssl or some other CA?

ysknkd: It seems that when generating a CSR using Java's keytool, the encoding type becomes PrintableString.

havetisyan (approved):

Issue #2811, opened 2024-11-27
Secure assertion deletion using Athenz Templates


ysknkd:
Using the Athenz Template allows for more secure application operations because it enables managing domain information without granting applications the authority to update policies or assertions.

However, there is a bit of a challenge when we want to delete certain assertions after applying a Template. Even if we remove the assertions we want to delete from the Template and provide the Template again, `addSolutionTemplate()` executes `processPolicy()` with `ignoreDelete = true`, so the existing assertions are not deleted.

https://github.com/AthenZ/athenz/blob/e3c60afb076adb37445c19efddfae0de3090fedc/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java#L5140-L5141
https://github.com/AthenZ/athenz/blob/e3c60afb076adb37445c19efddfae0de3090fedc/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java#L567-L585

As a result, we have to grant the application the authority to update assertions.

Is there a way to delete certain assertions without granting the application the authority to update assertions? Alternatively, is there a method using the Athenz Template to add a new version of the policy and activate it?

havetisyan:
It's kind of dangerous for templates to remove assertions since this could lead to unintended deletion of assertions resulting in outages. For example, you might have a user who has applied a template and then went ahead and added more assertions to that policy. If you apply an update, that will delete all those assertions that were not part of the template, we'll be deleting those assertions.

For this kind of scenario to work, the server needs to keep track of internally what assertions it created as part of the template (something that the user cannot change or set) and then it can delete those assertions only. It would be something similar to the resource ownership idea but not allow the user to set or change it and it has to be set the assertion level which we currently do not support.

We already have a separate delete assertion api - so we can easily update the authorization requirements for this api to have a unique authz check (e.g. similar to membership changes). So you can grant someone to have the capability to delete assertion but not update any. This would be pretty straightforward to introduce without breaking anyone.

As far multiple policy version - currently the templates do not have support for it. That seems somewhat advance use case for templates to start creating multiple policy versions.

ysknkd:
@havetisyan Thank you for your response.

I have understood that due to the risks associated with deleting assertions via templates, there is currently no feature for assertion deletion through templates.

[...]

PR #2902, opened 2025-03-07
Enable deletion of Assertion with delete {domain}:assertion.{assertionId} permission

TakuyaMatsu:
# Description
#2811 
I have added a permission that allows only the deletion of a specific assertion. Deleting an assertion can be performed by anyone who has either the “update: {domainName}:policy.{policyName}” or “delete: {domainName}:assertion.{assertionId}” permission.

# Contribution Checklist:


core/zms/src/main/rdl/Policy.rdli
```diff
@@ -159,7 +159,7 @@ resource Assertion DELETE "/domain/{domainName}/policy/{policyName}/assertion/{a
     Int64 assertionId; //assertion id
     String auditRef (header="Y-Audit-Ref"); //Audit param required(not empty) if domain auditEnabled is true.
     String resourceOwner (header="Athenz-Resource-Owner"); //Resource owner for the request
-    authorize ("update", "{domainName}:policy.{policyName}");
+    authenticate;
```
havetisyan, line 162: please update the comment section for the block to include what type of authorization statements will be carried out within the code.

core/zms/src/main/rdl/Policy.rdli
```diff
@@ -181,7 +181,7 @@ resource Assertion DELETE "/domain/{domainName}/policy/{policyName}/version/{ver
...
```
havetisyan, line 184: please update the comment section for the block to include what type of authorization statements will be carried out within the code.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -10589,6 +10607,16 @@ boolean isAllowedDeletePendingMembership(Principal principal, final String domai
         return pendingMember != null && principal.getFullName().equals(pendingMember.getRequestPrincipal());
     }
 
+    boolean isAllowedDeleteAssertion(Principal principal, final AthenzDomain domain, final String policyName, final Long assertionId) {
+          if (hasAccess(domain, "update", String.format("%s:%s%s", domain.getName(), POLICY_PREFIX, policyName), principal, null) == AccessStatus.ALLOWED) {
+              return true;
+          }
+          if (hasAccess(domain, "delete", String.format("%s:%s%d", domain.getName(), ASSERTION_PREFIX, assertionId), principal, null) == AccessStatus.ALLOWED) {
```
havetisyan, line 10614: this is the wrong authorization for the command. assertion is not a top level object within the domain and there is no requirement that the id is unique across the domain. It happens to be that way with our implementation but someone else might define their own store implementation and only support unique assertion ids per policy. The authorization would be 

authorize("delete", "{domainName}:policy.{policyName}.assertion.{assertionId}" )

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -10589,6 +10607,16 @@ boolean isAllowedDeletePendingMembership(Principal principal, final String domai
         return pendingMember != null && principal.getFullName().equals(pendingMember.getRequestPrincipal());
     }
 
+    boolean isAllowedDeleteAssertion(Principal principal, final AthenzDomain domain, final String policyName, final Long assertionId) {
+          if (hasAccess(domain, "update", String.format("%s:%s%s", domain.getName(), POLICY_PREFIX, policyName), principal, null) == AccessStatus.ALLOWED) {
```
havetisyan, line 10611: Instead of the String.format please use:

ResourceUtils.policyResourceName(domainName, policyName)

havetisyan (changes requested):

havetisyan:
Any reason only the delete assertion API was modified? If you're willing to grant someone access to delete assertions in the policy, isn't there a use case to also do the same for the put assertion api?

TakuyaMatsu:
> Any reason only the delete assertion API was modified? If you're willing to grant someone access to delete assertions in the policy, isn't there a use case to also do the same for the put assertion api?

Because the authority for PutAssertion is effectively the same as that for updating the policy, we decided not to create a separate permission for PutAssertion. (Creating a PutAssertion permission would introduce a vulnerability by allowing the addition of arbitrary privileges.)

havetisyan (changes requested):

havetisyan (approved):

Issue #2910, opened 2025-03-14
Auto-remove delegated assume roles/policies after a provider role is deleted to allow safe reuse of role names


hiragi-gkuth:
## Background

As mentioned in [role delegation](https://github.com/AthenZ/athenz/blob/master/docs/role_delegation.md),  the provider domain delegates the tenant domain to manage roles.

The problem is:
When the trust role in the provider domain is deleted, the corresponding assume role in the tenant domain is not automatically removed. As a result, if the trust role is later recreated under the same name but assigned a different policy, members of the tenant domain’s assume role may unintentionally gain new privileges.

The sequence is as follows:
```mermaid
sequenceDiagram
    participant PDAdmin as Provider Domain Admin
    participant TDAdmin as Tenant Domain Admin
    participant Athenz as Athenz ZMS

    Note over PDAdmin,TDAdmin: 1. Create delegated role in the Provider Domain
    PDAdmin->>Athenz: Create delegated role and policy named "get_secret" in Provider Domain
    Athenz-->>PDAdmin: role and policy("get_secret" grant 'get' to ~~~ on secretA) created


    Note over PDAdmin,TDAdmin: 2. Create assume role in the Tenant Domain
    TDAdmin->>Athenz: Create "assume_get_secret" referencing PD:role.get_secret
    Athenz-->>TDAdmin: "assume_get_secret" is established

    Note over PDAdmin,TDAdmin: 3. Provider Domain Admin delete "get_secret" role/policy
    PDAdmin->>Athenz: Delete delegated role and policy
    Athenz-->>PDAdmin: "get_secret" deleted
    Note right of Athenz: (But Tenant Domain's "assume_get_secret" still references<br>"PD:role.get_secret")

    Note over PDAdmin,TDAdmin: 4. After a while, recreate the "get_secret" with the same name,<br>but assign a new policy.
    PDAdmin->>Athenz: Create new delegated role and policy named "get_secret"
    Athenz-->>PDAdmin: role and policy("get_secret" grant 'get' to ~~~ on secretB) re-created

    Note over PDAdmin,TDAdmin: 5. Because the existing "assume_get_secret" references the same name,<br> Tenant Domain members unintentionally gain new privileges.
    Athenz-->>TDAdmin: (Unintentionally) "assume_get_secret" now has access<br> to the new policies on "get_secret"
    Note right of TDAdmin: Security issue: Old assume_get_secret members<br> gain access to new privileges
```

## Questions

[...]

PR #2919, opened 2025-04-01
[WIP] Remove the assume_role assertion when deleteRole

TakuyaMatsu:
# Description
#2910 

# Contribution Checklist:

### DIFF
--- a/clients/go/zms/client.go
+++ b/clients/go/zms/client.go
@@ -1206,12 +1206,12 @@ func (client ZMSClient) PutRole(domainName DomainName, roleName EntityName, audi
 	}
 }
 
-func (client ZMSClient) DeleteRole(domainName DomainName, roleName EntityName, auditRef string, resourceOwner string) error {
+func (client ZMSClient) DeleteRole(domainName DomainName, roleName EntityName, deleteAssumeRoleAssertion *bool, auditRef string, resourceOwner string) error {
 	headers := map[string]string{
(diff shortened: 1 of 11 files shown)


havetisyan:
Some comments on your POC PR:

a) While there is some minor ambiguity how we implement DELETE api with REST, but we're in the camp that DELETE endpoints cannot have query arguments so that will not be accepted.
b) Rather than making it as part of the delete api, a better option would to make this a s system setting. As a system admin, you may decide that you must always do clean-up so you can enable the system setting and then whenever a delegated role is deleted, carry out the cleanup.
c) The current implementation is not quite correct since you can't hide the clean up within the store implementation. The DB Service layer must know about the change since the last modified timestamp for both the policy and the domain must be updated so the change can be propagated to the ZTS servers.

TakuyaMatsu:
@havetisyan 
Thank you for your comments.
We have considered other approaches, but using query parameters seems to be the most effective for our use case.

For example, in the method you proposed in (b), the feature would be enabled system-wide via a system property. In that scenario, there is a risk that the assume_role assertion could be unintentionally deleted, which might cause issues in certain situations.
We also considered providing a flag for assume_role assertion deletion in a custom header or in the body of the DELETE request, but neither approach appeared suitable. Therefore, we would like to control whether or not this feature is enabled through a query parameter on the DELETE request, as proposed in this PR.

We would appreciate any further advice you might have.

PR #2935, opened 2025-04-16
Add restrictTo tag for sia-gce

YuchenWang01:
# Description
Add restrictTo tag for sia-gce

# Contribution Checklist:
- [] **The pull request does not introduce any breaking changes**

### DIFF
--- a/provider/gcp/sia-gce/authn.go
+++ b/provider/gcp/sia-gce/authn.go
@@ -25,6 +25,8 @@ import (
 )
 
 func GetGCEConfig(configFile, profileConfigFile, profileRestrictToKey, metaEndpoint, region string, provider provider.Provider) (*sc.Config, *sc.AccessProfileConfig, error) {
+
+	var profileConfig *sc.AccessProfileConfig
 	config, _, err := options.InitFileConfig(configFile, metaEndpoint, false, region, "", provider)
 	if err != nil {
 		log.Printf("Unable to process configuration file '%s': %v\n", configFile, err)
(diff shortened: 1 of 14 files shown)


provider/gcp/sia-gce/authn.go
```diff
@@ -36,24 +34,22 @@ func GetGCEConfig(configFile, profileConfigFile, metaEndpoint, region string, pr
 			log.Printf("Unable to process environment settings: %v\n", err)
 			// if we do not have settings in our environment, we're going
 			// to use fallback to retrieve values from the context ( metadata etc )
-			config, profileConfig, err = options.InitGenericProfileConfig(metaEndpoint, "", "", provider)
+			config, _, err = options.InitGenericProfileConfig(metaEndpoint, "", "", provider)
```
havetisyan, line 37: does this not change the behavior of the existing sia? We're ignoring the profile value we get from this API and instead get it from the GetGCEAccessProfile which seems to give preference to the config file as opposed to metadata.

provider/gcp/sia-gce/authn.go
```diff
@@ -44,16 +44,24 @@ func GetGCEConfig(configFile, profileConfigFile, metaEndpoint, region string, pr
...
 	}
 	if profileConfig == nil {
-		profileConfig, err := GetGCEAccessProfile(profileConfigFile, metaEndpoint, provider)
+		profileConfig, err := GetGCEAccessProfile(profileConfigFile, profileRestrictToKey, metaEndpoint, provider)
 		if err != nil {
 			log.Printf("Unable to determine user access management profile information: %v\n", err)
 		}
 		return config, profileConfig, nil
 	}
+	if profileConfig.ProfileRestrictTo == "" && profileRestrictToKey != "" {
```
havetisyan, line 53: We're repeating this block in two places. So we can leave this logic in this method and remove it from the GetGCEAccessProfile method. Then we just have to remove line 51 where we return from this method and then add profileConfig != nil to the if line on line 53. This way we only have the logic in a single place and not multiple.

havetisyan (approved):

PR #2957, opened 2025-05-07
add a log line in sia agent to show which spiffe uri is being used

abvaidya:
# Description

# Contribution Checklist:

## Attach Screenshots (Optional)


libs/go/sia/util/util.go
```diff
@@ -351,6 +351,7 @@ func GetSvcSpiffeUri(trustDomain, namespace, domain, service string) string {
 	} else {
 		uriStr = fmt.Sprintf("spiffe://%s/sa/%s", domain, service)
 	}
+	log.Printf("using spiffe uri: %s\n", uriStr)
```
havetisyan, line 354: our log should include the trustDomain and namespace values as well that could help determine what values were passed in (we already domain/service from the uri). So something like:

log.Printf("using spiffe uri: '%s' based on trustDomain: '%s' and namespace: '%s'\n", uriStr, trustDomain, namespace)

havetisyan (changes requested):

havetisyan (approved):

PR #2966, opened 2025-05-15
Ensure that the 'modified' column is updated even when column values do not change

TakuyaMatsu:
# Description
We expect the "modified" column in each table to be updated with the current time using ON UPDATE CURRENT_TIMESTAMP. However, in MySQL, if the columns targeted by the UPDATE statement do not actually change their values, the "modified" column (defined with ON UPDATE CURRENT_TIMESTAMP) will not be updated.

For example, in PutRole, if only the members are changed, there is no parameter change in the role table, and only the role_member table is updated.
Consequently, the "modified" column in the role table is not updated.

This PR ensures that the "modified" column is always updated whenever an UPDATE query is executed on each table.

# Contribution Checklist:


havetisyan:
@TakuyaMatsu thanks for the PR. Can you please extend some of the existing test cases in ZMS impl for this change? e.g. if we're calling putRole and only adding a new member then we should verify that the last modified value has changed. If we don't have a test case just adding a new member using putRole, then let's create a new one. thanks.

TakuyaMatsu:
@havetisyan 
Thank you for the feedback.
I've added a test case to verify that the 'modified' column is updated even when no changes are made.

TakuyaMatsu:
@havetisyan 
Thank you for the feedback.
I’ve added a test case to verify that the ‘modified’ column is updated even when no changes are made.

PR #2987, opened 2025-06-04
introduce a new domain meta attribute called oncall

havetisyan:
# Description

the attribute is expected to store the on-call support team/id that is responsible for this domain.

similar to other meta attributes, the DomainMetaStore interface has been updated to expose the following three methods:

    default boolean isValidOnCall(final String domainName, final String onCall) throws ServerResourceException;
    default void setOnCallDomain(final String domainName, final String onCall) throws ServerResourceException;
    default List<String> getValidOnCalls(final String userName) throws ServerResourceException;

the default no-op implementation has been updated as well.

# Contribution Checklist:


servers/zms/schema/updates/update-20250604.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `zms_server`.`domain` MODIFY COLUMN `on_call` VARCHAR(128) NOT NULL DEFAULT '';
```
chandrasekhar1996: should this be `ADD COLUMN` instead of `MODIFY COLUMN`?

havetisyan: Yes, Thanks!

havetisyan:
> Should we also add functionality to get a list of domains provided an on-call team/id? might be useful/intersting data to see how many products/prod domains is an oncall team associated with

Only if someone requests it. We did it for business services and I don't see anyone ever using it. it just creates unnecessary new query arguments. I would expect whatever system the on-call team is using will provide the same functionality so no point of using athenz for that query (e.g. users can use servicenow to look at a business service to see all the domains associated with it).

PR #2992, opened 2025-06-09
Feat: ConfigFiles for Multiple GitHub Action Provider Environments

mlajkim:
# Description

# Contribution Checklist:


libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceGithubActionsProp.java
```diff
@@ -0,0 +1,93 @@
+package com.yahoo.athenz.instance.provider.impl;
```
havetisyan, line 1: please add the Athenz Authors header block (see other files in the repo)

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceGithubActionsProp.java
```diff
@@ -0,0 +1,93 @@
...
+    // No-argument constructor
+    public InstanceGithubActionsProp() {
+    }
+
+    // Method to add properties
+    public void addProperties(String issuer, String providerDnsSuffix, String audience, String enterprise, String jwksUri) {
+        if (issuer == null || providerDnsSuffix == null || audience == null || jwksUri == null) {
+            throw new IllegalArgumentException("One of the required properties is null");
+        }
+        properties.put(issuer, new Prop(providerDnsSuffix, audience, enterprise, jwksUri));
```
havetisyan, line 43: we should add a check to see if the issuer is already set or not. otherwise, if the admin makes a mistake and specifies the same issuer in two different blocks, then the second one will override the first one and there would be no indication that it has taken place until the end users complain about their actions not working. e.g. both blocks have the same issuer:


`{
  "props": [
    {
      "provider_dns_suffix": "example-suffix1,example-suffix2",
      "enterprise": "enterprise1",
      "audience": "https://audience1.com",
      "issuer": "https://issuer1.com",
      "jwks_uri": "https://issuer1.com/jwks"
    },
    {
      "provider_dns_suffix": "example-suffix3,example-suffix4",
      "enterprise": "enterprise2",
      "audience": "https://audience2.com",
      "issuer": "https://issuer1.com",
      "jwks_uri": "https://issuer2.com/jwks"
    }
  ]
}`

also please add a corresponding test case for it.

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceGithubActionsProp.java
```diff
@@ -0,0 +1,93 @@
...
+        }
+    }
+
+    // No-argument constructor
+    public InstanceGithubActionsProp() {
+    }
+
+    // Method to add properties
+    public void addProperties(String issuer, String providerDnsSuffix, String audience, String enterprise, String jwksUri) {
+        if (issuer == null || providerDnsSuffix == null || audience == null || jwksUri == null) {
```
havetisyan, line 40: do we only care about nulls or should we also check for empty strings?

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceGithubActionsProp.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    public String getEnterprise(String issuer) {
+        if (!properties.containsKey(issuer)) {
+            return null;
+        }
+        return properties.get(issuer).enterprise;
+    }
+
+    public Boolean hasEnterprise (String issuer) {
```
havetisyan, line 75: please remove the space after the method name

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceGithubActionsProvider.java
```diff
@@ -76,49 +87,81 @@ public Scheme getProviderScheme() {
...
         // get default/max expiry time for any generated tokens - 6 hours
 
         certExpiryTime = Long.parseLong(System.getProperty(GITHUB_ACTIONS_PROP_CERT_EXPIRY_TIME, "360"));
 
-        // initialize our jwt processor
+        String githubIssuer = System.getProperty(GITHUB_ACTIONS_PROP_ISSUER, GITHUB_ACTIONS_ISSUER);
+
+        props.addProperties(
+            githubIssuer,
+            System.getProperty(GITHUB_ACTIONS_PROP_PROVIDER_DNS_SUFFIX, "github-actions.athenz.io"), // determine the dns suffix. if this is not specified we'll just default to github-actions.athenz.cloud
```
havetisyan, line 148: let's fix the comment to say githib-actions.athenz.io instead of athenz.cloud (the original code had the issue).

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceGithubActionsProvider.java
```diff
@@ -76,49 +87,81 @@ public Scheme getProviderScheme() {
...
+            System.getProperty(GITHUB_ACTIONS_PROP_AUDIENCE, "athenz.io"), // lookup the zts audience. if not specified we'll default to athenz.io
+            System.getProperty(GITHUB_ACTIONS_PROP_ENTERPRISE), // determine if we're running in enterprise mode
+            extractGitHubIssuerJwksUri(githubIssuer, System.getProperty(GITHUB_ACTIONS_PROP_JWKS_URI))
+        );
 
-        githubIssuer = System.getProperty(GITHUB_ACTIONS_PROP_ISSUER, GITHUB_ACTIONS_ISSUER);
-        jwtProcessor = JwtsHelper.getJWTProcessor(new JwtsSigningKeyResolver(extractGitHubIssuerJwksUri(githubIssuer), null));
+        try {
```
havetisyan, line 154: I'm not sure if I like the order of objects here since it kind of creates a somewhat ambiguous set up. I can configure both a file and property settings and they both will be set accordingly which is probably not what we expect. I'd like to have the following order where the preference is given to the file config and the property settings are read only if the file config is not set. So we should have the following logic:

// we should not catch any exceptions and instead any errors
// need to be reported to the caller as failures

initializeFromFilePath();

// at this point if our configuration is set as in we have valid
// entries in our props objects, we're done and nothing else to do

if (!props.isEmpty()) {
  return;
}

// otherwise add a new issuer entry based on the configured property values

String githubIssuer = System.getProperty(GITHUB_ACTIONS_PROP_ISSUER, GITHUB_ACTIONS_ISSUER);
.... (like you have the code above).

So with this model there is only a single way to configure the provider - either valid file config or system properties.

havetisyan: So just to follow-up, in the InstanceGithubActionsProp class you need to define a new member method called isEmpty that returns boolean whether or not member field properties hashmap has entries or not.

mlajkim:
Closed as no longer maintained due to other priorities

Will reopen/create a new PR once required

PR #2993, opened 2025-06-09
add additional settings when creating a group

tsultanov00:
# Description
Extended group creation to include additional settings that can be configured for the group
image
image

# Contribution Checklist:


ui/src/components/group/AddGroupAdvancedSettings.js
```diff
@@ -0,0 +1,224 @@
...
+
+    render() {
+        return [
+            <SettingRow
+                key={'setting-row-auditEnabled'}
+                name='auditEnabled'
+                label='Audit'
+                type='switch'
+                disabled={
+                    this.props.delegated ||
```
chandrasekhar1996: we are not passing delegated to this class and groups also do not have delegation. Need to remove this check and in all remaining places in file as well

tsultanov00: Good catch, removed all mentions of delegated

ui/src/components/group/AddGroupAdvancedSettings.js
```diff
@@ -0,0 +1,211 @@
...
+            <SettingRow
+                key={'setting-row-auditEnabled'}
+                name='auditEnabled'
+                label='Audit'
+                type='switch'
+                disabled={
+                    !this.props.isDomainAuditEnabled ||
+                    this.props.members.length
+                }
+                desc={ADD_ROLE_AUDIT_DESC}
```
chandrasekhar1996: should be group description for all the settings here

ui/src/__tests__/components/history/__snapshots__/HistoryList.test.js.snap
```diff
@@ -380,10 +380,11 @@ exports[`HistoryList should render 1`] = `
           data-testid="flatPicker"
           placeholder="Start Date (Optional)"
           type="hidden"
-          value="2024-10-20 07:51"
+          value=""
```
chandrasekhar1996: can you check why this value is not populated anymore? It should be based on the test since we set startDate and endDate. Same for RoleHistoryList test as well, possible might be related to changes done in ui/src/components/utils/DateUtils.js

chandrasekhar1996 (changes requested):

ui/src/components/group/AddGroup.js
```diff
@@ -349,69 +452,112 @@ class AddGroup extends React.Component {
...
+                </SectionDiv>
+
+                <SectionDiv title={auditToolTip}>
+                    <StyledInputLabel style={reviewTriggerStyle} />
+                    <ButtonDiv style={reviewTriggerStyle}>
+                        <StyledButton
+                            secondary
+                            size={'small'}
+                            onClick={this.addMember}
+                            data-wdio={'add-role-member'}
```
chandrasekhar1996: group-member

ui/src/components/flatpicker/FlatPicker.js
```diff
@@ -90,6 +90,9 @@ export default class FlatPicker extends React.Component {
         flatpickr('.' + fpClass, {
             onChange: this.onChange,
             onClose: this.onClose,
+            onReady: (dates, dateStr, instance) => {
+                instance.altInput.id = 'member-expiry';
```
chandrasekhar1996: this would give every flatpickr the id "member-expiry" which is not the case since we also use flatpickr for history dates

chandrasekhar1996: based on line 111, it does have data-testid='flatPicker', we cannot use that?

tsultanov00: > based on line 111, it does have data-testid='flatPicker', we cannot use that?

This creates a hidden input which we cannot target in our tests



> this would give every flatpickr the id "member-expiry" which is not the case since we also use flatpickr for history dates

I assigned the prop ID so each flatpicker will have its own ID

ui/src/components/group/AddGroupAdvancedSettings.js
tsultanov00: Refactored the constants so that they can be reused for roles and groups

ui/src/__tests__/components/history/__snapshots__/HistoryList.test.js.snap
tsultanov00: The test used a date object as opposed to a string representation of the date. I made a small change in DateUtils to cater for that, the date is now reflected in the value once again

ui/src/components/group/AddGroup.js
```diff
@@ -349,69 +452,112 @@ class AddGroup extends React.Component {
...
+                </SectionDiv>
+
+                <SectionDiv title={auditToolTip}>
+                    <StyledInputLabel style={reviewTriggerStyle} />
+                    <FlatPickrStyled style={reviewTriggerStyle}>
+                        <Flatpicker
+                            onChange={memberExpiryDateChanged}
+                            clear={this.state.memberExpiry}
+                            id='addrole'
+                            data-testid='memberExpiry123'
```
ArtjomsPorss: `memberExpiry` without 123 seems better

ui/src/components/utils/DateUtils.js
```diff
@@ -61,7 +61,7 @@ class DateUtils {
     // converts a datetime returned by the datetime picker into an
     // RDL "Timestamp" (e.g. "2018-10-09T22:55:57.389Z")
     uxDatetimeToRDLTimestamp(ux) {
-        if (ux) {
+        if (ux || ux.length) {
```
ArtjomsPorss: should be `&&` instead of `||`

ui/src/components/group/AddGroup.js
```diff
@@ -349,69 +452,112 @@ class AddGroup extends React.Component {
...
+                    </ContentDiv>
+                </SectionDiv>
+
+                <SectionDiv title={auditToolTip}>
+                    <StyledInputLabel style={reviewTriggerStyle} />
+                    <FlatPickrStyled style={reviewTriggerStyle}>
+                        <Flatpicker
+                            onChange={memberExpiryDateChanged}
+                            clear={this.state.memberExpiry}
+                            id='addrole'
```
ArtjomsPorss: `addrole`here should be different, it's not a role. It should should be less generic, probably something like `add-group-member-expiry-picker` or similar.

ui/src/components/group/AddGroup.js
tsultanov00: no idea where that came from 😅

ui/src/components/utils/DateUtils.js
tsultanov00: Doing so returns an empty string if a Date object is passed in since objects don't have a .length property. This allows for both string representation of a date and a Date object to be passed in. I also noticed that after a date has been chosen, opening the FlatPicker menu again and clicking outside of the box clears the selection but really sets the value as an empty array and so submitting the changes caused a runtime error to occur. This takes care of that scenario also.

ArtjomsPorss: currently if `ux` is `null`, it will attempt to call `ux.length`
maybe then it could be something like this:
```js
if (ux instanceof Date || typeof ux === 'number' || (typeof ux === 'string' && ux.length)) {
...
}
```

tsultanov00: I've added a check to test for a Date object and non empty string, as that is all that we use throughout the repo

PR #2996, opened 2025-06-12
Add support to a different client_id jwt claim name

yosrixp:
# Description

# Contribution Checklist:

### DIFF
--- a/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/token/DefaultOAuthJwtAccessToken.java
+++ b/libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/token/DefaultOAuthJwtAccessToken.java
@@ -29,7 +29,7 @@ import org.slf4j.LoggerFactory;
 public class DefaultOAuthJwtAccessToken implements OAuthJwtAccessToken {
 
     private static final Logger LOG = LoggerFactory.getLogger(DefaultOAuthJwtAccessToken.class);
-    private final String jwtClientIdClaimName = System.getProperty("security.jwt.claim.client_id.name", OAuthJwtAccessToken.CLAIM_CLIENT_ID);
+    private static final String JWT_CLIENT_ID_CLAIM_NAME = System.getProperty("athenz.auth.jwt.claim.client_id.name", OAuthJwtAccessToken.CLAIM_CLIENT_ID);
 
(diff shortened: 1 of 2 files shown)


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/token/DefaultOAuthJwtAccessToken.java
```diff
@@ -29,6 +29,7 @@
 public class DefaultOAuthJwtAccessToken implements OAuthJwtAccessToken {
 
     private static final Logger LOG = LoggerFactory.getLogger(DefaultOAuthJwtAccessToken.class);
+    private final String jwtClientIdClaimName = System.getProperty("security.jwt.claim.client_id.name", OAuthJwtAccessToken.CLAIM_CLIENT_ID);
```
havetisyan: the properties in auth-core must start with "athenz.auth."

havetisyan: this must be static field. there is no need to extract the value from system property for every invocation. just need to update the test case accordingly

havetisyan (changes requested):

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/oauth/token/DefaultOAuthJwtAccessToken.java
yosrixp: Let me fix it.

yosrixp: I had it initially static but because the class is used multiple times so the property value won't be changed. Also if you have by any chance different JWTClaimsSet from different sources one would work as the static member will be initialized once. so you can't have it both ways.

havetisyan: not sure what you mean. The system property is set for the application. so you can't have the same class handle different property values within the same application. As such, having it as a static is the correct approach.

havetisyan (approved):

PR #2998, opened 2025-06-13
add oTel instrumentation to SIA

py4chen:
This PR adds oTel instrumentation to SIA agent, and initializes oTel provider before the start of RunAgent to export oTel metrics to oTel collector (or prometheus).

image

# Description

Here is a sample oTel configuration fields available in the sia json config:

```
{
  "version": "1.0.0",
  "service": "{{ .Values.athenz.service }}",
  "accounts": [
    {
      "domain":  "{{ .Values.athenz.domain }}",
      "account": "{{ .Values.athenz.awsAccount }}"
    }
  ],
  "otel": {
    "collector_endpoint": "yahoo-otel-collector-ds-collector.kube-addons.svc.cluster.local:4318"
     “mtls”: false, # optional, use mtls client cert to authenticate otel collector only if it is true
     “service_instance_id”: “”, # optional, used for specifying the instance ID field for chronosphere
  }
}
```

The following are optional, but will not be explicitly released to the public:
```
“otel”: {
    “client_cert_path”: “”, # optional, default to the first service identity configured in SIA, mTLS only
    “client_key_path”: “”, # optional, default to the first service identity configured in SIA, mTLS 
   “ca_cert_path”: “”, # optional, default to AthenzCACertFile (which is /var/lib/sia/ca.cert.pem) 
}    
```

SIA users can also configure the oTel via environment variables instead of the json config:

<meta charset="utf-8"><b style="font-weight:normal;" id="docs-internal-guid-86b36738-7fff-5189-86ba-f8e8445a6c2c"><div dir="ltr" style="margin-left:0pt;" align="left">
oTel json field | Environment Variable | Sample values
-- | -- | --
collector_endpoint | OTEL_COLLECTOR_ENDPOINT | yahoo-otel-collector-ds-collector.kube-addons.svc.cluster.local:4318
mtls | OTEL_MTLS | “True”, “False”, “true”, “false”, “1”, “0”
service_instance_id | OTEL_SERVICE_INSTANCE_ID |  
client_cert_path | OTEL_CLIENT_CERT_PATH |  
client_key_path | OTEL_CLIENT_KEY_PATH |  
ca_cert_path | OTEL_CA_CERT_PATH |  

</div></b>

--

</div><br /></b>

Here are the oTel metrics that it collects:

* Run status Increment counters for each SIA runs

```
sia_agent_command_result_total{function="GetRoleCertificates",instance="a9df150b-94f8-4bc9-ba2c-3b988c5d9b23",job="sia",result="success"} 1


sia_agent_command_result_total{function="refreshSvc",instance="a9df150b-94f8-4bc9-ba2c-3b988c5d9b23",job="sia",result="success"} 1
```

*Expiry Secs metric: remaining seconds until expiry for service cert and role cert

```
sia_service_cert_validity_remaining_secs_ratio{cname="sshca.stage.sshra-ws-func",instance="5df1a21e-66db-4d4c-9edd-6042602ecea8",job="sia"} 604799
```

* ZTS interaction: latency and response code from the athenz ZTS client.

```
http_client_request_duration_seconds_bucket{http_request_method="POST",http_response_status_code="200",instance="a9df150b-94f8-4bc9-ba2c-3b988c5d9b23",job="sia",network_protocol_name="http",network_protocol_version="1.1",server_address="zts.athenz.ouroath.com",server_port="4443",url_scheme="https",le="0.25"} 1
```

# Contribution Checklist:


go.mod
```diff
@@ -27,6 +27,13 @@ require (
 	github.com/patrickmn/go-cache v2.1.0+incompatible
 	github.com/stretchr/testify v1.10.0
 	github.com/tailscale/peercred v0.0.0-20240214030740-b535050b2aa4
+	github.com/theparanoids/crypki v1.20.7
```
py4chen, line 30: crypki provides the package for cert reloading (from the disk).

libs/go/sia/otel/metricset.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		log.Printf("Error creating metric for %s: %v\n", metricNameRoleCertValidityRemainingSecs, err)
+	}
+
+	metricSet.agentCmdResultCounter, err = meter.Int64Counter(
+		metricAgentCommandResult,
+		metric.WithDescription("Counts the total number of agent command executions by type and result"),
+	)
+	if err != nil {
+		panic(err) // or handle gracefully
```
havetisyan: we must not panic for metrics. we should just log the error and continue processing without any metrics.

libs/go/sia/agent/agent.go
```diff
@@ -666,42 +670,50 @@ func runAgentCommand(siaCmd, ztsUrl string, opts *sc.Options) {
 		if len(failures) != 0 {
 			util.ExecuteScript(opts.RunAfterCertsErrParts, strings.Join(failures, ","), false)
 			if !skipErrors {
+				otel.RecordAgentCommandResult("rolecert", false)
```
havetisyan: any reason why we are differentiating the calls between command line options and the regular calls: e.g.

rolecert vs. GetRoleCertificates
post vs. registerSvc
token vs. accessTokenRequest
etc.

why not use the function name everywhere - e.g. GetRoleCertificates everywhere and not rolecert

havetisyan (changes requested):

havetisyan:
@py4chen also please add the Athenz Authors header block at the top of all the new files

libs/go/sia/agent/agent.go
py4chen: Earlier, I was considering putting the switch case name into the call if it's not the default case, and using the function name if it is the default case. But right, that made it confusing. 
I have unified them to use function names.

py4chen:
(rebased to master branch to fix conflicts)

havetisyan:
@py4chen I'm trying to understand the impact if there is nothing configured for the CollectorEndpoint. As in, I just update, I get the otel enabled sia agent but I haven't configured anything so I'm not expecting any impact.

I see that the init call will return NilShutdown function which does nothing. However, I see that for every command in the agent, we execute RecordAgentCommandResult method (what exactly happens here?) and what is calling the init method in otel? Let's talk about this tomorrow.

py4chen:
> @py4chen I'm trying to understand the impact if there is nothing configured for the CollectorEndpoint. As in, I just update, I get the otel enabled sia agent but I haven't configured anything so I'm not expecting any impact.
> 
> I see that the init call will return NilShutdown function which does nothing. However, I see that for every command in the agent, we execute RecordAgentCommandResult method (what exactly happens here?) and what is calling the init method in otel? Let's talk about this tomorrow.

`RecordAgentCommandResult` method would send metrics to oTel metric interfaces. Then on the other side, the oTel provider initilaized in the `init` function will export the metrics to the otel collector.

If nothing is configured for the CollectorEndpoint, no oTel provider will be initiliazed, and a nil function is returned for shutdown. In this case, all the metrics sent to the interface will be dropped.

py4chen:
> > @py4chen I'm trying to understand the impact if there is nothing configured for the CollectorEndpoint. As in, I just update, I get the otel enabled sia agent but I haven't configured anything so I'm not expecting any impact.
> > I see that the init call will return NilShutdown function which does nothing. However, I see that for every command in the agent, we execute RecordAgentCommandResult method (what exactly happens here?) and what is calling the init method in otel? Let's talk about this tomorrow.
> 
> `RecordAgentCommandResult` method would send metrics to oTel metric interfaces. Then on the other side, the oTel provider initilaized in the `init` function will export the metrics to the otel collector.
> 
> If nothing is configured for the CollectorEndpoint, no oTel provider will be initiliazed, and a nil function is returned for shutdown. In this case, all the metrics sent to the interface will be dropped.

I have made the changes to ensure that no 3rd-party oTel packages are invoked when the oTel is disabled (CollectorEndpoint field is empty in the config). All the oTel related invocation are routed to our wrapper package 'otel', which is the only package imports the opensource oTel packages.

I have tested this PR changes by building a sia-eks container image and tested it in an EKS cluster.

libs/go/sia/otel/http_metric_sender.go
```diff
@@ -0,0 +1,186 @@
...
+	oTelEnabled = true
+
+	// For single TLS: directly init if the CA file is ready.
+	if !oTelCfg.MTLS {
+		if _, err := os.Stat(oTelCfg.CACertPath); err == nil {
+			log.Println("oTel: TLS CA cert is ready, initializing OTel SDK")
+			shutdownFn, err := initializeOTelSDK(context.Background(), oTelCfg)
+			if err != nil {
+				log.Printf("oTel: SDK init failed: %v", err)
+				return NilShutdown
```
havetisyan, line 67: at this point we should set oTelEnabled back false since our init failed.

libs/go/sia/otel/http_metric_sender.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		return nil
+	}
+
+	log.Println("oTel: waiting for TLS files to be ready before starting OTel provider")
+	go func() {
+		if err := waitForTLSFileReady(oTelCfg); err != nil {
+			log.Printf("oTel: failed to wait for TLS files ready, err: %v\n", err)
+			shutdownCh <- NilShutdown
+			return
```
havetisyan, line 92: at this point we should set oTelEnabled back false since our init failed.

libs/go/sia/otel/http_metric_sender.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+		shutdownFn, err := initializeOTelSDK(context.Background(), oTelCfg)
+		if err != nil {
+			log.Printf("oTel: SDK init failed: %v", err)
+			shutdownCh <- NilShutdown
+			return
```
havetisyan, line 99: at this point we should set oTelEnabled back false since our init failed.

py4chen:
I fixed the unittest, and make sure they have passed locally:
```
--- PASS: TestTouchDoneFile (0.00s)
PASS
ok      github.com/AthenZ/athenz/libs/go/sia/util       2.319s
?       github.com/AthenZ/athenz/libs/go/sia/verify     [no test files]
```

PR #3008, opened 2025-07-01
fetch latest enterprise appliance IPs from msd

tsultanov00:
# Description
* Updated API call to fetch latest enterprise appliance IPs from MSD

# Contribution Checklist:

### DIFF
--- a/ui/src/api.js
+++ b/ui/src/api.js
@@ -1923,11 +1923,26 @@ const Api = (req) => {
             });
         },
 
-        getInstances(domainName, serviceName, category) {
+        getInstances(domainName, serviceName, category, _csrf) {
+            const body = {
+                domainServices: [
+                    {
+                        domainName: domainName,
+                        serviceNames: [serviceName],
+                    },
+                ],
+                resolveStaticWorkloads: category === 'static',
+                fetchDynamicTypeWorkloads: category === 'dynamic',
+            };
             return new Promise((resolve, reject) => {
                 fetchr
                     .read('instances')
-                    .params({ domainName, serviceName, category })
+                    .params({ category, body })
+                    .clientConfig({
+                        headers: {
+                            'x-csrf-token': _csrf,
+                        },
+                    })
                     .end((err, workloadList) => {
                         if (err) {
                             reject(err);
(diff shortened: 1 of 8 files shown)


ui/src/server/handlers/api.js
```diff
@@ -3044,21 +3044,27 @@ Fetchr.registerService({
 
 Fetchr.registerService({
     name: 'instances',
-    read(req, resource, params, config, callback) {
-        req.clients.msd.getWorkloadsByService(
-            { domainName: params.domainName, serviceName: params.serviceName },
+    create(req, resource, params, body, config, callback) {
```
ArtjomsPorss: what is the idea behind switching from read to create request type? Are we doing any changes with this request on the backend?

ArtjomsPorss (approved):

ui/src/server/handlers/api.js
tsultanov00: It wasn't initially documented, but from digging a bit further into the fetchr src code, there is a flag that we can use to allow POST requests with the read operation. I have made some changes to reflect this

ui/src/api.js
```diff
@@ -1923,11 +1923,26 @@ const Api = (req) => {
...
-        getInstances(domainName, serviceName, category) {
+        getInstances(domainName, serviceName, category, _csrf) {
+            const body = {
+                domainServices: [
+                    {
+                        domainName: domainName,
+                        serviceNames: [serviceName],
+                    },
+                ],
+                resolveStaticWorkloads: true,
```
chandrasekhar1996: one small improvement: depending on the category we can set resolveStaticWorkloads/fetchDynamicTypeWorkloads to true/false. 
If category is static then we do not look at the dynamic response so will be efficient if we set fetchDynamicTypeWorkloads to false and vice versa with fetchDynamicTypeWorkloads

tsultanov00: Good observation, I have made a change to reflect this

ui/src/components/service/InstanceTable.js
```diff
@@ -83,7 +83,7 @@ export default class InstanceTable extends React.Component {
                         service={this.props.service}
                         onUpdateSuccess={this.props.onSubmit}
                         color={color}
-                        key={item.uuid}
+                        key={i}
```
tsultanov00, line 86: uuid property doesn't exist under item, this gets rid of a react warning

PR #3011, opened 2025-07-03
allow deletes by req principal in self serve roles

abvaidya:
# Description

# Contribution Checklist:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
@@ -5430,7 +5430,8 @@ public class ZMSImpl implements Authorizer, KeyStore, ZMSHandler {
         // delete operation if the principal matches membername
 
         final String normalizedMember = normalizeDomainAliasUser(memberName);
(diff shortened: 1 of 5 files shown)


servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -5457,6 +5458,18 @@ public void deleteMembership(ResourceContext ctx, String domainName, String role
         dbService.executeDeleteMembership(ctx, domainName, roleName, normalizedMember, auditRef, caller);
     }
 
+    private boolean isDeleteFromReqPrincipal(Principal principal, String memberName, Role role) {
+        if (role.getSelfServe() == Boolean.TRUE && role.getRoleMembers() != null) {
+            for (RoleMember rm : role.getRoleMembers()) {
+                if (rm.getMemberName().equals(memberName) && rm.getRequestPrincipal() != null &&
```
havetisyan: we can probably optimize the code here and not continue if there is a match. We're guaranteed that the memberName only appears one in the role. So once we find it, we can just return true/false and not continue looking at other members.

PR #3027, opened 2025-08-01
Add an automatic cleanup feature for delegated roles.

gyakami:
# Description
## Overview

This PR adds a feature to automatically delete related `assume_role` assertions in tenant domains when a provider role is deleted.
https://github.com/AthenZ/athenz/issues/2910

## Changes

1.  **Domain Model Extension**
    -   Added an `autoDeleteTenantAssumeRoleAssertions` flag to all domain types (Domain, DomainData, DomainMeta, SubDomain, TopLevelDomain, UserDomain).
    -   The default value is false (opt-in).

2.  **Database Schema Update**
    -   Added an `auto_delete_tenant_assume_role_assertions` column (TINYINT(1)) to the `domain` table.
    -   Added the update script `update-20250711.sql`.

3.  **JDBCConnection Class Extension**
    -   Added the `deleteAssumeRoleAssertions` method to find and delete `assume_role` assertions in a tenant domain that reference a specific provider role.

4.  **DBService Class Extension**
    -   Added the `executeDeleteAssumeRoleAssertions` methods.

5.  **ZMSImpl Class Extension**
    -   Added an `autoDeleteTenantAssumeRoleAssertions` flag (loaded from the configuration file).
    -   Extended the `deleteRole` method to also delete tenant `assume_role` assertions when the conditions are met.

6.  **Configuration File Update**
    -   Added the `athenz.zms.auto_delete_tenant_assume_role_assertions` setting to `zms.properties` (default is false).

## Behavior Details

1.  When a provider role is deleted, the system checks for tenant domains that trust this role.
2.  Automatic cleanup is performed if all of the following conditions are met:
    -   The `autoDeleteTenantAssumeRoleAssertions` flag is set to `true` in the server configuration.
    -   The `autoDeleteTenantAssumeRoleAssertions` flag is set to `true` for the tenant domain.
3.  The system searches for `assume_role` assertions within the tenant domain that reference the provider role being deleted.
4.  It deletes the corresponding assertions and updates the modification timestamp of the related policies.
5.  The delete operation is recorded in the audit log.

## Tests

-   Added unit tests (JDBCConnectionTest, DBServiceTest, ZMSImplTest).
-   Covers both success and failure cases.
-   Includes tests for various combinations of settings (server setting ON/OFF, domain setting ON/OFF).

## Impact

This change is opt-in and does not affect existing behavior. To use this new feature, both the server and the domain settings must be enabled.

# Contribution Checklist:


havetisyan:
@gyakami thanks for the PR. a couple of general comments before I start looking at your PR.
a) please run `mvn clean install` from the top level athenz directory to make sure all test requirements pass. I see there are no test cases for the new attribute in the zms-core module so that would fail since that module has 100% coverage.
b) for the DB update, please follow the instructions in the repo on how to update the schema: https://github.com/AthenZ/athenz/blob/master/servers/zms/schema/schema_update.md

gyakami:
@havetisyan 
Thanks for your feedback.
I've addressed the two points you raised:

a) I've added test cases for the new attribute in the zms-core module and confirmed that mvn clean install now passes.
b) I've updated the DB schema by following the instructions provided in the repository.

Could you please take another look when you have a moment? Thank you.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4890,6 +4895,16 @@ public void deleteRole(ResourceContext ctx, String domainName, String roleName,
         } catch (ServerResourceException ex) {
             throw ZMSUtils.error(ex);
         }
+        
+        // Automatically purge tenant‑side assume_role assertions that reference the role we are about to delete.
+        if (autoDeleteTenantAssumeRoleAssertions) {
+            Domain domain = dbService.getDomain(role.trust, useMasterCopyForSignedDomains);
```
havetisyan: This call will throw an exception and the delete role request will always be rejected if the role does not have a trust set.

Please add a test case where the setting is enabled but we're just deleting a regular role which would fail with the above code.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -4890,6 +4895,16 @@ public void deleteRole(ResourceContext ctx, String domainName, String roleName,
...
+            // Only proceed if the tenant domain explicitly opts‑in to the cleanup.
+            if (domain.autoDeleteTenantAssumeRoleAssertions) {
```
havetisyan: we typically don't access the fields directly (at some point we might move away from rdl and those may not be public) so please use the appropriate get method to get the attribute value.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4406,6 +4458,67 @@ void executeDeleteAssertion(ResourceContext ctx, String domainName, String polic
...
+                cacheStore.invalidate(domainName);
+
+                for (Policy policy : policies) {
+                    List<Assertion> deletedAssertions = policy.getAssertions();
+                    if (deletedAssertions == null || deletedAssertions.isEmpty()) {
+                        // DAO was unable to populate deleted assertions; skip logging.
+                        continue;
+                    }
+
+                    for (Assertion assertion : deletedAssertions) {
```
havetisyan, line 4302: is there any reason why we're creating an audit log entry per assertion instead of just one per policy?

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -2688,6 +2688,58 @@ void executeDeleteRole(ResourceContext ctx, String domainName, String roleName,
         }
     }
 
+    void executeDeleteRoleWithAssumeRoleAssertions(ResourceContext ctx, String domainName, String roleName,
```
havetisyan: I only see references to this function in the test files. Where is this used in the server?

servers/zms/schema/zms_server.sql
```diff
@@ -1,5 +1,5 @@
 -- MySQL Script generated by MySQL Workbench
--- Tue Jun  3 17:51:01 2025
+-- Fri Aug  8 14:47:06 2025
```
havetisyan, line 2: the PR is missing the updated workbench file. Please include that as well.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
```diff
@@ -4406,6 +4458,67 @@ void executeDeleteAssertion(ResourceContext ctx, String domainName, String polic
...
+                // nothing to clean up.  Simply return without raising an error so that the
+                // caller can continue deleting the provider role.
+                if (policies == null || policies.isEmpty()) {
+                    return;   // nothing to do
+                }
+
+                con.updateDomainModTimestamp(domainName);
+                cacheStore.invalidate(domainName);
+
+                for (Policy policy : policies) {
```
havetisyan, line 4286: During the api call, the last modification timestamp for the policy was updated. However, the domain last modified timestamp must be updated as well. So when we get the data here we need to create a set to include all the domains were affected and then for every domain:

a) update the last modification timestamp
b) invalidate the domain from the local cache.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -3346,6 +3367,67 @@ public boolean deleteAssertion(String domainName, String policyName, String vers
...
+
+        // 1) Locate every policy in the *tenant* domain that contains an
+        //    `assume_role` assertion pointing to the soon‑to‑be‑deleted provider
+        //    role ({providerDomain}:role.{providerRole}).
+        Set<Integer> policyIds = new HashSet<>();
+        List<Policy> policies  = new ArrayList<>();
+
+        try (PreparedStatement ps = con.prepareStatement(SQL_SELECT_TENANT_ASSUME_ROLE_POLICY_IDS)) {
+            ps.setString(1, domainName);
+            ps.setString(2, providerDomainName + ":role." + providerRoleName);
```
havetisyan: Please use: ResourceUtils.roleResourceName(providerDomainName, providerRoleName)

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -3346,6 +3367,67 @@ public boolean deleteAssertion(String domainName, String policyName, String vers
...
+
+        // Nothing to do if no matching policies were found.
+        if (policyIds.isEmpty()) {
+            return policies;
+        }
+
+        // 2) Remove the matching assume_role assertions themselves. 
+        try (PreparedStatement ps = con.prepareStatement(SQL_DELETE_TENANT_ASSUME_ROLE_ASSERTIONS)) {
+            ps.setString(1, domainName);
+            ps.setString(2, providerDomainName + ":role." + providerRoleName);
```
havetisyan: Please use: ResourceUtils.roleResourceName(providerDomainName, providerRoleName), and for that matter instead of calling the api twice maybe create a final string at the beginning of the method and then use it in both places

havetisyan (changes requested):

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
gyakami: I have fixed it!

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
gyakami: I have fixed it!

servers/zms/schema/zms_server.sql
gyakami: I've added the updated workbench file!

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
gyakami: You're right, my mistake. I've deleted executeDeleteRoleWithAssumeRoleAssertions as we're now only using executeDeleteAssumeRoleAssertions.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
gyakami: I modified it to use the get method.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
gyakami: I've changed the logic to only delete the role if the trust domain is empty and added a test case for it.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
gyakami: There was no specific reason.  I've updated the code to group audit logs by policy.

servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
gyakami: Isn't it only the tenant domain that the Delegated Role trusts (the domainName in this method) that's affected?

gyakami:
@havetisyan 
My apologies for the delay. I have replied to all the comments and/or revised the code. Please take a look when you have a moment.

servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
```diff
@@ -1517,6 +1517,80 @@ public void testExecuteDeleteServiceIdentity() throws ServerResourceException {
...
+            fail();
+        } catch (ResourceException ex) {
+            assertEquals(ResourceException.NOT_FOUND, ex.getCode());
+            assertTrue(ex.getMessage().contains(
+                    "unable to delete assume role assertions"));
+        } finally {
+            zms.dbService.defaultRetryCount = saveRetryCount;
+            zms.dbService.store             = saveStore;
+        }
+    }
```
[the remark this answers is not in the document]
gyakami: I'm sorry, I have deleted the test.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -3348,6 +3369,69 @@ public boolean deleteAssertion(String domainName, String policyName, String vers
...
+                executeUpdate(ps, caller);
+            }
+        } catch (SQLException ex) {
+            throw sqlError(ex, caller);
+        }
+
+        // Return the snapshot of policies that existed *before* we removed the
+        // assertions – the caller uses this for auditing and notifications.
+        return policies;
+    }
```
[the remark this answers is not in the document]
gyakami: > There's a critical issue in this method that will lead to incorrect audit logs. The method currently calls savePolicySettings, which in turn calls listAssertions to fetch all assertions for a policy. This means that when a policy has other assertions besides the assume_role one being deleted, the returned Policy object will contain all of them. The DBService layer then incorrectly logs all of these assertions as deleted.

This comment is incorrect. The savePolicySettings() method only sets policy metadata from the ResultSet; it does not fetch assertions, nor does it call listAssertions. Therefore, This problem will not occur.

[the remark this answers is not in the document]
gyakami: > Additionally, the method updates policy modification timestamps in a loop, which is inefficient as it results in multiple database round-trips.

I've updated the code to use a single query now.

gyakami:
@havetisyan 
I’ve also addressed the Gemini Code Review comments, so please take a look when you have a moment.

gyakami:
@havetisyan
When I look at the GitHub Actions logs, the following tests for athenz-zms-server are failing. However, the tests succeed on my Mac. It also seems that the number of tests is different. I'm stuck because I don't know the reason for this. Do you have any ideas what might be causing this?

```
PutGroupMembershipNotificationTaskTest.testGenerateAndSendPostPutGroupMembershipNotificationNotifyGroups
PutGroupMembershipNotificationTaskTest.testGenerateAndSendPostPutGroupMembershipNotificationNullOrgGroup
```

GitHub Actions
`[ERROR] Tests run: 1589, Failures: 2, Errors: 0, Skipped: 0`
Mac
`[INFO] Tests run: 1588, Failures: 0, Errors: 0, Skipped: 0`

havetisyan (approved):

PR #3030, opened 2025-08-02
ParameterManagerPrivateKeyStore implementation for KeyStore using GCP Parameter Manager

psasidhar:
# Description
Providing KeyStore implementation for GCP using Parameter Manager
Moving common code among Cloud Parameter storage into a static methond in a Util class that can be reused

# Contribution Checklist:


libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/PrivateKeyStoreUtil.java
```diff
@@ -0,0 +1,76 @@
...
+import java.util.function.Function;
+
+public class PrivateKeyStoreUtil {
+    private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    private static final String ZMS_SERVICE = "zms";
+    private static final String ZTS_SERVICE = "zts";
+    private static final String MSD_SERVICE = "msd";
+
+    private static final String ATHENZ_PROP_ZMS_KEY_NAME    = "athenz.aws.zms.key_name";
```
havetisyan: For the key names, we should make them cloud specific. It's not right to say I have a keystore implementation for gcp but my property name is based on aws. So it should be:

"athenz." + cloud + "." + service + ".key_name"

add add the cloud as an argument to the getPrivateKeyFromCloudParameter method

libs/java/server_gcp_common/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+  <artifactId>athenz-server-gcp-common</artifactId>
+  <name>athenz-server-gcp-common</name>
+  <packaging>jar</packaging>
+
+  <dependencyManagement>
+    <dependencies>
+      <dependency>
+        <groupId>com.google.cloud</groupId>
+        <artifactId>libraries-bom</artifactId>
+        <version>26.64.0</version>
```
havetisyan: we should not hard code the version here but use the version parameter from the main pom.xml

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/key/impl/ParameterManagerPrivateKeyStoreFactory.java
```diff
@@ -0,0 +1,56 @@
...
+            throw new RuntimeException("Failed to create ParameterManagerClient in ParameterManagerPrivateKeyStore", ex);
+        }
+    }
+
+    public static ParameterManagerClient createParameterManagerClient(String location) throws IOException {
+        if (isGlobalLocation(location)) {
+            return ParameterManagerClient.create();
+        }
+
+        String apiEndpoint = String.format("parametermanager.googleapis.com:443", location);
```
havetisyan: where is location used?

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/key/impl/ParameterManagerPrivateKeyStoreFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+        ParameterManagerSettings parameterManagerSettings =
+                ParameterManagerSettings.newBuilder().setEndpoint(apiEndpoint).build();
+
+        return ParameterManagerClient.create(parameterManagerSettings);
+    }
+
+    public static boolean isGlobalLocation(String location) {
+        return "global".equalsIgnoreCase(location);
```
havetisyan: we use "global" in a couple of places so we should probably define as a static const

havetisyan (changes requested):

libs/java/server_gcp_common/pom.xml
```diff
@@ -0,0 +1,52 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.yahoo.athenz</groupId>
+    <artifactId>athenz</artifactId>
+    <version>1.12.23-SNAPSHOT</version>
+    <relativePath>../../../pom.xml</relativePath>
+  </parent>
+
+  <groupId>com.yahoo.athenz</groupId>
```
havetisyan, line 13: let' define a code coverage block as well - hopefully with 100% coverage

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/PrivateKeyStoreUtil.java
psasidhar: Good point. Updated the code.

libs/java/server_gcp_common/pom.xml
psasidhar: fixed it

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/key/impl/ParameterManagerPrivateKeyStoreFactory.java
psasidhar: Oversight. Matched the non-global location endpoint with the documentation here. 

https://cloud.google.com/secret-manager/parameter-manager/docs/list-parameter-versions

Still working with Google Support for testing this with real values in us-central1.

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/key/impl/ParameterManagerPrivateKeyStoreFactory.java
psasidhar: done

libs/java/server_gcp_common/pom.xml
psasidhar: done

havetisyan (approved):

PR #3034, opened 2025-08-06
ui - switch from zms to msd for policy creation

ArtjomsPorss:
# Description
Switch Athenz UI from calling ZMS to calling MSD for microsegmentation policy creation and deletion.
- AddSegmentation now calls `createOrUpdateTransportPolicy` for both creation and editing of microsegmentation policy
- AddSegmentation - fixed validation to prevent submission of policy without an identifier or source service
- AddSegmentation - removed old calls involved in policy creation (`addRole, getRole, deleteRole, getPolicy, addPolicy, addAssertionProp, addAssertionConditions, editMicrosegmentation, getAssertionId`)
- AddSegmentation - refactored `onSubmit` and `onEditSubmit` to remove large chunk of duplicated code
- AddSegmentation - moved out direct call to `validateMicrosegmentationPolicy` to the thunk
- fixed some async issues
- RuleRow - `deleteTransportPolicy` calls MSD
- RuleRow and RulesList - fixed show error after submit/edit/delete of microsegmentation policy
- thunks/policies and roles - added force update flag for `getRoles` and `getPolicies` - when we need to force refresh upon deletion of microsegmentation policy, defaults to `false` for existing calls
- handlers/api - added server-side validation for put/update transport-policy call

# Contribution Checklist:


ui/src/components/microsegmentation/AddSegmentation.js
```diff
@@ -787,89 +655,16 @@ class AddSegmentation extends React.Component {
...
+                errorMessage: 'Identifier is required.',
+                saving: 'todo',
+            });
+            return 1;
+        }
         return 0;
     }
 
-    validateMicrosegmentationPolicy(
-        category,
```
ArtjomsPorss, line 794: moved out to microsegmentation thunks

ui/src/components/microsegmentation/AddSegmentation.js
```diff
@@ -888,310 +683,201 @@ class AddSegmentation extends React.Component {
...
             return;
         }
-
+        // add all members to one list
         this.addMember();
 
+        await this.submit(skipValidation);
+    }
+
+    async submit(skipValidation = false) {
```
ArtjomsPorss, line 742: moved common submission logic into `submit` function - previously was repeated by both `onSubmit` and `onEditSubmit`

ui/src/api.js
```diff
@@ -1196,6 +1196,32 @@ const Api = (req) => {
             });
         },
 
+        async createOrUpdateTransportPolicy(data, _csrf) {
```
chandrasekhar1996: why async + await instead of returning promise?

ArtjomsPorss: good point, they are redundant - removed `async/await`, just returning the Promise

ui/src/api.js
```diff
@@ -1196,6 +1196,30 @@ const Api = (req) => {
             });
         },
 
+        createOrUpdateTransportPolicy(data, _csrf) {
```
chandrasekhar1996, line 1199: lets follow the same pattern as other API calls. I don't think this will have the same effect.

```
return new Promise((resolve, reject) => {
                fetchr
                    .update('transport-policy')
                    .params(data)
                    .clientConfig({
                        headers: {
                            'x-csrf-token': _csrf,
                        },
                    })
                    .end((err, data) => {
                        if (err) {
                            reject(err);
                        } else {
                                resolve(data);
                        }
                    });
```

ui/src/server/handlers/api.js
```diff
@@ -970,6 +970,168 @@ Fetchr.registerService({
...
+            // Validate required fields before processing
+            if (
+                !params.domain ||
+                !params.service ||
+                !params.category ||
+                !params.pesList ||
+                !params.protocol ||
+                !params.identifier
+            ) {
+                throw new Error(
```
chandrasekhar1996: I don't think this validation is necessary but I think it should be returned 
`return callback(new Error(Missing required parameters: domain, service, category, instance list, protocol, or identifier));`

ui/src/components/microsegmentation/AddSegmentation.js
```diff
@@ -410,138 +406,7 @@ class AddSegmentation extends React.Component {
...
-                                    RequestUtils.fetcherErrorCheckHelper(err)
-                                );
-                            });
-                    } else {
-                        reject(RequestUtils.xhrErrorCheckHelper(err));
-                    }
-                });
-        });
-    }
+    createOrUpdateTransportPolicy(data, roleName, policyName) {}
```
chandrasekhar1996: should be deleted?

ui/src/components/microsegmentation/AddSegmentation.js
```diff
@@ -888,310 +683,201 @@ class AddSegmentation extends React.Component {
...
-                } else {
-                    this.setState({
-                        errorMessage: err,
-                        saving: 'todo',
-                    });
-                }
+        // if no changes - return
+        if (!this.hasChanges()) {
+            // if no changes - return
+            this.setState({
```
chandrasekhar1996, line 719: why throw an error here? let's just refresh without any changes if possible

ui/src/components/microsegmentation/AddSegmentation.js
```diff
@@ -888,310 +683,201 @@ class AddSegmentation extends React.Component {
...
+            await this.props.createOrUpdateTransportPolicy(
+                this.props.domain,
+                data,
+                this.props._csrf,
+                this.getRoleName(),
+                this.getPolicyName()
+            );
+            this.props.onSubmit();
+        } catch (err) {
+            this.props.showError(RequestUtils.fetcherErrorCheckHelper(err));
```
chandrasekhar1996: if we use showError here, is the add segmentation modal still visible?

ui/src/redux/thunks/microsegmentation.js
```diff
@@ -87,6 +109,63 @@ export const editMicrosegmentation =
...
+            );
+            dispatch(loadingSuccess('validateMicrosegmentationPolicy'));
+            return Promise.resolve();
+        } catch (e) {
+            dispatch(loadingFailed('validateMicrosegmentationPolicy'));
+            return Promise.reject(e);
+        }
+    };
+
 export const deleteTransportRule =
```
chandrasekhar1996: do we need this anymore?

ui/src/redux/thunks/microsegmentation.js
```diff
@@ -57,6 +62,23 @@ export const getInboundOutbound =
...
+            await dispatch(getInboundOutbound(domainName));
+            dispatch(loadingSuccess('createOrUpdateTransportPolicy'));
+            return Promise.resolve();
+        } catch (e) {
+            dispatch(loadingFailed('createOrUpdateTransportPolicy'));
+            return Promise.reject(e);
+        }
+    };
+
 export const editMicrosegmentation =
```
chandrasekhar1996: still required?

chandrasekhar1996 (changes requested):

ui/src/api.js
ArtjomsPorss: fetcher returns a promise when `.end()` is called without callback. Additional wrapping into `new Promise` is redundant, same goes for callback - we didn't do any additional logic in the callback.

ui/src/server/handlers/api.js
ArtjomsPorss: wrapped in callback as requested.

Added server side validation in case user attempts to bypass client side validation, and to make sure the data is there to setup request body further down

ui/src/components/microsegmentation/AddSegmentation.js
ArtjomsPorss: We are not really throwing an error here, just UX feedback to the user that they have not made any changes, displayed on the modal they are attempting to submit

ui/src/components/microsegmentation/AddSegmentation.js
ArtjomsPorss: not visible, the showError displays the error via alert at the top

ui/src/redux/thunks/microsegmentation.js
ArtjomsPorss: removed

ui/src/redux/thunks/microsegmentation.js
ArtjomsPorss: removed

ui/src/components/microsegmentation/AddSegmentation.js
chandrasekhar1996: I think it should be still visible. Let's say user used the same ID for a new policy then MSD server will return the error saying ID already in use. It would be better if the modal with user entry is still open as they can make small changes and try again instead of typing everything again

ArtjomsPorss: just double checked, upon error it does keep the modal open. 
Modal disappears only on successful call or upon change of state. Change of state causes re-render, and re-render doesn't happen if call failed.
So I made the change you suggest - it displays error on modal now

PR #3037, opened 2025-08-06
fix function return to be int to avoid casting

havetisyan:
# Description

the user input for the timeout is Integer. internally all max/min settings are integers but the function returns long which then causes the value to be cast back to int for the token. the proper fix is to return the value from the timeout function as int and not long

# Contribution Checklist:


psasidhar (approved):

kreeksec:
Thank for all @psasidhar @havetisyan ,

**Acknowledge contributed thanks to** @odaysec

PR #3039, opened 2025-08-12
Adding ConfigProviderParameterManager implementation for GCP

psasidhar:
# Description

# Contribution Checklist:

### DIFF
--- a/libs/java/server_gcp_common/pom.xml
+++ b/libs/java/server_gcp_common/pom.xml
@@ -38,6 +38,11 @@
       <version>${project.parent.version}</version>
     </dependency>
     <dependency>
+      <groupId>com.yahoo.athenz</groupId>
+      <artifactId>athenz-server-common</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
       <groupId>com.google.cloud</groupId>
       <artifactId>google-cloud-parametermanager</artifactId>
     </dependency>
--- a/libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/config/impl/ConfigProviderParameterManager.java
+++ b/libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/config/impl/ConfigProviderParameterManager.java
@@ -1,0 +1,179 @@
+/*
+ * Copyright The Athenz Authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.athenz.server.gcp.common.config.impl;
+
+import com.google.cloud.parametermanager.v1.*;
+import com.yahoo.athenz.common.server.util.config.ConfigEntry;
+import com.yahoo.athenz.common.server.util.config.providers.ConfigProvider;
+import io.athenz.server.gcp.common.utils.ParameterManagerClientHelper;
+import jakarta.annotation.Nullable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.IOException;
+import java.lang.invoke.MethodHandles;
+import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import static io.athenz.server.gcp.common.Consts.*;
+
+/**
+ * A configuration provider implementation that retrieves configuration parameters from Google Cloud's Parameter Manager.
+ * <p>
+ * This provider connects to the GCP Parameter Manager service to fetch configuration values stored as parameters.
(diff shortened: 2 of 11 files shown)


libs/java/server_gcp_common/pom.xml
```diff
@@ -52,5 +52,11 @@
       <version>${logback.server.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.yahoo.athenz</groupId>
+      <artifactId>athenz-server-common</artifactId>
+      <version>${project.parent.version}</version>
+      <scope>compile</scope>
```
havetisyan: why is this compile only?

libs/java/server_gcp_common/pom.xml
```diff
@@ -16,7 +16,7 @@
   <packaging>jar</packaging>
 
   <properties>
-    <code.coverage.min>1.00</code.coverage.min>
+    <code.coverage.min>0.97</code.coverage.min>
```
havetisyan: any reason we couldn't get 100%?

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/config/impl/ConfigProviderParameterManager.java
```diff
@@ -0,0 +1,176 @@
...
+
+    private static final Logger LOG = LoggerFactory.getLogger(MethodHandles.lookup().lookupClass());
+
+    @Override
+    public @Nullable ConfigSourceParameterManager tryToBuildConfigSource(String sourceDescription) {
+        if (!sourceDescription.startsWith(PROVIDER_DESCRIPTION_PREFIX)) {
+            return null;
+        }
+        String location = System.getProperty(ATHENZ_PROP_GCP_LOCATION, GLOBAL_LOCATION);
+        String projectId = System.getProperty(ATHENZ_PROP_GCP_PROJECT_ID, ATHENZ_GCP_PROJECT_ID_DEFAULT);
```
havetisyan: why do we have a default project id? that just sounds wrong.

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/utils/ParameterManagerClientHelper.java
```diff
@@ -0,0 +1,83 @@
...
+
+        return ParameterManagerClient.create(parameterManagerSettings);
+    }
+
+    public static boolean isGlobalLocation(String location) {
+        return GLOBAL_LOCATION.equalsIgnoreCase(location);
+    }
+
+    public static ParameterVersion getLatestParameterVersion(ParameterManagerClient client, ParameterName parameterName) {
+        LOG.info("getLatestParameterVersion: {}", parameterName);
```
havetisyan: why do we have this info call - doesn't this going to be logged every minute for all the parameters all the time?

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/utils/ParameterManagerClientHelper.java
```diff
@@ -0,0 +1,83 @@
...
+
+        // Iterate through all versions and find the latest one based on createTime.
+        return StreamSupport
+                .stream(listParameterVersionsPagedResponse.iterateAll().spliterator(), false)
+                .max(Comparator.comparing(ParameterVersion::getCreateTime, TIMESTAMP_COMPARATOR))
+                .orElse(null);
+    }
+
+    public static ParameterVersion getLatestParameterVersion(ParameterManagerClient client, String projectId, String location, String parameter) {
+        LOG.info("getLatestParameterVersion: {}", parameter);
```
havetisyan: same comment as above. this seems like it will just flood the server log with unnecessary details.

havetisyan (changes requested):

libs/java/server_gcp_common/pom.xml
psasidhar: removed it

libs/java/server_gcp_common/pom.xml
psasidhar: I think this is because "Const" class and the ParameterManagerClientHelper class that has only static methods.

consts
ParameterManagerClientHelper

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/config/impl/ConfigProviderParameterManager.java
psasidhar: I was trying to avoid GCP library throwing INVALID_ARGUMENT exception when the project id is empty, but handling this explicitly would make more sense. 

Added code to throw IllegalArgumentException, and so this will surface up during server startup. only when ConfigProviderParameterManager is configured and the sourceDescription matches and yet the project_id is not set.

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/utils/ParameterManagerClientHelper.java
psasidhar: removed it

libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/utils/ParameterManagerClientHelper.java
psasidhar: removed it

libs/java/server_gcp_common/pom.xml
psasidhar: updated it to get back to 100

havetisyan (approved):

PR #3087, opened 2025-10-03
fix: clear role members when set to trust role

TakuyaMatsu:
# Description
When creating or editing a Role with the Trust field set via SolutionTemplate, a role is created that has both Trust and role members. We confirmed the problem mainly in the following two patterns:

* Adding a SolutionTemplate that sets both RoleMember and Trust to the server and using it.
* Applying a SolutionTemplate that sets Trust to an already-created role that already has members.

This change fixes the issue so that it does not occur in the patterns described above.

# Contribution Checklist:

### DIFF
--- a/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
+++ b/servers/zms/src/main/java/com/yahoo/athenz/zms/DBService.java
@@ -5168,7 +5168,7 @@ public class DBService implements RolesProvider, DomainProvider {
                 firstEntry = auditLogSeparator(auditDetails, firstEntry);
                 auditDetails.append(" \"add-role\": ");
                 if (!processRole(con, originalRole, domainName, roleName, templateRole,
-                        admin, null, auditRef, true, auditDetails)) {
+                        admin, null, auditRef, StringUtil.isEmpty(templateRole.getTrust()), auditDetails)) {
                     return false;
                 }
 
@@ -5460,7 +5460,7 @@ public class DBService implements RolesProvider, DomainProvider {
 
         List<RoleMember> roleMembers = role.getRoleMembers();
         List<RoleMember> newMembers = new ArrayList<>();
-        if (roleMembers != null && !roleMembers.isEmpty()) {
+        if (StringUtil.isEmpty(templateRoleTrust) && roleMembers != null && !roleMembers.isEmpty()) {
             for (RoleMember roleMember : roleMembers) {
                 RoleMember newRoleMember = new RoleMember();
 
--- a/servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
+++ b/servers/zms/src/test/java/com/yahoo/athenz/zms/DBServiceTest.java
@@ -4086,8 +4086,8 @@ public class DBServiceTest {
     @Test
     public void testApplySolutionTemplateUpdateRoleByTrustRole() throws ServerResourceException {
 
-        String caller = "testApplySolutionTemplateRoleWithBothTrustAndMembers";
-        String domainName = "solutiontemplate-withtrustrole";
+        String caller = "testApplySolutionTemplateUpdateRoleByTrustRole";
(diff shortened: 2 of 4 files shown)


havetisyan:
I'm not sure if I follow this PR fix. It is completely invalid to specify a role with both trust and members. If you do that through the API, you should get back 400 request error.

Now, as an admin, you're forcing an incorrect state through solution template. So the server must not apply that template since it's invalid if it would have come through API.

As an admin, it's your responsibility to generate the valid template (this is why it's an admin operation and users are not allowed to create template).

The correct fix for this problem would be that the server must validate the solution template and not just assume that admin did the right thing and reject the template during startup and require the admin to fix the template to make it valid

TakuyaMatsu:
@havetisyan 

Thank you for the feedback.

I agree admins should ship valid templates. This change is meant as a guardrail to prevent human mistakes from creating a bad state.

### What I’m seeing right now

If I apply a SolutionTemplate that sets trust to a role that already has members (and the template doesn’t set roleMembers), the merged role ends up with both. getRole looks fine, but /sys/modified_domains shows both fields, e.g.:
```
{
  "name": "home.takuya:role.hoge",
  "modified": "2025-10-08T00:18:19.170Z",
  "roleMembers": [
    { "memberName": "user.takuya", "principalType": 1 }
  ],
  "trust": "hoge"
}
```
That breaks the “trust XOR members” rule and confuses anything reading modified_domains.

### Plan
* Validate: reject any template that defines both trust and roleMembers for the same role (same idea as the 400 the API would return).
* Merge rule: if a template sets trust on an existing role and doesn’t specify roleMembers, clear existing members so the result stays valid and the op is idempotent.
This PR implements the merge rule.

havetisyan:
@TakuyaMatsu your PR just masks invalid state and allows it to continue on so we shouldn't promote that.

1. If you specify a template that includes both role members and trusts - it should be rejected during startup
2. If you specify template with trust and apply it to an existing role which is already referenced in the template but has members, it should be rejected.

In both case the correct operation is to reject the invalid state. Otherwise, the merge may not be what the user wants and as such it should not be allowed

TakuyaMatsu:
@havetisyan 
> 1. If you specify a template that includes both role members and trusts - it should be rejected during startup

Right now we only validate JSON format at startup. Should I add logic here to stop startup if the template changes both trust and roleMembers?
https://github.com/AthenZ/athenz/blob/fa063a2d72f992da41420b0c95117d5c8a4e9858/servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java#L1293-L1317

> 2. If you specify template with trust and apply it to an existing role which is already referenced in the template but has members, it should be rejected.

My understanding is that SolutionTemplate is merge-oriented. In practice, when applying a template to an existing role, fields like members are merged rather than hard-replaced. If we reject applying a template that sets trust to a role that already has members, we’d be introducing an implicit rule—“you can’t set trust via SolutionTemplate on roles with existing members.” I’d prefer to avoid that.

havetisyan:
1. Yes, after we parse the solution templates, we should go through the objects and validate that the roles don't contain both members and trust attributes.
2. Solution Templates feature is merge oriented but it's not designed to handle different types of roles. If you have a delegated role and your template is for a regular member role, there is some disconnect that should not be allowed. If your role is a delegated role and you apply a template which makes the role as member role, you can't just ignore the members and return success since that assumes according to the template it was supposed to be a member role. Similarly, if you have a member role and your template tries to make the role as delegated, again, you can't just return success since that means you think the role is now a delegated role but in reality its not.

The server should check the role when modifying and it's not the same type as the template role that it's trying to apply, it should reject the operation.

gyakami:
@havetisyan
I'll take over this task. As you suggested, I plan to change the API to return an error. However, since this is a breaking change, I'd like to propose a phased approach.
First, I want to emit warning logs only, to prompt affected users to fix their configurations. Then, we can eventually transition to returning an error. Is this approach acceptable?
I've created PR #3159 for this. I would appreciate your review.

Issue #3100, opened 2025-10-11
Identity Assertion Authorization Grant support in ZTS


havetisyan:
As we're starting to look at using Athenz in some of Agentic AI use cases with the use of access tokens, seems like we need to implement token exchange as described in https://datatracker.ietf.org/doc/draft-ietf-oauth-identity-assertion-authz-grant. This would allow ZTS to interoperate with other Identity Providers that are issuing JAGs.

The additional extension would to add support in Athenz to issue JAGs as well. This would require a partial implementation of the token exchange rfc: https://www.rfc-editor.org/rfc/rfc8693.html as described in the aboved mention draft.

PR #3138, opened 2025-12-05
allow to return jag token with subset of scopes

havetisyan:
# Description

since the mcp client might ask for multiple roles, there is no need to reject the jag issue token if the user only has a subset of role access. instead we'll just return the subset. #3100 

# Contribution Checklist:


yosrixp (approved):
Looks good

PR #3146, opened 2025-12-09
x509 otel certificate refresh events in cert_refresher library.

balamanova:
# Description

Added OpenTelemetry metrics for X.509 certificate refresh events in `cert_refresher` library.

## Implementation follows SIA OTel pattern

This implementation mirrors the existing SIA (Go) OTel metrics pattern from `libs/go/sia/otel/metricset.go`:

### Metrics (aligned with SIA)

| This PR (Java) | SIA (Go) |
|----------------|----------|
| `athenz_cert_refresher.refresh.result_total{function,result}` | `sia.agent_command.result_total{function,result}` |
| `athenz_cert_refresher.service_cert.validity.remaining_secs{name}` | `sia.service_cert.validity.remaining_secs{cname}` |
| `athenz_cert_refresher.refresh.result_last_timestamp{function,result}` | *New - tracks when context was last updated* |

### Attributes (same as SIA)
- `function` - identifies the operation (e.g., "cert_refresh")
- `result` - "success" or "failure" (same values as SIA)
- `name` - certificate subject name (similar to SIA's `cname`)

### Configuration
- Disable with: `-Dathenz.cert_refresher.otel_disabled=true`
- Component name: `-Dathenz.cert_refresher.otel_component_name=my-service`

# Contribution Checklist:

### DIFF
--- a/libs/java/cert_refresher/src/test/java/com/oath/auth/KeyRefresherTest.java
+++ b/libs/java/cert_refresher/src/test/java/com/oath/auth/KeyRefresherTest.java
@@ -31,13 +31,12 @@ public class KeyRefresherTest {
 
     private final ClassLoader classLoader = this.getClass().getClassLoader();
     private final MessageDigest md = MessageDigest.getInstance("MD5");
-    MockedConstruction<OpenTelemetryCertReloadEventEmitter> mockedOtel = mockConstruction(OpenTelemetryCertReloadEventEmitter.class);
 
     public KeyRefresherTest() throws NoSuchAlgorithmException {
     }
 
     @Test
-    public void haveFilesBeenChangedTestFilesAlteredAndOtelEnabled() throws Exception {
+    public void haveFilesBeenChangedTestFilesAltered() throws Exception {
 
         TrustStore mockedTrustStore = Mockito.mock(TrustStore.class);
         TrustManagerProxy mockedTrustManagerProxy = Mockito.mock(TrustManagerProxy.class);
@@ -118,8 +117,8 @@ public class KeyRefresherTest {
     }
 
     @Test
-    public void scanForFileChangesTestNoChangesAndOtelEnabled() throws Exception {
-        System.setProperty("athenz.cert_refresher.otel_enabled", "true");
+    public void scanForFileChangesTestNoChanges() throws Exception {
+
         TrustStore mockedTrustStore = Mockito.mock(TrustStore.class);
         TrustManagerProxy mockedTrustManagerProxy = Mockito.mock(TrustManagerProxy.class);
         KeyManagerProxy mockedKeyManagerProxy = Mockito.mock(KeyManagerProxy.class);
@@ -137,16 +136,11 @@ public class KeyRefresherTest {
         keyRefresher.startup(100);
         Thread.sleep(3000);
         keyRefresher.shutdown();
-
-        assertEquals(mockedOtel.constructed().size(), 1);
-        OpenTelemetryCertReloadEventEmitter mockOtelMetric = mockedOtel.constructed().get(0);
-        verifyNoInteractions(mockOtelMetric);
-        System.clearProperty("athenz.cert_refresher.otel_enabled");
     }
 
     @Test
-    public void scanForFileChangesTestWithChangesAndOtelEnabled() throws Exception {
-        System.setProperty("athenz.cert_refresher.otel_enabled", "true");
+    public void scanForFileChangesTestWithChanges() throws Exception {
+
         TrustStore mockedTrustStore = Mockito.mock(TrustStore.class);
         TrustManagerProxy mockedTrustManagerProxy = Mockito.mock(TrustManagerProxy.class);
         KeyManagerProxy mockedKeyManagerProxy = Mockito.mock(KeyManagerProxy.class);
@@ -168,16 +162,10 @@ public class KeyRefresherTest {
         Thread.sleep(3000);
         assertTrue(listener.keyChanged);
         keyRefresher.shutdown();
-
-        assertEquals(mockedOtel.constructed().size(), 1);
-        OpenTelemetryCertReloadEventEmitter mockOtelMetric = mockedOtel.constructed().get(0);
-        verify(mockOtelMetric, atLeastOnce()).recordCertRefresh(anyString());
-        System.clearProperty("athenz.cert_refresher.otel_enabled");
     }
 
     @Test
-    public void testGenerateKeyRefresherFromCaCertAndOtelEnabled() throws Exception {
-        System.setProperty("athenz.cert_refresher.otel_enabled", "true");
+    public void testGenerateKeyRefresherFromCaCert() throws Exception {
 
         KeyRefresher keyRefresher = Utils.generateKeyRefresher("truststore.jks", "gdpr.aws.core.cert.pem",
                 "unit_test_gdpr.aws.core.key.pem");
@@ -201,10 +189,6 @@ public class KeyRefresherTest {
         Thread.sleep(500);
         assertNotNull(keyRefresher);
         keyRefresher.shutdown();
-        assertEquals(mockedOtel.constructed().size(), 5);
-        OpenTelemetryCertReloadEventEmitter mockOtelMetric = mockedOtel.constructed().get(0);
-        verify(mockOtelMetric, atLeastOnce()).recordCertRefresh(anyString());
-        System.clearProperty("athenz.cert_refresher.otel_enabled");
     }
 
     static class TestKeyRefresherListener implements KeyRefresherListener {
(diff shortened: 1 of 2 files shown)


libs/java/cert_refresher/src/main/java/com/oath/auth/KeyRefresher.java
```diff
@@ -136,14 +167,26 @@ && haveFilesBeenChanged(trustStore.getFilePath(), lastTrustManagerChecksum)) {
...
+                            if (otelMetric != null) {
+                                otelMetric.recordCertRefreshFailure(ex.getMessage());
+                            }
                         }
-                        //Signal key change event
-                        if (keyRefresherListener != null) {
-                            keyRefresherListener.onKeyChangeEvent();
+                    } else {
+                        if (otelMetric != null) {
+                            otelMetric.exportServiceCertMetric(athenzPublicCert);
```
psasidhar: with the default DEFAULT_RETRY_CHECK_FREQUENCY at 1hr, we get one metric from each client. this might be too often? when all customers use it on all their servers. 

since we are parsing the certificate inside exportServiceCertMetric to get the remaining validity, we can send the metric a less frequent intervals, if needed.

libs/java/cert_refresher/src/main/java/com/oath/auth/OpenTelemetryCertReloadEventEmitter.java
```diff
@@ -0,0 +1,153 @@
...
+        refreshResultCounter.add(1, attrs);
+        resultLastTimestampGauge.set(timestamp, attrs);
+    }
+
+    /**
+     * Export service certificate validity metric.
+     *
+     * @param certPath path to the certificate file
+     */
+    public void exportServiceCertMetric(String certPath) {
```
psasidhar: perhaps we need to rename the method to more neutral "exportCertMetric"? since people role certificates also in addition to service certificates?

libs/java/cert_refresher/src/main/java/com/oath/auth/KeyRefresher.java
balamanova: https://github.com/AthenZ/athenz/pull/3146/files

libs/java/cert_refresher/src/main/java/com/oath/auth/OpenTelemetryCertReloadEventEmitter.java
```diff
@@ -0,0 +1,153 @@
...
+        String result = success ? RESULT_SUCCESS : RESULT_FAILURE;
+        long timestamp = System.currentTimeMillis() / 1000;
+        
+        Attributes attrs = Attributes.builder()
+                .put("function", "cert_refresh")
+                .put("result", result)
+                .build();
+        
+        refreshResultCounter.add(1, attrs);
+        resultLastTimestampGauge.set(timestamp, attrs);
```
psasidhar: do we need to send timestamp explicity, won't that be automatically be available with refreshResultCounter?

balamanova: I'm using OTel Metrics API (counter + gauge), not OTel Events/Logs API. Prometheus is a pull-based time-series database.

Prometheus only stores scrape timestamps, not event timestamps. I can detect that a refresh happened using increase(), but the timestamp is approximate (within scrape interval). 

With this timestamp I save exact timestamp when the refresh occurred

libs/java/cert_refresher/src/main/java/com/oath/auth/KeyRefresher.java
```diff
@@ -102,6 +107,32 @@ public KeyRefresher(final String athenzPublicCert, final String athenzPrivateKey
...
+     * 
+     * Behavior:
+     * - If athenz.cert_refresher.otel_disabled=true: OTel disabled, return null
+     * - If athenz.cert_refresher.otel_disabled=false or not set: Initialize OTel metrics
+     * 
+     * @return OpenTelemetryCertReloadEventEmitter instance or null if disabled
+     */
+    private OpenTelemetryCertReloadEventEmitter initOtelMetrics() {
+        String otelDisabledProp = System.getProperty(PROP_OTEL_DISABLED);
+        boolean otelDisabled = "true".equalsIgnoreCase(otelDisabledProp);
```
havetisyan: The default behavior here should be false. We can't introduce new functionality which could cause problems and generate more metrics if nobody is looking at them.

havetisyan: Instead of using the "true" (or "false") , please specify a static string and use that. Alternatively you can convert to Boolean and use Boolean.TRUE/Boolean.FALSE

libs/java/cert_refresher/src/main/java/com/oath/auth/KeyRefresher.java
```diff
@@ -136,14 +167,22 @@ && haveFilesBeenChanged(trustStore.getFilePath(), lastTrustManagerChecksum)) {
...
                     if (keyFilesChanged) {
-                        
-                        keyManagerProxy.setKeyManager(Utils.getKeyManagers(athenzPublicCert, athenzPrivateKey));
-                        if (LOGGER.isDebugEnabled()) {
-                            LOGGER.debug("KeyRefresher detected changes. Reloaded Key managers");
-                        }
-                        //Signal key change event
-                        if (keyRefresherListener != null) {
-                            keyRefresherListener.onKeyChangeEvent();
+                        try {
```
havetisyan: why did we add a new try/catch block here? we already have one in place where we log any errors?

libs/java/cert_refresher/src/main/java/com/oath/auth/OpenTelemetryCertReloadEventEmitter.java
havetisyan: there is no need to generate a timestamp attribute. the metric already tells you that there is a failure in the given time period and the exact timestamp is not really needed and will not be used.

libs/java/cert_refresher/pom.xml
```diff
@@ -56,6 +56,12 @@
       <version>${guava.version}</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+      <version>${opentelemetry.version}</version>
+      <optional>true</optional>
```
havetisyan, line 68: have we verified that if we enable otel metrics and the jars are not available, the code works as expected without throwing any errors?

balamanova: nice point. It will throw NoClassDefFoundError. I fix with try catch, if the jar not available so it logs a warning message explaining the situation

libs/java/cert_refresher/src/main/java/com/oath/auth/KeyRefresher.java
```diff
@@ -102,6 +107,38 @@ public KeyRefresher(final String athenzPublicCert, final String athenzPrivateKey
...
+     * 
+     * Behavior:
+     * - If athenz.cert_refresher.otel_disabled=false: Initialize OTel metrics
+     * - If athenz.cert_refresher.otel_disabled=true or not set: OTel disabled, return null
+     * - If OTel JARs are not available on classpath: Log warning and return null
+     * 
+     * @return OpenTelemetryCertReloadEventEmitter instance or null if disabled or unavailable
+     */
+    private OpenTelemetryCertReloadEventEmitter initOtelMetrics() {
+        String otelDisabledProp = System.getProperty(PROP_OTEL_DISABLED);
```
havetisyan: this double negative setting doesn't look right. since we want the default to be false and only enable when configured (at list for this release), we should have the setting called

athenz.cert_refresher.otel_enabled and the default value would be false if not configured. then the check is more natural:

boolean otelEnabled = Boolean.parseBoolean(System.getProperty(PROP_OTEL_ENABLED, "false"));

libs/java/cert_refresher/src/main/java/com/oath/auth/OpenTelemetryCertReloadEventEmitter.java
```diff
@@ -0,0 +1,141 @@
...
+     *
+     * @param cert the X.509 certificate
+     */
+    public void exportCertMetric(X509Certificate cert) {
+        if (cert == null) {
+            LOGGER.warn("exportCertMetric: called with null cert");
+            return;
+        }
+        
+        String name = cert.getSubjectX500Principal().getName();
```
havetisyan: The name here is the full subject which may not work that well for athenz based certificates since we're including per region based details in the ou field. For example, if you get a certificate from us-west-2 but then us-west-2 was OOR, then you'll get a cert from another region and in grafana they won't be combined together.

I suggest in addition we only include the CN part of the subject (or maybe both?) which includes the athenz service identity which is always identical regardless of what region issued the certificate. Take a look at final Crypto.extractX500DnField(dn, BCStyle.CN) method in athenz-auth-core

havetisyan (approved):

havetisyan (approved):

Issue #3149, opened 2025-12-10
Go tests use a deprecated package and require manual test server configuration


frumioj:
In (for example) [libs/go/sia/aws/meta](libs/go/sia/aws/meta), the test servers use https://github.com/dimfeld/httptreemux which is no longer maintained. The way the testServer is currently written, the TCP listener must be started with a manual call to net.Listen which doesn't easily work in all scenarios (such as particular proxy environments, or where an HTTP server may already be running on a particular port). 

A better way would be to use the https://pkg.go.dev/net/http/httptest package and http handlers, all from the golang std packages.

frumioj:
Partially resolved by https://github.com/AthenZ/athenz/pull/3150

havetisyan:
Thanks for all the PRs

PR #3167, opened 2026-01-02
Httptest3

frumioj:
# Description
Further net/httptest package usage, and apparent bug in json map

# Contribution Checklist:
- [X] **The pull request does not introduce any breaking changes**
- [X] **I have read the [contribution guidelines](https://github.com/AthenZ/athenz/blob/master/CONTRIBUTING.md).**
- [X] **Create an issue and link to the pull request.**

### DIFF
--- a/provider/github/sia-actions/authn_test.go
+++ b/provider/github/sia-actions/authn_test.go
@@ -46,8 +46,8 @@ func TestGetOIDCToken(t *testing.T) {
 	ts := startHttpServer(validToken, http.StatusOK)
 	defer ts.Close()
 
-t.Setenv("ACTIONS_ID_TOKEN_REQUEST_URL", ts.URL+"/oidc?type=jwt")
-t.Setenv("ACTIONS_ID_TOKEN_REQUEST_TOKEN", "test-token")
+	t.Setenv("ACTIONS_ID_TOKEN_REQUEST_URL", ts.URL+"/oidc?type=jwt")
+	t.Setenv("ACTIONS_ID_TOKEN_REQUEST_TOKEN", "test-token")
 
 	_, claims, err := GetOIDCToken("https://athenz.io")
 	assert.Nil(t, err)


frumioj:
Should fully resolve #3149

provider/harness/sia-harness/authn.go
```diff
@@ -115,7 +115,8 @@ func GetOIDCToken(audience, harnessUrl string) (string, map[string]interface{},
 	}
 
 	signatureAlgorithms := []jose.SignatureAlgorithm{jose.RS256, jose.RS384, jose.RS512, jose.PS256, jose.PS384, jose.PS512, jose.ES256, jose.ES384, jose.ES512, jose.EdDSA}
-	oidcToken := jsonData["data"].(string)
+
+	oidcToken := jsonData["value"].(string)
```
havetisyan: why was this field name changed? harness returns the oidc token in the data field and not value so the test should be updated accordingly.

havetisyan (changes requested):

provider/harness/sia-harness/authn.go
frumioj: When I ran the test, it failed, so I inserted a printf and noted that it was looking for a value that was not in the JSON. So I changed it to use what was actually there.

havetisyan: The unit tests was incorrectly modified. It originally had on line 37 in athenz_test.go:

		io.WriteString(w, "{\"data\": \""+token+"\"}")

but this was changed to:

		io.WriteString(w, "{\"value\": \""+token+"\"}")

so both authn.go and authn_test.go need to be reverted back to field called "data"

frumioj: OK - apologies!

frumioj: Reverted and retested. Works now.

Issue #3054, opened 2025-08-27
jwk_uri property is now mandatory in v1.12.1, breaking backward compatibility


ysknkd:
It appears that since version v1.12.1, the jwk_uri property has become a mandatory requirement. As a result, AuthZpeClient can no longer be used if this property is not configured.

To the best of my knowledge, in v1.11.x, this property was not required for AuthZpeClient to function. This change in behavior introduces a backward compatibility issue.

https://github.com/AthenZ/athenz/blob/346e2250ac4e09d94d69564de08e6be0b4c6fd05/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java#L202-L205

Would it be possible to consider modifying the logic? For instance, with a change like the following:

```java
    public static void initializeAccessTokenSignKeyResolver() {
        final String keyPath = System.getProperty(ZpeConsts.ZPE_PROP_JWK_PRIVATE_KEY_PATH);
        final String certPath = System.getProperty(ZpeConsts.ZPE_PROP_JWK_X509_CERT_PATH);
        SSLContext sslContext = null;
        if (keyPath != null && !keyPath.isEmpty() && certPath != null && !certPath.isEmpty()) {
            try {
                KeyRefresher keyRefresher = Utils.generateKeyRefresher(null, certPath, keyPath);
                keyRefresher.startup();
                sslContext = Utils.buildSSLContext(keyRefresher.getKeyManagerProxy(),
                        keyRefresher.getTrustManagerProxy());
            } catch (Exception ex) {
                LOG.error("Unable to initialize key refresher: {}", ex.getMessage());
            }
        }

        String serverUrl = System.getProperty(ZpeConsts.ZPE_PROP_JWK_URI);
        if (serverUrl != null || !serverUrl.isEmpty()) {
            setAccessTokenSignKeyResolver(serverUrl, sslContext);
        }
    }
```

havetisyan:
Yes, that's part of the 1.12.x release which is a major release and as such there will be breaking changes. We also removed quite a few deprecated methods as well part of that part major release. For jjwt processing, we moved from jjwt to nimbus.

The problem with jwt validation without the jwk-uri is that whenever you rotate the key used to sign token, all clients break unless they have access to the corresponding public key and with the use of jwk uri, the library can just automatically fetch the key from ZTS and not fail. 

What is your use case that you don't want this? Don't you guys periodically rotate your keys?

ysknkd:
[...]

PR #3198, opened 2026-02-10
Allow wrapper-controlled init to fix jwk_uri backward compatibility (Issue #3054)

sh1myama:
# Description
## Context
* In v1.12.x, jwk_uri became mandatory and the check runs in a static initializer, which prevents wrappers from setting system properties before AuthZpeClient loads. (Issue #3054)

## Changes
* Move access-token resolver setup out of the static initializer into AuthZpeClient.init() so callers can set properties before initialization.
* Keep the existing jwk_uri requirement for access-token validation, but make its initialization explicit and controllable.
* Add support for an optional athenz.zpe.jwk_proxy_uri and pass it into setAccessTokenSignKeyResolver to enable proxied JWK fetches.

## Outcome
* Restores upgrade path for Role Token users by avoiding premature jwk_uri validation during class load, while preserving access-token behavior when init() is called.

# Contribution Checklist:

### DIFF
--- a/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
+++ b/clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
@@ -155,7 +155,7 @@ public class AuthZpeClient {
     }
     
     static {
-
+        
         // load public keys
 
         setPublicKeyStoreFactoryClass(System.getProperty(ZpeConsts.ZPE_PROP_PUBLIC_KEY_CLASS, ZPE_PKEY_CLASS));
@@ -175,7 +175,12 @@ public class AuthZpeClient {
         // load the x509 issuers
         
         setX509CAIssuers(System.getProperty(ZpeConsts.ZPE_PROP_X509_CA_ISSUERS));
+    }
 
+    public static void init() {
+        if (LOG.isDebugEnabled()) {
+            LOG.debug("Init: load the ZPE");
+        }
         // initialize the access token signing key resolver
 
         initializeAccessTokenSignKeyResolver();
@@ -183,12 +188,6 @@ public class AuthZpeClient {
         // save the last zts api call time, and the allowed interval between api calls
 
         setMillisBetweenZtsCalls(Long.parseLong(System.getProperty(ZPE_PROP_MILLIS_BETWEEN_ZTS_CALLS, Long.toString(30 * 1000 * 60))));
-    }
-
-    public static void init() {
-        if (LOG.isDebugEnabled()) {
-            LOG.debug("Init: load the ZPE");
-        }
     }
 
     public static void close() {
@@ -203,6 +202,7 @@ public class AuthZpeClient {
         if (serverUrl == null || serverUrl.isEmpty()) {
             throw new IllegalArgumentException("Missing required property: " + ZpeConsts.ZPE_PROP_JWK_URI);
         }
+        String proxyUrl = System.getProperty(ZpeConsts.ZPE_PROP_JWK_PROXY_URI);
 
         final String keyPath = System.getProperty(ZpeConsts.ZPE_PROP_JWK_PRIVATE_KEY_PATH);
         final String certPath = System.getProperty(ZpeConsts.ZPE_PROP_JWK_X509_CERT_PATH);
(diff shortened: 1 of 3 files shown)


havetisyan:
@sh1myama the AuthZpeClient.init() was never required for ZPE functionality. So by moving the actual keyresolver initialization into that function will break most zpe users, so not the correct solution.

I'll check in the next couple of days in more detail and will propose how to address your use case.

sh1myama:
@havetisyan -san

> @sh1myama the AuthZpeClient.init() was never required for ZPE functionality. So by moving the actual keyresolver initialization into that function will break most zpe users, so not the correct solution.
> 
> I'll check in the next couple of days in more detail and will propose how to address your use case.

Thank you for checking.

In the following Java decentralized access control sample, the documentation says to initialize the ZPE Client in init when performing authorization checks, so I understood that calling init is mandatory.
https://github.com/AthenZ/athenz/blob/master/docs/example_java_decentralized_access.md#authorization-checks

Even if we apply this change and modify it so that initializeAccessTokenSignKeyResolver is called from init, my understanding is that the behavior will not change as long as the implementation is written to always call init itself.

Thank you in advance.

havetisyan:
@sh1myama here is what changes we need to make:

1. introduce a static variable tokenSignKeyResolverInitialized that key resolver is initialized with default value false
2. update the initializeAccessTokenSignKeyResolver to take a boolean argument whether or not return exceptions or just return without any errors.
3. in the static initializer call the method with false value. if the variables are not set, it will just return without any exceptions, otherwise it will set the initializer and set the tokenSignKeyResolverInitialized to true.
4. In the init function, introduce a new system property athenz.zpe.jwk_uri_required with a default value of true. If the value is false, then do nothing and return. If the value is true (default value), it will check the value for tokenSignKeyResolverInitialized. If it's false then it will call the initializeAccessTokenSignKeyResolver method with the true argument so that it will return an exception if there are any errors (current behavior).

this would allow your applications to set the either set the jwks_uri to complete the sign key resolver or jwks_uri_required to false in your application and ignore calling the method completely.

sh1myama:
@havetisyan -san
Sorry for the late reply. Thank you for your suggestion. I will review the content you provided.

sh1myama:
@havetisyan -san
I have revised the code based on the advice you provided.
Since athenz.zpe.jwk_uri_required is not necessary in our environment, I have not added it.
Do you have any concerns about these changes?

clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
```diff
@@ -178,17 +179,22 @@ public String toString() {
...
         // initialize the access token signing key resolver
 
-        initializeAccessTokenSignKeyResolver();
+        initializeAccessTokenSignKeyResolver(false);
 
         // save the last zts api call time, and the allowed interval between api calls
-
-        setMillisBetweenZtsCalls(Long.parseLong(System.getProperty(ZPE_PROP_MILLIS_BETWEEN_ZTS_CALLS, Long.toString(30 * 1000 * 60))));
+        if (tokenSignKeyResolverInitialized) {
+            setMillisBetweenZtsCalls(Long.parseLong(System.getProperty(ZPE_PROP_MILLIS_BETWEEN_ZTS_CALLS, Long.toString(30 * 1000 * 60))));
```
havetisyan: let's move this setMillisBetweenZtsCalls as the last line in the initializeAccessTokenSignKeyResolver method. this way we don't have to repeat the call twice once here and then once again in init method

havetisyan (changes requested):

clients/java/zpe/src/main/java/com/yahoo/athenz/zpe/AuthZpeClient.java
sh1myama: Thank you for the review. Sorry for the late reply. I have fixed the points you pointed out.
https://github.com/AthenZ/athenz/pull/3198/changes/11f2a300c1ada6c0ead8febb77920547d011a131

havetisyan:
@sh1myama looks like you need to add some test cases for the new changes:

[INFO] --- jacoco:0.8.14:check (jacoco-check) @ athenz-zpe-java-client ---
[INFO] Loading execution data file /home/runner/work/athenz/athenz/clients/java/zpe/target/jacoco.exec
[INFO] Analyzed bundle 'athenz-zpe-java-client' with 29 classes
Warning:  Rule violated for bundle athenz-zpe-java-client: lines covered ratio is 0.9062, but expected minimum is 0.9075

sh1myama:
My apologies. I added the test code and confirmed in my local environment that the coverage requirements are satisfied.

havetisyan (approved):

PR #3210, opened 2026-02-27
Defining snapshop api for MSD

psasidhar:
# Description

Introducing a Snapshot api to take a named snapshot of the TransportPolicy at a given time. 

# Contribution Checklist:


clients/go/msd/client.go
```diff
@@ -1102,6 +1102,184 @@ func (client MSDClient) PostKubernetesNetworkPolicyRequest(domainName DomainName
...
+			return data, err
+		}
+		json.Unmarshal(contentBytes, &errobj)
+		if errobj.Code == 0 {
+			errobj.Code = resp.StatusCode
+		}
+		if errobj.Message == "" {
+			errobj.Message = string(contentBytes)
+		}
+		return data, errobj
```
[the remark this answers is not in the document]
psasidhar: autogenerated code

core/msd/src/main/java/com/yahoo/athenz/msd/MSDSchema.java
```diff
@@ -1047,6 +1079,112 @@ private static Schema build() {
...
+            .exception("CONFLICT", "ResourceError", "")
+
+            .exception("FORBIDDEN", "ResourceError", "")
+
+            .exception("NOT_FOUND", "ResourceError", "")
+
+            .exception("TOO_MANY_REQUESTS", "ResourceError", "")
+
+            .exception("UNAUTHORIZED", "ResourceError", "")
+;
```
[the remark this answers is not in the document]
psasidhar: autogenerated

clients/go/msd/client.go
```diff
@@ -1102,6 +1102,184 @@ func (client MSDClient) PostKubernetesNetworkPolicyRequest(domainName DomainName
 	}
 }
 
+func (client MSDClient) CreateTransportPolicySnapshot(domainName DomainName, serviceName EntityName, snapshot *TransportPolicySnapshotRequest, resourceOwner string) (*TransportPolicySnapshotMetadata, error) {
```
yosrixp, line 1105: Can you add snapshot name validation in RDL? simple one that doesn't allow snapshot name with spaces and long names. 
i.e 

- test snapshot (invalid)
- {name}{2,50} (valid) and beyond that limit invalid. 
- stable (valid)
- canary-v1 (valid)

basically we can have regexp similar to what we have under Names.tdl

clients/go/msd/client.go
```diff
@@ -1102,6 +1102,184 @@ func (client MSDClient) PostKubernetesNetworkPolicyRequest(domainName DomainName
...
+			errobj.Code = resp.StatusCode
+		}
+		if errobj.Message == "" {
+			errobj.Message = string(contentBytes)
+		}
+		return data, errobj
+	}
+}
+
+func (client MSDClient) GetTransportPolicySnapshot(domainName DomainName, serviceName EntityName, snapshotName EntityName, matchingTag string) (*TransportPolicySnapshot, string, error) {
```
yosrixp, line 1215: I believe we need to update the TransportPolicy RDL to include snapshot query param as well and keep it optional to be backward compatible.

clients/go/msd/client.go
psasidhar: the name is defined as EntityName, which has standard regex we use in other places. We can revise it later if needed.

clients/go/msd/client.go
psasidhar: if needed, we can consider that.

PR #3219, opened 2026-03-04
Add server-aws-common to ZMS/ZTS assembly tarballs

### LINKED ISSUE #3214: server_aws_common not included in ZMS/ZTS assembly tarballs
## Background

The 1.11-to-1.12 migration guide states: "Make sure athenz-server-aws-common module along with all of its dependencies is deployed as part of your application."

However, the ZMS and ZTS assembly tarballs (`assembly/zms/target/athenz-zms-*-bin.tar.gz` and `assembly/zts/target/athenz-zts-*-bin.tar.gz`) do not include `athenz-server-aws-common` or its transitive dependencies (AWS SDK v2 rds, sts, ssm, s3, kms, etc.). Neither `assembly/zms/pom.xml` nor `assembly/zts/pom.xml` declares `athenz-server-aws-common` as a dependency, and `servers/zms`/`servers/zts` also don't depend on it.

This means the factory classes (`AWSObjectStoreFactory`, `AWSCertRecordStoreFactory`) are loaded dynamically via config properties, but the JARs they live in are not in the classpath when using the assembly tarballs.

## What we had to do

We build Athenz from source and use the assembly tarballs to deploy ZMS/ZTS on EC2. To make the AWS factories work with v1.12.x, we had to:

1. Run `mvn dependency:copy-dependencies` on the `server_aws_common` module after the main build
2. Manually copy `athenz-server-aws-common-*.jar` and all resolved transitive JARs into the ZMS/ZTS lib directories

## Question

Is this the intended workflow for users deploying from the assembly tarballs? Or would it make sense to include `athenz-server-aws-common` (and its deps) in the assembly when building for AWS deployments?

gjoranv:
## Summary

Adds `athenz-server-aws-common` as a dependency in both `assembly/zms/pom.xml` and `assembly/zts/pom.xml`. This ensures the AWS factory classes (`AWSObjectStoreFactory`, `AWSCertRecordStoreFactory`) and their transitive dependencies (AWS SDK v2) are included in the assembly tarballs.

Without this, users deploying from the assembly tarballs as described in the [ZMS setup docs](https://athenz.github.io/athenz/setup_zms/) get `ClassNotFoundException` at runtime when using the AWS factory classes.

Fixes #3214

## Trade-off

This adds the AWS SDK v2 JARs to both tarballs unconditionally, increasing their size. For non-AWS deployments these JARs are unnecessary. If that's a concern, an alternative approach would be a Maven profile (e.g., `-Paws`) that conditionally includes the dependency. Happy to implement that instead if preferred.

### DIFF
--- a/assembly/zms/pom.xml
+++ b/assembly/zms/pom.xml
@@ -38,6 +38,11 @@
       <version>${project.parent.version}</version>
     </dependency>
     <dependency>
+      <groupId>${project.groupId}</groupId>
+      <artifactId>athenz-server-aws-common</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
       <groupId>ch.qos.logback</groupId>
       <artifactId>logback-classic</artifactId>
       <version>${logback.server.version}</version>
--- a/assembly/zts/pom.xml
+++ b/assembly/zts/pom.xml
@@ -39,6 +39,11 @@
     </dependency>
     <dependency>
       <groupId>${project.groupId}</groupId>
+      <artifactId>athenz-server-aws-common</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>${project.groupId}</groupId>
       <artifactId>athenz-instance-provider</artifactId>
       <version>${project.parent.version}</version>
     </dependency>


havetisyan:
@gjoranv can you update both of yours to be signed. the failed DCO check has the details. thanks.

gjoranv:
Done, thanks!

Issue #3233, opened 2026-03-11
ZTS Token Exchange: include SPIFFE ID claim in issued ID Token (JWT SVID support)


ctyano:
## Description
  Today, ZTS supports OAuth 2.0 Token Exchange for issuing an OIDC-style id_token. It would be very useful if the id_token returned by ZTS could include a SPIFFE ID (e.g., spiffe://...) so that the exchanged ID Token can be directly used as a JWT SVID credential by workloads.

  I'm aware Athenz can already issue JWT SVIDs via the Copper/Argos procedure, but Token Exchange is increasingly the common integration point for external identity providers and workload identity flows, and having ZTS return a JWT SVID-compatible ID Token would reduce operational complexity (fewer moving parts, simpler client bootstrap, and a single standardized exchange endpoint).

## Background / motivation
  We have services that already perform OAuth Token Exchange to obtain an ID Token for downstream authn/authz. If ZTS includes the SPIFFE ID in that issued ID Token, those services can reuse the same token as a JWT SVID for SPIFFE-aware systems, improving interoperability while keeping the identity anchored to ZTS-issued policy decisions.

## Request

  - For Token Exchange requests that result in issued_token_type=urn:ietf:params:oauth:token-type:id_token, include a SPIFFE ID claim in the returned ID Token when it can be derived from    the subject token (or otherwise reliably determined).
  - Clarify the claim name/format in docs (e.g., spiffe claim with spiffe://... URI).

  Acceptance criteria

  - When exchanging into an ID Token, the response id_token contains a SPIFFE ID claim (when available).
  - Unit/integration tests cover presence/absence cases and validate no behavior regression.

## Proposed contribution
  If this approach is acceptable, I can open a PR implementing the claim propagation + tests + documentation updates.

havetisyan:
The approach is fine. However, ZTS does not generate any spiffe uri values - it only has the validators to validate them. So in this mode, do we expect that the client would provide the spiffe value or what? if the spiffe value contains the namespace component - e.g. `spiffe://<trust-doman>/ns/<namespace>/sa/<service-identity>` then how would server know what is the namespace value?

ctyano:
Thanks for the review. I think we can address it cleanly with an enhancement PR that binds the source credential's identifier (the authenticated Principal name) to the Principal name in the SPIFFE ID of the exchanged token, at the Authority class where we validate the Token Exchange subject token.

[...]

PR #3254, opened 2026-03-22
Add SPIFFE claim support to ZTS ID token exchange

ctyano:
# Description

#3233

# Contribution Checklist:

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -2697,6 +2697,27 @@ public class ZTSImpl implements ZTSHandler {
         }
     }
 
+    String extractSpiffeIdFromToken(final OAuth2Token token) {
+        if (token == null) {
+            return null;
+        }
+
+        final Object spiffeClaim = token.getClaim(IdToken.CLAIM_SPIFFE);
+        if (spiffeClaim != null) {
+            final String spiffeId = spiffeClaim.toString();
+            if (!StringUtil.isEmpty(spiffeId) && spiffeId.startsWith(ZTSConsts.ZTS_CERT_SPIFFE_URI)) {
+                return spiffeId;
+            }
+        }
+
+        final String subject = token.getSubject();
+        if (!StringUtil.isEmpty(subject) && subject.startsWith(ZTSConsts.ZTS_CERT_SPIFFE_URI)) {
+            return subject;
+        }
+
+        return null;
+    }
+
     AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx, Principal principal,
             AccessTokenRequest accessTokenRequest, final String principalDomain, final String caller) {
 
@@ -2787,6 +2808,11 @@ public class ZTSImpl implements ZTSHandler {
         accessToken.setIssuer(issuerResolver.getAccessTokenIssuer(ctx.request(), accessTokenRequest.isUseOpenIDIssuer()));
         accessToken.setScope(new ArrayList<>(roles));
         accessToken.setPrincipalIssuer(principal.getIssuerIdentity());
+
+        final String spiffeId = extractSpiffeIdFromToken(subjectToken);
+        if (spiffeId != null) {
+            accessToken.setCustomClaim(IdToken.CLAIM_SPIFFE, spiffeId);
+        }
 
         // if we have a certificate used for mTLS authentication then
         // we're going to bind the certificate to the access token
@@ -2940,6 +2966,11 @@ public class ZTSImpl implements ZTSHandler {
         accessToken.setIssuer(issuerResolver.getAccessTokenIssuer(ctx.request(), accessTokenRequest.isUseOpenIDIssuer()));
         accessToken.setScope(new ArrayList<>(roles));
         accessToken.setPrincipalIssuer(principal.getIssuerIdentity());
+
+        final String spiffeId = extractSpiffeIdFromToken(subjectToken);
+        if (spiffeId != null) {
+            accessToken.setCustomClaim(IdToken.CLAIM_SPIFFE, spiffeId);
+        }
 
         // include the act claim in our response. we're going to use
         // the act claim from the original token and then add our new
@@ -3099,6 +3130,11 @@ public class ZTSImpl implements ZTSHandler {
         idToken.setIssueTime(iat);
         idToken.setAuthTime(iat);
         idToken.setPrincipalIssuer(principal.getIssuerIdentity());
+
+        final String spiffeId = extractSpiffeIdFromToken(subjectToken);
+        if (spiffeId != null) {
+            idToken.setSpiffe(spiffeId);
+        }
 
         // for user principals we're going to use the default 1 hour while for
         // service principals 12 hours as the max timeout, unless the client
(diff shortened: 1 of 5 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2864,6 +2857,15 @@ AccessTokenResponse processAccessTokenDelegationRequest(ResourceContext ctx, Pri
                     caller, requestDomainName, principalDomain);
         }
 
+        // now let's verify that our principal is authorized to carry
+        // out token delegation from source to target domain
+
+        final String resource = sourceDomainName + ":" + requestDomainName;
+        if (!authorizer.access(ZTSConsts.ZTS_ACTION_TOKEN_SOURCE_EXCHANGE, resource, principal, null)) {
```
havetisyan: why do we need this authorization check? 

First, we've already verified at the beginnig of the method that princpal name is the same as the subject in the actor token.
Next, when validating the actor subject token in the token request parser, we've already verified that the subject token includes the mayAct claim and the value matches to the subject of the actor token. So the authorization has already been granted and verified so not sure we're adding yet another authorization check here?

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2652,6 +2620,27 @@ AccessTokenResponse processAccessTokenExchangeRequest(ResourceContext ctx, Princ
         }
     }
 
+    String extractSpiffeIdFromToken(final OAuth2Token token) {
+        if (token == null) {
+            return null;
+        }
+
+        final Object spiffeClaim = token.getClaim(IdToken.CLAIM_SPIFFE);
```
havetisyan, line 2707: so the claim itself is called spiffe but that's only valid for ZTS itself. If we're doing an exchange from another identity provider, then most likely this isn't going to work. 

if it's a spiffe svid, then spiffie uri is in the sub claim and that's supported.

So if it's any other claim used by that identity provider then it's not supported. This may not be a problem if we're only dealing with ZTS tokens or spiffe svids but something to be aware of. maybe we can add comments about this limitation

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3029,14 +3035,13 @@ AccessTokenResponse processIdTokenExchangeRequest(ResourceContext ctx, Principal
...
                 false, principalDomain, caller);
 
-        // make sure our principal is authorized to request a jag token
+        // make sure our principal is authorized to request an id token
         // exchange for the given roles
 
-        Principal subjectPrincipal = createPrincipalForName(subjectIdentity);
         for (String idTokenGroup : idTokenGroups) {
-            if (!authorizer.access(ZTSConsts.ZTS_ACTION_ID_TOKEN_EXCHANGE, idTokenGroup, subjectPrincipal, null)) {
+            if (!authorizer.access(ZTSConsts.ZTS_ACTION_ID_TOKEN_EXCHANGE, idTokenGroup, principal, null)) {
```
havetisyan: why was this changed? it's an incorrect change. 

We're checking if the principal in the subject token is authorized for the change not the principal who requested the exchange (which could be some proxy token exchange broker).

a few lines above, we verify that the principal in the request is authorized to request the token exchange on behalf of the subject principal and here we must verify that subject principal is authorized for the requested role

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3020,6 +3059,14 @@ AccessTokenResponse processIdTokenExchangeRequest(ResourceContext ctx, Principal
         idToken.setIssueTime(iat);
         idToken.setAuthTime(iat);
 
+        final String spiffeId = extractSpiffeIdFromToken(subjectToken);
+        if (spiffeId != null) {
+            // If we're copying SPIFFE claim into the requested ID token, verify it maps
+            // to the authenticated principal to prevent mismatched SPIFFE identities.
+            verifySpiffeIdMatchesAthenzPrincipal(spiffeId, principalName, caller, domainName, principalDomain);
```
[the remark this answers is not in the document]
ctyano: Fixed.
The SPIFFE ID is extracted from the subject token, so it should be verified against the token subject, not the caller of the exchange request. I updated the verification target from principalName to subjectIdentity.
I also updated testIdTokenExchangeSuccessWithSpiffeClaim so the SPIFFE ID corresponds to the subject identity.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
ctyano: Thank you for the clarification. I removed the additional authorization check.
As you pointed out, the actor token subject is already checked against the request principal, and the subject token mayAct claim is validated by the token request parser, so this extra token_source_exchange authorization check is not needed in this flow.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
ctyano: I also reverted this check back to using subjectPrincipal. The caller authorization to perform the exchange on behalf of the subject principal was handled by the preceding principal-level exchange check.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
ctyano: I added a comment to clarify the current limitation.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2710,6 +2731,11 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
         accessToken.setIssuer(issuerResolver.getAccessTokenIssuer(ctx.request(), accessTokenRequest.isUseOpenIDIssuer()));
         accessToken.setScope(new ArrayList<>(roles));
 
+        final String spiffeId = extractSpiffeIdFromToken(subjectToken);
+        if (spiffeId != null) {
+            accessToken.setCustomClaim(IdToken.CLAIM_SPIFFE, spiffeId);
+        }
```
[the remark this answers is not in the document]
ctyano: Fixed.
Before propagating the spiffe claim into the issued access token, the code now verifies that the SPIFFE ID corresponds to subjectToken.getSubject() using verifySpiffeIdMatchesAthenzPrincipal.
I also added a negative test for the mismatched SPIFFE ID case.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2862,6 +2897,11 @@ AccessTokenResponse processAccessTokenDelegationRequest(ResourceContext ctx, Pri
...
```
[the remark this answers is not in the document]
ctyano: Fixed.
The delegation flow now performs the same SPIFFE validation before copying the spiffe claim into the issued access token. The SPIFFE ID is verified against subjectToken.getSubject().
I also added a negative test for the mismatched SPIFFE ID case.

havetisyan (approved):

PR #3267, opened 2026-03-29
Add BSD support to libs/go/sia/util

tuftedocelot:
# Description
Generifies https://github.com/AthenZ/athenz/pull/3093 to support all BSDs

# Contribution Checklist:


tuftedocelot:
@olgeni I'm curious if this PR works for you on FreeBSD?

dgl:
@tuftedocelot this actually makes this file attempt to be compiled on all platforms (`go help buildconstraint` has the rules), basically `bsd` isn't a `GOOS` so this is always compiled. Calling it this for clarity makes sense, but then it needs a build constraint like:

```
//go:build freebsd || openbsd || netbsd || dragonfly
```

...except if you're adding build constraints the only difference is windows, so to avoid duplication this should be a `_windows.go` and a `_unix.go` with a build constraint on `unix` (note unix can't appear in the filename either, but this way does avoid hardcoding the list of Unix-like platforms).

Also looks like this is missing a Signed-off-by.

PR #3326, opened 2026-04-24
Support parallel domain fetch and upload in CloudZmsSyncer

gyakami:
# Description
Parallelize domain fetch and upload in CloudZmsSyncer to improve sync performance in environments with a large number of domains. Separate thread pools are used for updates and refreshes respectively. Thread counts are configurable, defaulting to 1 (same as existing behavior) for backward compatibility.

- Add parallel domain fetch and upload support to `CloudZmsSyncer` using configurable thread pools for both domain updates and domain refreshes
- Introduce `DomainUploadTask` class to track async upload tasks with `Future`-based result handling
- Convert domain counters from `int` to `AtomicInteger` for thread-safe concurrent access
- Separate refresh count limiting from `shouldRefreshDomain()` into the caller, making the method simpler and the limit enforcement more explicit
- Add robust error handling: `InterruptedException` restores the interrupt flag and marks remaining tasks as failed; `ExecutionException` records individual domain failures
- New configuration parameters: `domain_update_fetch_threads` and `domain_refresh_fetch_threads` (default: 1, preserving single-threaded behavior)

# Contribution Checklist:

### DIFF
--- a/libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
+++ b/libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
@@ -362,6 +362,8 @@ public class CloudZmsSyncer {
                     updateExecutor.shutdown();
                     refreshExecutor.shutdown();
                 }
+                awaitTermination(updateExecutor);
+                awaitTermination(refreshExecutor);
             }
 
         } catch (Exception ex) {
@@ -393,6 +395,19 @@ public class CloudZmsSyncer {
                 " : number-domain-deleted-failures: " + getNumDomainsDeletedFailed();
         LOG.info(sb);
         return retStatus;
+    }
+
+    static void awaitTermination(ExecutorService executor) {
+        try {
+            if (!executor.awaitTermination(30, TimeUnit.SECONDS)) {
+                LOG.warn("executor did not terminate within timeout, forcing shutdown");
+                executor.shutdownNow();
+            }
+        } catch (InterruptedException e) {
+            LOG.warn("interrupted while awaiting executor termination");
+            executor.shutdownNow();
+            Thread.currentThread().interrupt();
+        }
     }
 
     static ThreadFactory newNamedThreadFactory(String prefix) {
--- a/libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
+++ b/libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
@@ -34,6 +34,8 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.nio.file.StandardCopyOption;
 import java.util.*;
+import java.util.concurrent.*;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import static io.athenz.syncer.common.zms.Config.SYNC_CFG_PARAM_STATE_PATH;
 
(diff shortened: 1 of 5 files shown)


libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
```diff
@@ -300,25 +327,41 @@ boolean syncDomains(final Map<String, DomainState> stateMap) throws Exception {
...
                     LOG.debug("no change so no upload of domain: {}", domainName);
+                    processedDomains.add(domainState);
+                }
+            }
+
+            LOG.info("processing domain updates with {} thread(s), count: {}", domainUpdateFetchThreads, updateDomains.size());
+            LOG.info("processing domain refreshes with {} thread(s), count: {}", domainRefreshFetchThreads, refreshDomains.size());
+
+            ExecutorService updateExecutor = Executors.newFixedThreadPool(domainUpdateFetchThreads);
+            ExecutorService refreshExecutor = Executors.newFixedThreadPool(domainRefreshFetchThreads);
```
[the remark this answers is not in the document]
gyakami: CloudZmsSyncer is a one-shot object, so there’s no point in reusing the thread pool. I’ve also implemented thread naming.
https://github.com/AthenZ/athenz/pull/3326/changes/07fc71fa9ee2ee064d945142975f7b03ab5f6caa

libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
```diff
@@ -300,25 +327,41 @@ boolean syncDomains(final Map<String, DomainState> stateMap) throws Exception {
...
+                retStatus = collectProcessedDomains(refreshTasks) && retStatus;
+            } finally {
+                if (Thread.currentThread().isInterrupted()) {
+                    updateExecutor.shutdownNow();
+                    refreshExecutor.shutdownNow();
+                } else {
+                    updateExecutor.shutdown();
+                    refreshExecutor.shutdown();
                 }
             }
```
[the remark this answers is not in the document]
gyakami: Added awaitTermination with a 30-second timeout after shutdown() / shutdownNow() to ensure full resource cleanup. Note that collectProcessedDomains already blocks on all future.get() calls, so by the time finally runs, all tasks have completed — but the explicit await is a good defensive measure.
https://github.com/AthenZ/athenz/pull/3326/commits/70f707b3b0817583cf2435915eecd3a85f3849d9

libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
```diff
@@ -300,25 +327,45 @@ boolean syncDomains(final Map<String, DomainState> stateMap) throws Exception {
...
-                    if (refreshDom) {
-                        ++numDomainsRefreshed;
-                    }
+                boolean timeToRefresh = shouldRefreshDomain(domainState, now, domainRefreshTimeout);
+
+                if (uploadDom) {
+                    updateDomains.add(domainName);
+                } else if (timeToRefresh && refreshDomains.size() < allowedRefreshCount) {
+                    refreshDomains.add(domainName);
+                    numDomainsRefreshed.incrementAndGet();
```
havetisyan: this seems to be in the incorrect place. we haven't yet refreshed the domain, it was just added to the list to be refreshed. so we should update this counter after we have successfully refreshed it. maybe add a flag to uploadDomain method that is set by the executor calling it.

libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
```diff
@@ -300,25 +327,45 @@ boolean syncDomains(final Map<String, DomainState> stateMap) throws Exception {
...
+            LOG.info("processing domain refreshes with {} thread(s), count: {}", domainRefreshFetchThreads, refreshDomains.size());
+
+            ExecutorService updateExecutor = Executors.newFixedThreadPool(domainUpdateFetchThreads,
+                    newNamedThreadFactory("domain-update"));
+            ExecutorService refreshExecutor = Executors.newFixedThreadPool(domainRefreshFetchThreads,
+                    newNamedThreadFactory("domain-refresh"));
+            try {
+                List<DomainUploadTask> updateTasks = submitUploadTasks(updateExecutor, updateDomains);
+                List<DomainUploadTask> refreshTasks = submitUploadTasks(refreshExecutor, refreshDomains);
+                retStatus = collectProcessedDomains(updateTasks) && retStatus;
```
havetisyan: somewhat confusing to have "& retStatus" at the end here since it's not really used or set before, so let's remove it.

havetisyan (changes requested):
@gyakami thanks for the PR. nice improvement to speed up the sync process if you have large number of domains and you always want to sync them up within a minute if you're running with a cron job. just a couple of minor comments to address.

libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
gyakami: I appreciate you pointing that out. I've fixed it by adding an isRefresh flag to uploadDomain.
https://github.com/AthenZ/athenz/pull/3326/changes/496de07444cd1d340efe135ffda20aff465553c6

libs/java/syncer_common/src/main/java/io/athenz/syncer/common/zms/CloudZmsSyncer.java
gyakami: Fixed this one too.
https://github.com/AthenZ/athenz/pull/3326/commits/081fd4d886ecd2082d2801a5f9adf25c99e66a91

havetisyan (approved):

PR #3373, opened 2026-05-22
ui - add tf icons to tf-managed roles, services, policies; show zms-cli command.

ArtjomsPorss:
# Description
## Terraform icons in Athenz UI
This is a change to Athenz UI where for TF-owned roles, services, policies a TF icon will is displayed.
The icon is shown in these places:

| Location                                                   | When the icon appears |
|------------------------------------------------------------| --------------------- |
| **Roles** (domain role list)                               | Role metadata is owned (`objectOwner` or `metaOwner`) |
| **Role members** (add-member dialog title and role header) | Role membership is owned (`membersOwner`) — not on individual member rows |
| **Policies** (domain policy list and role policy tab)      | Policy object or assertions are owned (`objectOwner` or `assertionsOwner`) |
| **Services** (service list and service header)             | Service object is owned (`objectOwner`) |

## Athenz UI provides zms-cli command to use

If you try to change a Terraform-managed resource through the UI (delete role, add/remove member, edit role settings, add/delete policy assertion, delete service, delete public key, and similar), the operation **fails with a warning** instead of applying the change. The dialog then shows a **copy-ready `zms-cli` command** that can perform the same action while bypassing resource-owner enforcement.

When ownership blocks the action, the UI generates a command appropriate to what you attempted, including (non-exhaustive):

| UI action                                                 | Typical `zms-cli` subcommand                                              |
|-----------------------------------------------------------|---------------------------------------------------------------------------|
| Delete role                                               | `delete-role`                                                             |
| Update role settings (description, review flags, etc.)    | `set-role-description`, `set-role-member-expiry`, or related `set-role-*` |
| Add / remove role member                                  | `add-member`, `add-temporary-member`, or `delete-member`                  |
| Delete policy                                             | `delete-policy`                                                           |
| Add / delete policy assertion (policy or role-policy tab) | `add-assertion` / `delete-assertion`                                      |
| Delete service                                            | `delete-service`                                                          |
| Add / Delete service public key                           | `add-public-key` / `delete-public-key`                                    |

### Recommended Workflow for the users
1. **Identify** managed resources via the Terraform icon before editing.
2. **Change** ownership, membership, policies, and services in Terraform; plan and apply through your standard process.
3. **Avoid** UI or `zms-cli -r ignore` except for documented emergencies.
4. **Reconcile** after any emergency CLI use by updating Terraform code/state and applying so Athenz and Terraform match again.

Screenshot 2026-05-22 at 15 38 07

> Icon beside TF-owned role name, when viewing specific role / Service:

Screenshot 2026-05-22 at 15 39 01

> When attempting to make change to TF-owner resources, a corresponding zms-cli command is generated and presented to the user with a warning:

Screenshot 2026-05-22 at 15 40 14


ui/src/components/utils/zmsCliCommands.js
```diff
@@ -0,0 +1,388 @@
...
+export const ZMS_CLI_RESOURCE_OWNER_FLAG = 'ignore';
+
+/** Shell-safe token for displayed zms-cli lines (bash/zsh copy-paste). */
+export function shellQuote(str) {
+    if (str === undefined || str === null) {
+        return '';
+    }
+    const s = String(str);
+    if (/[^a-zA-Z0-9./_-]/.test(s)) {
+        return '"' + s.replace(/"/g, '\\"') + '"';
```
tsultanov00: This quoting is not shell-safe. Values like audit refs can come from user-entered justification in audit-enabled domains, and double quotes still allow command substitution in bash/zsh. For example, an ownership-blocked role delete could generate `zms-cli -d my.domain -r ignore -a "$(touch /tmp/athenz-poc)" delete-role tf-role`, which runs `touch /tmp/athenz-poc` when pasted. Use single-quote shell escaping instead, e.g. wrap in '...' and replace embedded ' with '\''.

ui/src/components/microsegmentation/AddStaticInstances.js
```diff
@@ -127,8 +134,23 @@ class AddStaticInstances extends React.Component {
...
-                    errorMessage: RequestUtils.xhrErrorCheckHelper(err),
+                    errorMessage: errMsg,
+                    resourceOwnershipCliCommand:
+                        resolveResourceOwnershipCliOnError(
+                            isServiceResourceHostsManaged(
+                                this.props.resourceOwnership
+                            ),
+                            err,
+                            () =>
+                                cliAddServiceHost(
```
tsultanov00: This fallback command does not match the failed operation. The expectation is that the suggested zms-cli command should reproduce the UI action, which is adding an MSD static workload via putStaticWorkload with both type and name. In reality, add-service-host updates the ZMS service identity hosts list and drops the static workload type, so following this hint mutates a different backend object and won’t add the static instance the user attempted to create.

It might be worth adding a note so that users are clear about the effects of the command.

ui/src/components/utils/zmsCliCommands.js
ArtjomsPorss: fixed

ui/src/components/microsegmentation/AddStaticInstances.js
ArtjomsPorss: reverted the static instances checks and zms-cli - out of current scope

ui/src/config/default-config.js
```diff
@@ -251,6 +286,7 @@ const config = {
         serverCipherSuites:
             'TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256:TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256:TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384:TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384:TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256:TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256:TLS_DHE_RSA_WITH_AES_128_GCM_SHA256:TLS_DHE_RSA_WITH_AES_256_GCM_SHA384:TLS_DHE_RSA_WITH_CHACHA20_POLY1305_SHA256',
         callCloudSSO: true,
+        allowedDevOrigins: ['localhost'],
```
tsultanov00, line 289: fyi this will be overwritten by allowedDevOrigins defined in extended config which has a different value.

ArtjomsPorss: yes, developers can override it via extended-config.js if they decide to connect to locally running instance via different hostname

tsultanov00 (approved):

Issue #3385, opened 2026-06-10
ZMS Service Identity: allow scoped clientId self-update for OAuth client metadata


ctyano:
## Description

Athenz should support OAuth 2.1 Dynamic Client Registration (DCR) for MCP-required ID-JAG flows by allowing an OAuth Client to update the `clientId` field on its own Athenz `ServiceIdentity`.

Today, `clientId` is handled as a service system metadata attribute. Updating it requires `update` access to `sys.auth:meta.service.clientid.{domain}`, which effectively limits the operation to Athenz system administrators such as members of `sys.auth:role.admin` in the default setup.

This issue proposes adding a backwards-compatible, opt-in path where the OAuth Client/service identity can update its own `clientId` when the caller is a member of the target service identity domain's `admin` role.

## Background / motivation

MCP requires OAuth 2.1 DCR support. For ID-JAG issuance, Athenz needs a way for OAuth Clients to register or rotate their OAuth client identifier and keep the corresponding Athenz `ServiceIdentity.clientId` value in sync.

The current model requires Athenz system-level metadata privileges to set `clientId`. That does not scale well for DCR because domain administrators should be able to manage OAuth client registration for service identities in their own domains without requiring `sys.auth:role.admin` membership.

## Request

- Allow an OAuth Client/service principal to update only the `clientId` field of its own `ServiceIdentity`.
- Require the caller to be a member of `{domain}:role.admin`, where `{domain}` is the domain that owns the target `ServiceIdentity`.
- Keep the existing `sys.auth` system metadata authorization path working unchanged.
- Add opt-in DCR enablement flags following the pattern used by `autoDeleteTenantAssumeRoleAssertions` in #3027:
  - a server-side property, for example `athenz.zms.oauth_dcr_client_id_update_enabled=false`
  - a domain metadata flag, for example `oauthDcrClientIdUpdateEnabled=false`
- Preserve backwards compatibility by keeping the new behavior disabled by default.

## Acceptance criteria

- Existing `clientId` updates by authorized `sys.auth` administrators continue to work unchanged.
- When the server flag or target domain flag is disabled, domain-admin self-update of `clientId` is rejected.
- When both flags are enabled, a service identity can update its own `clientId` only if it is a member of the target domain's `admin` role.
- A caller cannot update another service identity's `clientId`.
- A caller cannot use this path to update other service system metadata attributes.
- A caller that is not a target-domain admin is rejected.
[...]

PR #3386, opened 2026-06-10
Allow scoped clientId self-update for OAuth client metadata

ctyano:
# Description

#3385

# Contribution Checklist:


servers/zms/src/test/java/com/yahoo/athenz/zms/ZMSServiceIdentityTest.java
```diff
@@ -1104,6 +1105,92 @@ public void testPutServiceIdentitySystemMeta() {
...
+
+        zmsImpl.clientIdSelfUpdate = false;
+        RsrcCtxWrapper sysAdminCtx = zmsTestInitializer.generateServiceSysAdmin("putServiceIdentitySystemMeta",
+                "service-system-meta-dcr-sysadmin", "adminsvc");
+        ServiceIdentitySystemMeta sysAdminMeta = new ServiceIdentitySystemMeta().setClientId("sys-auth-client-id");
+        zmsImpl.putServiceIdentitySystemMeta(sysAdminCtx, domainName, serviceName, "clientid", auditRef,
+                sysAdminMeta);
+        serviceRes = zmsImpl.getServiceIdentity(ctx, domainName, serviceName);
+        assertEquals(serviceRes.getClientId(), "sys-auth-client-id");
+    }
```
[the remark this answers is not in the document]
ctyano: I added the domain deletion at the end of testPutServiceIdentitySystemMetaClientIdSelfUpdate.

servers/zms/schema/zms_server.sql
```diff
@@ -58,6 +58,7 @@ CREATE TABLE IF NOT EXISTS `zms_server`.`domain` (
   `slack_channel` VARCHAR(80) NOT NULL DEFAULT '',
   `on_call` VARCHAR(128) NOT NULL DEFAULT '',
   `auto_delete_tenant_assume_role_assertions` TINYINT(1) NOT NULL DEFAULT 0,
+  `client_id_self_update` TINYINT(1) NOT NULL DEFAULT 0,
```
havetisyan, line 61: the PR is missing the updated mwb file. see https://github.com/AthenZ/athenz/blob/master/servers/zms/schema/schema_update.md on how to update and generate the updated zms_server.sql file

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -1768,7 +1772,8 @@ public Domain postTopLevelDomain(ResourceContext ctx, String auditRef, String re
                 .setSshCertSignerKeyId(detail.getSshCertSignerKeyId())
                 .setSlackChannel(detail.getSlackChannel())
                 .setOnCall(detail.getOnCall())
-                .setAutoDeleteTenantAssumeRoleAssertions(detail.autoDeleteTenantAssumeRoleAssertions);
+                .setAutoDeleteTenantAssumeRoleAssertions(detail.autoDeleteTenantAssumeRoleAssertions)
+                .setClientIdSelfUpdate(detail.clientIdSelfUpdate);
```
havetisyan: we should utilize getClientIdSelfUpdate() instead of getting the field directly. can you also fix the line above as you're making this change.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2085,7 +2090,8 @@ public Domain postUserDomain(ResourceContext ctx, String name, String auditRef,
                 .setContacts(detail.getContacts())
                 .setEnvironment(detail.getEnvironment())
                 .setSlackChannel(detail.getSlackChannel())
-                .setOnCall(detail.getOnCall());
+                .setOnCall(detail.getOnCall())
+                .setClientIdSelfUpdate(detail.clientIdSelfUpdate);
```
havetisyan: we should utilize getClientIdSelfUpdate() instead of getting the field directly.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2192,7 +2198,8 @@ public Domain postSubDomain(ResourceContext ctx, String parent, String auditRef,
                 .setX509CertSignerKeyId(detail.getX509CertSignerKeyId())
                 .setSshCertSignerKeyId(detail.getSshCertSignerKeyId())
                 .setSlackChannel(detail.getSlackChannel())
-                .setOnCall(detail.getOnCall());
+                .setOnCall(detail.getOnCall())
+                .setClientIdSelfUpdate(detail.clientIdSelfUpdate);
```
havetisyan: we should utilize getClientIdSelfUpdate() instead of getting the field directly.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -7126,6 +7134,49 @@ public void putServiceIdentitySystemMeta(ResourceContext ctx, String domainName,
...
+    void verifyAuthorizedServiceIdentitySystemMetaOperation(Principal principal, String domainName,
+            String serviceName, String attribute, ServiceIdentitySystemMeta meta, String caller) {
+
+        final String systemResource = SYS_AUTH + ":meta.service." + attribute + "." + domainName;
+        if (isAllowedSystemAccess(principal, ZMSConsts.ACTION_UPDATE, systemResource)) {
+            return;
+        }
+
+        AthenzDomain domain = getAthenzDomain(domainName, false, true);
+        if (isAllowedClientIdSelfUpdate(principal, domain, serviceName, attribute, meta)) {
```
havetisyan: this method here looks confusing. if I'm checking an authorization for updating some system meta attribute why am i calling a method called isAllowedClientIdSelfUpdate when my attribute got nothing to do with client id self update. rather than check if the attribute is the meta client id is within that method, bring it out to clarify the use. e.g.

// special authorization check of the meta client id - only
// if the option is enabled at the server level
if (clientIdSelfUpdate && ZMSConsts.SYSTEM_META_CLIENT_ID.equals(attribute)) {
  isAllowedClientIdSelfUpdate();
}

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -7126,6 +7134,49 @@ public void putServiceIdentitySystemMeta(ResourceContext ctx, String domainName,
...
 
+    void verifyAuthorizedServiceIdentitySystemMetaOperation(Principal principal, String domainName,
+            String serviceName, String attribute, ServiceIdentitySystemMeta meta, String caller) {
+
+        final String systemResource = SYS_AUTH + ":meta.service." + attribute + "." + domainName;
+        if (isAllowedSystemAccess(principal, ZMSConsts.ACTION_UPDATE, systemResource)) {
+            return;
+        }
+
+        AthenzDomain domain = getAthenzDomain(domainName, false, true);
```
havetisyan: we're wasting this getAthenzDomain for every system meta call where the attribute is not the client id. must be moved within the isAllowedClientIdSelfUpdate method.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -7126,6 +7134,49 @@ public void putServiceIdentitySystemMeta(ResourceContext ctx, String domainName,
...
+            return false;
+        }
+
+        if (principal == null || domain == null || domain.getDomain() == null || meta == null ||
+                !Boolean.TRUE.equals(domain.getDomain().getClientIdSelfUpdate())) {
+            return false;
+        }
+
+        final String clientId = meta.getClientId();
+        if (StringUtil.isEmpty(clientId)) {
```
havetisyan: if the clientId is empty and authz return false - how does the client remove the client id if it wants to do that? is that not possible?

havetisyan (changes requested):

servers/zms/schema/zms_server.sql
ctyano: Added the matching zms_server.mwb update for the new client_id_self_update column. The generated SQL and migration file were already included.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
ctyano: Updated this to use `getClientIdSelfUpdate()` (and also `getAutoDeleteTenantAssumeRoleAssertions()`) instead of direct field access.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
ctyano: Updated this to use `getClientIdSelfUpdate()` instead of direct field access.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
ctyano: Updated this to use `getClientIdSelfUpdate()` instead of direct field access.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
ctyano: Moved the domain lookup into `isAllowedClientIdSelfUpdate()` after the cheap checks. Other system meta updates than `clientid` no longer load the domain through this path.

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
ctyano: Thanks for pointing this out. I kept this unchanged for now because this PR scopes delegated authorization to setting a non-empty target client ID via `service.<service>.clientid.<clientid>`.
Allowing an empty/null value would extend the behavior from scoped client ID assignment to self-service removal, which seems slightly broader than the original purpose of this PR.
Existing `sys.auth` system access can still clear the field through the same system meta API.
Do you think self-service removal should be included in this PR, or could we handle it separately with an explicit permission model and tests?

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
ctyano: @havetisyan Thanks for pointing this out. Moved the server flag and clientid attribute check to the caller, so `isAllowedClientIdSelfUpdate()` is only invoked for the client ID self-update path.

havetisyan (approved):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
havetisyan: it's fine. in the future, just add comments in the code so the intention is clear

PR #3407, opened 2026-07-04
Feat: Access Token with Multiple Domain Supported for Full Token Exchange

mlajkim:
# Background

Athenz Access Tokens currently work well when all requested role scopes belong to the same domain as the token audience. This has been enough for most existing use cases, as token exchange was not a primary flow and was only recently introduced to support ID-JAG.

With AI agent and MCP-style flows, token exchange becomes a core part of the request chain. For example:

```text
 home.mlajkim.claude -> mcp-hub.api-mcp -> api.api
```

In this flow, the first Access Token is issued by `home.mlajkim.claude` for the MCP Hub audience, but the caller also needs to explicitly delegate API access that the MCP server may exchange later.

## Problem

Today, the original subject token cannot cleanly carry scopes from multiple domains. This blocks flows where the token audience is `mcp-hub`, but the delegated downstream permission is in another domain, such as `api:role.docs-getter`.

Relying solely on the subject's role membership for token exchange is not an option. We strictly cannot allow downstream services to exchange a token for scopes not present in the original token, as this would result in permission escalation. The original requester must explicitly define which permissions are passed into the exchange chain.

Therefore, the first access token must carry the exact role scopes that are allowed to be exchanged later, even when those scopes belong to domains other than the audience domain.

## Changes

This PR adds multi-domain role scope support for Athenz Access Tokens.

When an Access Token is requested with an explicit audience:

* Scopes in the audience domain continue to be emitted as short role names.
* Scopes outside the audience domain are emitted as fully qualified role scopes.
* Token exchange validates requested roles against the scopes present in the subject token.

Example:

`audience=mcp-hub`
`scope=mcp-hub:role.api-mcp-accessor api:role.docs-getter`

can produce a token with:

```json
{
  "aud": "mcp-hub",
  "scp": [
     "api-mcp-accessor", // implies it belongs to the aud "mcp-hub"
     "api:role.docs-getter" // not yet need for the aud "mcp-hub", but the "mcp-hub" will need it to exchange
  ]
}

```

This keeps existing single-domain behavior backward compatible while allowing explicit cross-domain delegation for token exchange.

## Minor notes

- AccessToken's scopes can contain up to 20 domains by default, but customizable w/ `athenz.zts.access_token_max_domains`
- Requires `aud` only when the requested scopes contain 2 or more different domains
- Existing single-domain request is still supported, such as `scope=demo:role.readers`
- Domain-wide scope still supported, such as `scope=api:domain` or `scope=api:domain mcp-hub:role.accessors & audience=mcp-hub`
- The audience domain is required to appear among the scope domains. This mean the following example is **NOT** acceptable:
  - ~`{ aud": "mcp-hub", "scp": ["api:role.reader"] }`~

# Contribution Checklist:


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2856,47 +2854,119 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
+        for (String subjectScope : subjectScopes) {
+            if (subjectScope.startsWith(roleScopePrefix)) {
+                requestedRoles.add(subjectScope.substring(roleScopePrefix.length()));
+            } else if (!subjectScope.contains(OAuthTokenScope.OBJECT_ROLE) &&
+                    (StringUtil.isEmpty(sourceDomainName) || requestDomainName.equals(sourceDomainName))) {
+                requestedRoles.add(subjectScope);
+            }
+        }
+        return requestedRoles.toArray(new String[0]);
+    }
```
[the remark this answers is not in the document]
mlajkim: Handled in https://github.com/AthenZ/athenz/pull/3407/commits/3f885b0f7f936f3adbf55d8d9d74b37d9cb6e783

servers/zts/src/main/java/com/yahoo/athenz/zts/token/OAuthTokenScope.java
```diff
@@ -239,7 +239,7 @@ void addScopeDomain(final String scopeDomainName, final String scope, boolean en
             throw error("empty domain name", scope);
         }
         final String domainName = getDomainName();
-        if (enforceMaxDomainCheck && domainName != null && !scopeDomainName.equals(domainName)) {
+        if (enforceMaxDomainCheck && maxDomains == 1 && domainName != null && !scopeDomainName.equals(domainName)) {
              throw error("Multiple domains in scope", scope);
         }
```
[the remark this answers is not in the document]
mlajkim: Handled in https://github.com/AthenZ/athenz/pull/3407/commits/3f885b0f7f936f3adbf55d8d9d74b37d9cb6e783

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2856,47 +2854,119 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
             return null;
         }
 
-        String[] scopeRoles = scopeClaim.toString().split(" ");
+        final String sourceDomainName = subjectToken.getAudience();
+        Set<String> scopeRoles = new HashSet<>(Arrays.asList(scopeClaim.toString().split("\\s+")));
+        if (scopeRoles.isEmpty() || scopeRoles.contains("")) {
+            LOGGER.error("subject token does not contain any scope roles");
+            return null;
+        }
```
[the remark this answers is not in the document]
mlajkim: Handled in https://github.com/AthenZ/athenz/pull/3407/commits/3f885b0f7f936f3adbf55d8d9d74b37d9cb6e783

servers/zts/src/test/java/com/yahoo/athenz/zts/ZTSImplAccessTokenTest.java
```diff
@@ -4505,6 +4577,21 @@ public void testTokenExchangeRequestedRolesNullScopeInRequest() {
...
 
+    @Test
+    public void testTokenExchangeRequestedRolesTrimSubjectTokenScope() {
+        AccessTokenRequest accessTokenRequest = Mockito.mock(AccessTokenRequest.class);
+        OAuth2Token subjectToken = Mockito.mock(OAuth2Token.class);
+        String requestDomainName = "testdomain";
+
+        Mockito.when(subjectToken.getClaim(AccessToken.CLAIM_SCOPE_STD)).thenReturn(" admin  writer ");
+        Mockito.when(accessTokenRequest.getScope()).thenReturn("testdomain:role.admin");
+
```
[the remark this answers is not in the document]
mlajkim: Handled in https://github.com/AthenZ/athenz/pull/3407/changes/f1c274c3d67986a6f211d67206b5b662220303b9

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2856,47 +2854,123 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
+        for (String subjectScope : subjectScopes) {
+            if (AccessTokenScope.OBJECT_OPENID.equals(subjectScope)) {
+                continue;
+            }
+            if (subjectScope.startsWith(roleScopePrefix)) {
+                requestedRoles.add(subjectScope.substring(roleScopePrefix.length()));
+            } else if (!subjectScope.contains(":") &&
+                    (StringUtil.isEmpty(sourceDomainName) || requestDomainName.equals(sourceDomainName))) {
+                requestedRoles.add(subjectScope);
+            }
```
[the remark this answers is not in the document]
mlajkim: Handled in https://github.com/AthenZ/athenz/pull/3407/changes/f1c274c3d67986a6f211d67206b5b662220303b9

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2856,47 +2854,123 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
+            final String requestDomainName, String[] requestedRoles) {
+
+        for (String requestedRole : requestedRoles) {
+            if (subjectScopes.contains(requestDomainName + OAuthTokenScope.OBJECT_ROLE + requestedRole)) {
+                continue;
+            }
+            if ((StringUtil.isEmpty(sourceDomainName) || requestDomainName.equals(sourceDomainName)) &&
+                    subjectScopes.contains(requestedRole)) {
+                continue;
+            }
```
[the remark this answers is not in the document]
mlajkim: Handled in https://github.com/AthenZ/athenz/pull/3407/changes/f1c274c3d67986a6f211d67206b5b662220303b9

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2856,47 +2854,123 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
             return null;
         }
 
-        String[] scopeRoles = scopeClaim.toString().split(" ");
+        final String sourceDomainName = subjectToken.getAudience();
+        final String scopeClaimStr = scopeClaim.toString().trim();
+        if (scopeClaimStr.isEmpty()) {
+            LOGGER.error("subject token does not contain any scope roles");
+            return null;
+        }
```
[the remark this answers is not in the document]
mlajkim: Handled in https://github.com/AthenZ/athenz/pull/3407/changes/f1c274c3d67986a6f211d67206b5b662220303b9

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2856,47 +2854,125 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
+            final String requestDomainName) {
+
+        Set<String> requestedRoles = new HashSet<>();
+        final String roleScopePrefix = requestDomainName + OAuthTokenScope.OBJECT_ROLE;
+        for (String subjectScope : subjectScopes) {
+            if (AccessTokenScope.OBJECT_OPENID.equals(subjectScope)) {
+                continue;
+            }
+            if (subjectScope.startsWith(roleScopePrefix)) {
+                requestedRoles.add(subjectScope.substring(roleScopePrefix.length()));
```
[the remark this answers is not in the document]
mlajkim: https://github.com/AthenZ/athenz/pull/3407/commits/f4c20ea3a34a680921fd48a395ae7e0c46b1d146

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -104,13 +104,14 @@
...
 import java.util.*;
-import java.util.UUID;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 import com.fasterxml.jackson.core.StreamReadConstraints;
 
 import javax.crypto.SecretKey;
+import javax.xml.validation.Schema;
+import javax.xml.validation.Validator;
```
mlajkim:
```suggestion
```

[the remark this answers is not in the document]
mlajkim: https://github.com/AthenZ/athenz/pull/3407/commits/6af9695073f0b87bd404c8f07afa5c22e2825b4b

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
```diff
@@ -375,6 +375,19 @@ public static byte[] convertSignatureFromP1363ToDERFormat(byte[] signature, fina
...
+            if (seq.size() != 2) {
+                return false;
+            }
+            return ((ASN1Integer) seq.getObjectAt(0)).getValue().signum() <= 0 ||
+                    ((ASN1Integer) seq.getObjectAt(1)).getValue().signum() <= 0;
+        } catch (Exception ex) {
+            return false;
+        }
+    }
+
```
mlajkim:
```suggestion
```

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
```diff
@@ -498,6 +511,9 @@ public static boolean verify(String message, PublicKey key, String signature,
         try {
             byte [] sig = ybase64Decode(signature);
             String signatureAlgorithm = getSignatureAlgorithm(key.getAlgorithm(), digestAlgorithm);
+            if (signatureAlgorithm.endsWith(ECDSA) && hasInvalidECDSASignatureValue(sig)) {
+                return false;
+            }
```
mlajkim:
```suggestion
```

libs/java/auth_core/src/main/java/com/yahoo/athenz/auth/util/Crypto.java
```diff
@@ -542,6 +558,9 @@ public static boolean verify(byte[] message, PublicKey key, byte[] signature,
                                  String digestAlgorithm) throws CryptoException {
         try {
             String signatureAlgorithm = getSignatureAlgorithm(key.getAlgorithm(), digestAlgorithm);
+            if (signatureAlgorithm.endsWith(ECDSA) && hasInvalidECDSASignatureValue(signature)) {
+                return false;
+            }
```
mlajkim:
```suggestion
```

libs/java/auth_core/src/test/java/com/yahoo/athenz/auth/util/CryptoTest.java
```diff
@@ -1505,10 +1505,12 @@ public void testNullSignatureValidation() throws NoSuchAlgorithmException {
...
-        assertFalse(Crypto.verify(serviceToken.getBytes(StandardCharsets.UTF_8), keyPair.getPublic(),
-                testDerSignature, Crypto.SHA256));
-
-        System.clearProperty(Crypto.ATHENZ_CRYPTO_SIGNATURE_PROVIDER);
+        try {
+            assertFalse(Crypto.verify(serviceToken.getBytes(StandardCharsets.UTF_8), keyPair.getPublic(),
+                    testDerSignature, Crypto.SHA256));
+        } finally {
+            System.clearProperty(Crypto.ATHENZ_CRYPTO_SIGNATURE_PROVIDER);
+        }
```
mlajkim:
```suggestion
        assertFalse(Crypto.verify(serviceToken.getBytes(StandardCharsets.UTF_8), keyPair.getPublic(),
                testDerSignature, Crypto.SHA256));

        System.clearProperty(Crypto.ATHENZ_CRYPTO_SIGNATURE_PROVIDER);
```

libs/java/cert_refresher/src/test/java/com/oath/auth/SocketTest.java
```diff
@@ -92,6 +92,9 @@ private void runPingServer(int port, KeyRefresher keyRefresher) throws Exception
 
                             while (running) {
                                 String line = is.readLine();
+                                if (line == null) {
+                                    break;
+                                }
```
mlajkim:
```suggestion
```

mlajkim:
@havetisyan please let me know what your thoughts on this PR.

it is a big change in terms of the direction but all I really want to implement is to have multiple domain token exchange and if you think there are better ways to implement this, I would love to hear (or during the Inc meeting)

servers/zts/src/test/java/com/yahoo/athenz/zts/ZTSImplAccessTokenTest.java
```diff
@@ -4458,6 +4559,90 @@ public void testGenerateScopeResponseNullDomain() {
...
+    @Test
+    public void testGenerateAccessTokenScopeListMultipleDomains() {
+
+        Map<String, Set<String>> rolesByDomain = new HashMap<>();
+        rolesByDomain.put("mcp-hub", new HashSet<>(Arrays.asList("api-mcp-accessor")));
+        rolesByDomain.put("api", new HashSet<>(Arrays.asList("docs-getter")));
+        rolesByDomain.put("weather", new HashSet<>(Arrays.asList("reader")));
+
+        List scopes = zts.generateAccessTokenScopeList(rolesByDomain, "mcp-hub");
+
```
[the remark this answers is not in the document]
mlajkim: https://github.com/AthenZ/athenz/pull/3407/commits/33abec2e8ad372e350e623ed8d942054f3d01145

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2944,7 +3043,7 @@ AccessTokenResponse processAccessTokenDelegationRequest(ResourceContext ctx, Pri
 
         // validate that the requested scopes are a subset of the subject token scopes
 
-        String[] requestedRoles = tokenExchangeRequestedRoles(accessTokenRequest, subjectToken, requestDomainName);
+        String[] requestedRoles = tokenExchangeRequestedRoles(accessTokenRequest, subjectToken, requestDomainName, true);
         if (requestedRoles == null) {
             throw requestError("Invalid scope for token exchange", caller, requestDomainName, principalDomain);
```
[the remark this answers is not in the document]
mlajkim: https://github.com/AthenZ/athenz/pull/3407/commits/33abec2e8ad372e350e623ed8d942054f3d01145

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2768,7 +2767,7 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
```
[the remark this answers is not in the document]
mlajkim: https://github.com/AthenZ/athenz/pull/3407/commits/33abec2e8ad372e350e623ed8d942054f3d01145

mlajkim:
@havetisyan (cc: @ctyano)

Hey, I think this PR is ready to be reviewed! Let me know what you think!

mlajkim:
(rebased)

mlajkim:
(update branch; still reviewable)

mlajkim:
(update branch; still reviewable)

mlajkim:
@havetisyan 

I don't mean to rush you, but we are trying to implement the token exchange cases internally so I would greatly appreciate your feedback on this when possible. Thanks 🙏

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2927,47 +2924,130 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
     }
 
+    boolean isPrincipalAuthorizedForTokenTargetExchange(final String targetDomainName,
+            final String sourceDomainName, final String role, Principal principal) {
+
+        // Prefer the target role domain for new tokens that can carry fully qualified scopes
+        // even when their audience is a different source domain. Keep the legacy source-domain
+        // resource for existing deployments that still authorize exchanges that way.
+
+        final String targetRoleResource = targetDomainName + ":" +
```
havetisyan: didn't quite understand why the targetDomainName is preferred here where we're doing the check in the targetDomainName itself. the idea is to tie the source and target domains in the authorization and your change completely removes that.

servers/zts/src/main/java/com/yahoo/athenz/zts/token/AccessTokenScope.java
```diff
@@ -24,6 +24,8 @@ public class AccessTokenScope extends OAuthTokenScope {
             System.getProperty(ZTSConsts.ZTS_PROP_OAUTH_OPENID_SCOPE, "false"));
     private static boolean supportRolesWithoutDomain = Boolean.parseBoolean(
             System.getProperty(ZTSConsts.ZTS_PROP_SCOPE_ROLE_WOUT_DOMAIN, "false"));
+    private static int maxDomains = Integer.parseInt(
+            System.getProperty(ZTSConsts.ZTS_PROP_ACCESS_TOKEN_MAX_DOMAINS, "20"));
```
havetisyan: the default value must be 1 to maintain existing functionality.

servers/zts/src/main/java/com/yahoo/athenz/zts/token/OAuthTokenScope.java
```diff
@@ -36,6 +36,7 @@ public class OAuthTokenScope {
     public static String OBJECT_ROLES   = "roles";
 
     Set<String> domainNames = new HashSet<>();
+    Set<String> domainScopeNames = new HashSet<>();
```
havetisyan: not sure if I followed what was the reason to add a new hashset for every token scope object? the only place it's used is in getRoleNames call though I'm not sure if it does anything since originally we were not setting the roleNames when sendScopeResponse was true, but that logic was changed

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2927,47 +2924,130 @@ AccessTokenResponse processAccessTokenImpersonationRequest(ResourceContext ctx,
...
             LOGGER.error("requested roles are not a subset of subject token roles");
             return null;
         }
     }
 
+    boolean subjectTokenHasRequestedRoles(Set<String> subjectScopes, final String sourceDomainName,
+            final String requestDomainName, String[] requestedRoles) {
+
+        for (String requestedRole : requestedRoles) {
+            if (subjectScopes.contains(requestDomainName + OAuthTokenScope.OBJECT_ROLE + requestedRole)) {
```
havetisyan: we should use ResourceUtils.roleResourceName for consistency

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3731,103 +3836,136 @@ AccessTokenResponse processAccessTokenStandardRequest(ResourceContext ctx, Princ
...
-                validate(requestedRole, TYPE_ENTITY_NAME, principalDomain, caller);
-            }
+        Set<String> scopeDomainNames = tokenScope.getDomainNames();
+        if (scopeDomainNames.size() > 1 && StringUtil.isEmpty(accessTokenRequest.getAudience())) {
+            throw requestError("Multiple scope domains require an audience", caller,
+                    ZTSConsts.ZTS_UNKNOWN_DOMAIN, principalDomain);
         }
 
-        // first retrieve our domain data object from the cache
+        final String domainName = StringUtil.isEmpty(accessTokenRequest.getAudience()) ?
```
havetisyan, line 3868: I'm somewhat concerned about this change. Unless I'm missing something - this could be pretty much any valid domain name and got nothing to do with the scope domains. That is not acceptable. Right now the applications are assuming that the audience in the access token is their domain name and then scope is the list of roles. Now, I can just create an access token with audience sports and then include roles from my domain and present that token to sports. The service may assume that since the audience is sports the user must have access to some role within the sports domain which is not the case. Am I missing some validation?

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: https://github.com/AthenZ/athenz/pull/3407/changes/f1af7785484d79172e7b6ad68405159b48fbfdaa

image

servers/zts/src/main/java/com/yahoo/athenz/zts/token/AccessTokenScope.java
mlajkim: Right.

applied the change:
https://github.com/AthenZ/athenz/commit/f1af7785484d79172e7b6ad68405159b48fbfdaa

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: Oops, I didn't mean to add such a breaking change.

Just removed it and roll backed to where it has been now:

https://github.com/AthenZ/athenz/pull/3407/commits/e7b256aee19edbb7cccb2cd40cb1458603e6a26a

servers/zts/src/main/java/com/yahoo/athenz/zts/token/OAuthTokenScope.java
mlajkim: Good point. The reason was to handle mixed multi-domain scopes such as:
`sports:domain weather:role.reader`

 With the old global `sendScopeResponse` behavior, once `sports:domain` is present we would not keep `roleNames` at all. That works for the old single-domain case, but for multi-domain scopes it would also drop the `weather:role.reader` filter and make `weather` unfiltered as well.

That said, I agree we don't need to keep this as object state. I removed the field and now use it only as local parsing state to remove role/group filters for domains that requested `{domain}:domain`.

https://github.com/AthenZ/athenz/pull/3407/commits/f11df9799b1b096c73c91251bebd8cfaeac32d31

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: You're right. I removed that carrier-token behavior as the following:
https://github.com/AthenZ/athenz/pull/3407/commits/6aba3c0150f423d2e154c361cd659ccad2635ed0

For standard access token requests, when `audience` is specified it must now be one of the requested scope domains. Multi-domain scope requests still require an explicit audience, but that audience can no longer be unrelated to the scope domains. I also added a negative test for `audience=sports&scope=weather:role.readers`

mlajkim:
@havetisyan

Hey, thank you for the review! I took some time to thoroughly go through the feedback and make the updates. Please let me know what you think, and no rush.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2964,43 +2963,104 @@ String[] tokenExchangeRequestedRoles(AccessTokenRequest accessTokenRequest, OAut
...
     }
 
+    boolean subjectTokenHasRequestedRoles(Set<String> subjectScopes, final String sourceDomainName,
+            final String requestDomainName, String[] requestedRoles) {
+
+        for (String requestedRole : requestedRoles) {
+            if (subjectScopes.contains(ResourceUtils.roleResourceName(requestDomainName, requestedRole))) {
+                continue;
+            }
+            if (requestDomainName.equals(sourceDomainName) && subjectScopes.contains(requestedRole)) {
```
havetisyan: lets move the requestDomainName.equals(sourceDomainName) outside of the for loop since those values do not change within the loop. and since we always have one domain where they match, it's most likely not going to be a wasted operation.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -2964,43 +2963,104 @@ String[] tokenExchangeRequestedRoles(AccessTokenRequest accessTokenRequest, OAut
...
+        for (String subjectScope : subjectScopes) {
+            if (AccessTokenScope.OBJECT_OPENID.equals(subjectScope)) {
+                continue;
+            }
+            if (subjectScope.startsWith(roleScopePrefix)) {
+                String roleName = subjectScope.substring(roleScopePrefix.length());
+                if (!roleName.isEmpty()) {
+                    requestedRoles.add(roleName);
+                }
+            } else if (!subjectScope.contains(":") && requestDomainName.equals(sourceDomainName)) {
```
havetisyan: same comment as above - let's move the requestDomainName.equals(sourceDomainName) outside of the loop again

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3088,7 +3148,7 @@ AccessTokenResponse processAccessTokenDelegationRequest(ResourceContext ctx, Pri
 
         AccessToken accessToken = new AccessToken();
         accessToken.setVersion(1);
-        accessToken.setJwtId(UUID.randomUUID().toString());
+        accessToken.setJwtId(java.util.UUID.randomUUID().toString());
```
havetisyan: out of curiosity, why was this change done in multiple places?

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3154,6 +3214,33 @@ String generateScopeResponse(Set<String> roles, final String domainName, boolean
         return sb.toString();
     }
 
+    List<String> generateAccessTokenScopeList(Map<String, Set<String>> rolesByDomain, final String audienceDomainName) {
+
+        List<String> scopes = new ArrayList<>();
+        for (String domainName : new TreeSet<>(rolesByDomain.keySet())) {
+            for (String role : new TreeSet<>(rolesByDomain.get(domainName))) {
+                if (domainName.equals(audienceDomainName)) {
```
havetisyan: same comment here - let's move the   domainName.equals(audienceDomainName) outside of the loop - here we're definitely don't need to keep checking for every role

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3920,8 +4044,14 @@ AccessTokenResponse processAccessTokenStandardRequest(ResourceContext ctx, Princ
         // does not match the returned list of roles then we need to return the updated
         // set of scopes
 
-        if (tokenScope.sendScopeResponse() || requestedRoles != null && requestedRoles.length != roles.size()) {
-            response.setScope(generateScopeResponse(roles, domainName, tokenScope.isOpenIdScope()));
+        if (scopeAdjusted || scopeDomainNames.size() > 1 || !scopeDomainNames.contains(domainName)) {
```
havetisyan: the last check !scopeDomainNames.contains(domainName) should be removed since it's not valid - we already verified at the beginning that's not allowed

servers/zts/src/test/java/com/yahoo/athenz/zts/token/AccessTokenScopeTest.java
```diff
@@ -200,25 +215,55 @@ public void testAccessTokenScopeMultipleDomains() {
         AccessTokenScope req1 = new AccessTokenScope("sports:domain sports:domain", null);
         assertNotNull(req1);
 
-        try {
```
havetisyan, line 203: any reason why these tests were modified or changed? I expect all the existing test cases to stay as is without any changes. I expect we have only added tests for the new functionality when the max domains is set to value > 1 and do not touch any of the existing ones.

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: Simply accepted the AI-assisted cleanup but I no longer see any reasons behind it, so I just reverted it here:

https://github.com/AthenZ/athenz/pull/3407/changes/97f5ac5efea694c7161fc1b6e12dddedd7527fbd

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: handled in https://github.com/AthenZ/athenz/pull/3407/changes/46fd150990f02a3f626c7d7925b42e5ddb240791

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: handled in https://github.com/AthenZ/athenz/pull/3407/changes/46fd150990f02a3f626c7d7925b42e5ddb240791

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: handled in https://github.com/AthenZ/athenz/pull/3407/changes/46fd150990f02a3f626c7d7925b42e5ddb240791

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: Correct; handled in here https://github.com/AthenZ/athenz/pull/3407/changes/97333bb9b174c1e0b12d2f0505d46e1d606c083d

servers/zts/src/test/java/com/yahoo/athenz/zts/token/AccessTokenScopeTest.java
mlajkim: This is likely because I initially made a destructive change to the multiple domains setting with a default of `20`, so I modified the existing test cases.

I have rolled back the original test code and simply implemented `domain >= 2` here: https://github.com/AthenZ/athenz/pull/3407/changes/048bc0839948ccc4f236d1b762ea9ae65c1a9c30

havetisyan (approved):

PR #3424, opened 2026-07-21
workaround support for CNs longer than 64 chars

havetisyan:
# Description
 When creating a new X500Name object, BC starting with 1.85 version enforces strict 64 char limit for cn as required by rfc5280. to support existing clients that might already have identities longer than 64 chars, changes were implemented to support those use cases while introducing other controls to limit the identity less than 64 chars.

# Contribution Checklist:


yosrixp (approved):
Looks good

PR #3427, opened 2026-07-21
Support role name in access token audience claim (role_in_aud_claim)

jeffreytolar:
# Description

Adds an option to place the role name in the `aud` claim of ZTS **access
tokens**, mirroring the existing `-role-in-aud-claim` support for OIDC id
tokens (added in #2190).

Today a standard (`client_credentials`) access token always uses the domain as
its audience (`aud=<domain>`), with roles carried only in the `scp`/`scope`
claim. A resource server that authorizes purely on the `aud` claim can
therefore only gate at "holds any role in the domain". This change lets such a
server enforce a tighter per-role gate using an access token — the same gating
`-role-in-aud-claim` already enables for id tokens.

The ZTS access-token endpoint already parsed the `role_in_aud_claim` request
parameter, but only the JAG token-exchange path honored it. This wires it into
the standard flow and exposes it through the `zts-accesstoken` CLI and SIA.

## Server (ZTS)
- `processAccessTokenStandardRequest` now sets the audience via a new
  `getAccessTokenAudience(...)` helper that mirrors `getIdTokenAudience`: when
  `role_in_aud_claim` is set and exactly one role is returned, the audience
  becomes `<domain>:<role>` instead of `<domain>`. Otherwise the audience is
  unchanged.
- No RDL / schema / generated-client change is needed — `role_in_aud_claim` is
  a form-body parameter already parsed by `AccessTokenRequest`.

## zts-accesstoken CLI
- New `-role-in-aud-claim` flag that adds `role_in_aud_claim=true` to the token
  request.

## SIA
- New `role_in_aud_claim` per-role option under `access_tokens`, threaded
  through the access-token request the SIA agent builds.

# Contribution Checklist:

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -2273,6 +2273,11 @@ public class ZTSImpl implements ZTSHandler {
                 clientId + ":" + idTokenGroups.get(0) : clientId;
     }
 
+    String getAccessTokenAudience(final String domainName, boolean roleInAudClaim, List<String> roles) {
+        return (roleInAudClaim && roles != null && roles.size() == 1) ?
+                domainName + ":" + roles.get(0) : domainName;
+    }
+
     List<String> processIdTokenGroups(final String principalName, IdTokenScope tokenRequest,
                                       final String clientIdDomainName, Boolean allScopePresent,
                                       final String principalDomain, final String caller) {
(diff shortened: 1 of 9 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3755,7 +3760,8 @@ AccessTokenResponse processAccessTokenStandardRequest(ResourceContext ctx, Princ
         AccessToken accessToken = new AccessToken();
         accessToken.setVersion(1);
         accessToken.setJwtId(UUID.randomUUID().toString());
-        accessToken.setAudience(domainName);
+        accessToken.setAudience(getAccessTokenAudience(domainName, accessTokenRequest.isRoleInAudClaim(),
```
havetisyan: We seem to be wasting a new array list object for every access token call. we should just pass the roles set directly and use set.iterator().next() when we need to fetch the item

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
jeffreytolar: Pushed 780b737b7752b46979596b12cf080edc527b8d2b - it turns out there was already a list version being created to set the scopes, so I instead extracted that out into a local to share between this check and the `setScopes` call

havetisyan (approved):

Issue #3365, opened 2026-05-18
Enhancement proposal: Transition from AWS temporary credentials to an `sts:GetWebIdentityToken` issued JWT for workload attestation in AWS


abvaidya:
## Description

In the current Athenz architecture, the Service Identity Agent (SIA) running inside AWS environments (such as EC2, ECS, Fargate, and Lambda) utilizes AWS temporary credentials (Access Key ID, Secret Access Key, and Session Token) as attestation data. This data is bundled into the `InstanceRegisterInformation` or `InstanceRefreshInformation` object and sent to the ZTS server (or a Copper Argos callback provider) to verify the workload's cloud identity.

While functional, this model has operational concerns:

1. **Synchronous External Calls & Latency:** To validate the temporary credentials, the ZTS server or the Copper Argos provider must make an active outbound API call to AWS STS (`sts:GetCallerIdentity`) for *every* instance registration or refresh request. This adds network latency and introduces a strict dependency on AWS uptime.
2. **Throttling and Rate Limits:** In large-scale multi-tenant deployments with thousands of dynamic microservices rotating or refreshing their identities concurrently, relying on synchronous `sts:GetCallerIdentity` calls risks hitting AWS STS API rate limits and throttling thresholds.

**Proposed Enhancement**

In Nov-2025 AWS introduced a new feature - [Outbound Identity Federation](https://aws.amazon.com/blogs/aws/simplify-access-to-external-services-using-aws-iam-outbound-identity-federation/), I propose to update SIA to use a signed JSON Web Token (JWT) issued via the AWS STS `sts:GetWebIdentityToken` API operation as the attestation data as an alternative to existing temporary credentials flow.

### How it Works:

1. **SIA Token Retrieval:** The SIA container/agent invokes the local AWS STS provider to call `GetWebIdentityToken`, specifying the target  ZTS endpoint as audience (`aud` claim). AWS STS returns a short-lived, cryptographically signed JWT representing the caller's IAM identity.
2. **Attestation Data Submission:** SIA generates a standard CSR and places this `WebIdentityToken` JWT into the `attestationData` field of the registration payload instead of AWS temporary credentials.
3. **Stateless Verification at ZTS:** When ZTS (or the Copper Argos provider) receives the request, it validates the JWT signature completely offline/locally using the public keys published at the AWS STS OpenID Connect (OIDC) JWKS discovery endpoint found in IAM -> Account Settings -> Outbound Identity Federation -> Token Issuer URL.

### Benefits:

[...]

PR #3432, opened 2026-07-26
feat:support for aws outbound identity federation

abvaidya:
# Description

# Contribution Checklist:

### DIFF
--- a/libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/AWSWebIdentityTokenAttestationValidator.java
+++ b/libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/AWSWebIdentityTokenAttestationValidator.java
@@ -56,22 +56,29 @@ import static com.yahoo.athenz.instance.provider.InstanceProvider.ZTS_INSTANCE_A
  *   <li>the audience matches the configured audience,</li>
  *   <li>the token's aws_account matches the domain's AWS account, or a launch
  *       authorization is granted for the token's account,</li>
- *   <li>the org_id is present in the configured allowlist, and</li>
- *   <li>the sub and principal_tags pass an optional adopter-supplied
- *       {@link AttrValidator}.</li>
+ *   <li>the org_id is present in the configured allowlist,</li>
+ *   <li>the iam role name in the token subject matches the requested athenz
+ *       service (the default binding that prevents one role from obtaining
+ *       another service's identity), and</li>
+ *   <li>the sub and principal_tags pass an adopter-supplied {@link AttrValidator}
+ *       if one is configured. When configured, the AttrValidator has the final
+ *       say on the principal and is the escape hatch for adopters whose iam role
+ *       names diverge from the service - it is consulted even when the default
+ *       role binding above does not match. Without one, the default binding is
+ *       strictly enforced.</li>
  * </ul>
  */
 public class AWSWebIdentityTokenAttestationValidator implements AWSAttestationValidator {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AWSWebIdentityTokenAttestationValidator.class);
 
-    static final String AWS_PROP_ATTESTATION_AUDIENCE = "athenz.zts.aws.attestation_audience";
-    static final String AWS_PROP_OIDC_ISSUER_REGEX    = "athenz.zts.aws.oidc_issuer_regex";
-    static final String AWS_PROP_OIDC_ALLOWED_ORG_IDS = "athenz.zts.aws.oidc_allowed_org_ids";
-    static final String AWS_PROP_OIDC_STS_CLAIM_NAME  = "athenz.zts.aws.oidc_sts_claim_name";
-    static final String AWS_PROP_PRINCIPAL_VALIDATOR_FACTORY_CLASS = "athenz.zts.aws.oidc_principal_validator_factory_class";
+    static final String AWS_PROP_WEB_IDENTITY_AUDIENCE = "athenz.zts.aws_web_identity_audience";
+    static final String AWS_PROP_WEB_IDENTITY_ISSUER_REGEX    = "athenz.zts.aws_web_identity_issuer_regex";
+    static final String AWS_PROP_WEB_IDENTITY_ALLOWED_ORG_IDS = "athenz.zts.aws_web_identity_allowed_org_ids";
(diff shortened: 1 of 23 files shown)


abvaidya:
implements #3365 for server side, client side will follow in a separate PR

havetisyan:
@abvaidya why is the sub claim validation optional without a default implementation? So what's stopping any of roles in my aws account to obtain any other service identity? e.g. if I have a sys.auth.cron role then it can request an identity cert for sys.auth.zts and by default zts will issue one?

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/AWSWebIdentityTokenAttestationValidator.java
```diff
@@ -0,0 +1,310 @@
...
+
+    final JwtsHelper jwtsHelper = new JwtsHelper();
+    final Map<String, JwtsSigningKeyResolver> issuersMap = new ConcurrentHashMap<>();
+    String oidcDiscoveryProxy;
+
+    @Override
+    public void initialize(SSLContext sslContext, Authorizer authorizer) {
+
+        this.authorizer = authorizer;
+        audience = System.getProperty(AWS_PROP_ATTESTATION_AUDIENCE, "athenz.io");
```
havetisyan: we should probably not have a default audience value here and let the admin define one so they don't screw up and let their server run with the default value all the time. and if you don't have one configured, we should error out.

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceAWSProvider.java
```diff
@@ -317,12 +343,13 @@ public InstanceConfirmation confirmInstance(InstanceConfirmation confirmation) t
...
-        // instance identity
-        
-        if (!verifyInstanceIdentity(info, awsAccount)) {
-            throw error("Unable to verify instance identity credentials");
+        // verify the instance identity via the configured attestation validator.
+        // depending on the attestation data this either validates the STS
+        // temporary credentials or the AWS web identity token
+
+        StringBuilder identityErrMsg = new StringBuilder(256);
+        if (!attestationValidator.validateIdentity(confirmation, info, awsAccount, identityErrMsg)) {
```
havetisyan, line 351: we have to be careful here since we'll be exposing some internal details about the error - before we were just returning generic unable to verify instance creds error. since this api is open to everyone and not mtls protected, we should be careful about not to expose error details from aws

havetisyan:
can we also update conf/aws.properties with new property descriptions.

also consider renaming the attributes to maintain consistency. all the existing ones are named athenz.zts.aws_ while all the new ones are athenz.zts.aws.

abvaidya:
> @abvaidya why is the sub claim validation optional without a default implementation? So what's stopping any of roles in my aws account to obtain any other service identity? e.g. if I have a sys.auth.cron role then it can request an identity cert for sys.auth.zts and by default zts will issue one?

sub claim validation can be internal / specific to adopters, so I thought it should to be happen in the optional PrincipalValidator. What would you recommend for default implementation for sub which can be universal?

havetisyan:
> > @abvaidya why is the sub claim validation optional without a default implementation? So what's stopping any of roles in my aws account to obtain any other service identity? e.g. if I have a sys.auth.cron role then it can request an identity cert for sys.auth.zts and by default zts will issue one?
> 
> sub claim validation can be internal / specific to adopters, so I thought it should to be happen in the optional PrincipalValidator. What would you recommend for default implementation for sub which can be universal?

I think the default implementation should require that the role is {domain}.{service}. If you want it something else, or you want to remove it completely, then it's up to the admin. But by default we validate that {domain}.{service} role gets an identity for {domain}.{service}

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/AWSWebIdentityTokenAttestationValidator.java
```diff
@@ -0,0 +1,379 @@
...
+
+        this.authorizer = authorizer;
+        audience = System.getProperty(AWS_PROP_WEB_IDENTITY_AUDIENCE, null);
+        issuerPattern = Pattern.compile(System.getProperty(AWS_PROP_WEB_IDENTITY_ISSUER_REGEX, AWS_DEFAULT_WEB_IDENTITY_ISSUER_REGEX));
+        stsClaimName = System.getProperty(AWS_PROP_WEB_IDENTITY_STS_CLAIM_NAME, AWS_DEFAULT_STS_CLAIM_NAME);
+        allowedOrgIds = new DynamicConfigCsv(CONFIG_MANAGER, AWS_PROP_WEB_IDENTITY_ALLOWED_ORG_IDS, null);
+        oidcDiscoveryProxy = System.getProperty(AWS_PROP_WEB_IDENTITY_DISCOVERY_PROXY, null);
+        principalValidator = newPrincipalValidator(sslContext);
+        if (audience == null) {
+            LOGGER.error("audience must be set for validation, no instance requests will be authorized.");
```
havetisyan, line 113: any reason why here and and on line 117 we're not throwing Illegal Argument exception since all the requests will be rejected anyway?

abvaidya: I was following our other such validations. This would cause an issue for someone wanting to continue with current validator right? since the default configured is a combined validator

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/AWSWebIdentityTokenAttestationValidator.java
```diff
@@ -0,0 +1,379 @@
...
+        try {
+            idToken = new IdToken(jwToken, signingKeyResolver);
+        } catch (Exception ex) {
+            errMsg.append("unable to parse and validate token: ").append(ex.getMessage());
+            return false;
+        }
+
+        // verify that the token audience is set for our service
+
+        if (!audience.equals(idToken.getAudience())) {
```
havetisyan, line 212: if we do not thrown an exception in the constructor then we'll get NPE here if audience is not configured

abvaidya: that is true, I will add a null check, throwing an exception may be a problem based on my comment above

havetisyan (approved):

PR #3435, opened 2026-07-29
feat: support for istio ambient mode

### LINKED ISSUE #3436: Support Istio ambient mesh SPIFFE SVID issuance in ZTS
## Summary

ZTS currently assumes every X.509 service certificate request has a Subject CN, DNS SANs, and a dot-separated `{domain}.{service}` SPIFFE service account name. None of that holds for Istio ambient mesh workload/control-plane identities issued through `cert-manager-istio-csr` → `cert-manager` → an external `athenz-issuer`:

- Workload and waypoint SVIDs have an **empty Subject** (SPIFFE spec) and **no DNS SANs** — identity lives entirely in the SPIFFE URI SAN.
- Control-plane certs (istiod, ztunnel, istio-csr) carry hostname DNS SANs but no `{instanceid}.instanceid.athenz...` SAN, since istio-csr generates its own CSR without one.
- The waypoint's Kubernetes ServiceAccount is dot-free (Gateway API requires the Service and SA to share a DNS-1035-compliant name), so the Athenz domain can't be encoded via `{domain}.{service}` and must instead come from an `athenz.io/domain` annotation.

Full design, sequence diagrams, and cluster verification: [`docs/design/istio-ambient-cert-flow.md`](docs/design/istio-ambient-cert-flow.md).

## Proposed changes

- **Skip CN validation for empty Subject** (`X509ServiceCertRequest.java`) — SPIFFE SVIDs have no Subject DN by spec; rely on URI SAN validation instead.
- **Bypass DNS hostname validation for SPIFFE-only SVIDs** (`ZTSImpl.java`) — when `sanDns` is empty and `sanUri` starts with `spiffe://`, skip `confirmInstance()`'s DNS check entirely.
- **Accept empty Subject `O` field** (`X509CertRequest.java`) — the waypoint's Envoy CSR presents `Subject: O=` (empty, not absent); treat an empty value the same as an absent one.
- **Accept Form 2 SPIFFE URI for dot-free SAs** (`X509ServiceCertRequest.java`) — `spiffe://{trustDomain}/ns/{ns}/sa/{service}` (no domain prefix), for SAs whose Athenz domain is carried via annotation rather than the SA name.
- **Fall back to SA-token instanceId when absent from CSR SANs** (`InstanceUtils.java`) — derive `instanceId` from the K8s SA UID in the validated identity token when no `{uid}.instanceid.athenz...` entry is present in the CSR.

## Verified

End-to-end on a live GKE cluster (Istio ambient + waypoint), including istiod/ztunnel/istio-csr control-plane certs, workload SVIDs (curl → httpbin), and a waypoint enforcing `AuthorizationPolicy` (GET allowed, POST denied). See the design doc for full sequence diagrams and cluster details.

psasidhar:
## Summary

Enables ZTS to issue SPIFFE SVIDs for an Istio ambient mesh — istiod, ztunnel and istio-csr control-plane certificates, plus per-workload and waypoint SVIDs — via cert-manager and [athenz-issuer](https://github.com/AthenZ/athenz-issuer).

All mesh identities use **dotted ServiceAccounts** (`{domain}.{service}`), so every SVID is a standard domain-qualified SPIFFE URI (`spiffe://<trust-domain>/ns/<namespace>/sa/<domain>.<service>`) validated by the existing `SpiffeUriTrustDomain` validator. **No new SPIFFE validator, no annotation-driven domain lookup, and no new Athenz policy/assertion shape are introduced** — the domain stays inside the identity and remains cryptographically bound to the ServiceAccount name attested in the K8s token.

The changes are the minimum needed to accept CSRs generated by Istio components, which ZTS previously rejected.

## Changes

**`X509ServiceCertRequest`** — SPIFFE SVIDs carry no Subject by spec, so the `CN == domain.service` check is skipped when the CSR Subject is empty. A CSR with neither a CommonName nor a SPIFFE URI is now explicitly rejected (previously it failed with a misleading CN error).

**`X509CertRequest`** — treat an empty Subject `O` field the same as an absent one. Istio's `GenCSRTemplate` unconditionally writes `Organization: []string{options.Org}`, producing `O=` with an empty value when unset, which failed validation against the allowed-values set.

**`InstanceK8SProvider` / `DefaultGCPGoogleKubernetesEngineValidator`** — DNS SAN validation is skipped only when the CSR carries no DNS SANs *and* has a SPIFFE URI SAN. Attestation (`confirmInstance`) always runs, so the SA token is fully verified regardless.

The same provider now also **binds the SPIFFE namespace to the attestation**. K8s attestation compares only the ServiceAccount *name*; the namespace in a request's SPIFFE URI was matched against the `namespace` request field, which athenz-issuer derives from that same URI — making the comparison tautological. The provider now requires the namespace in the CSR's SPIFFE URI to equal the namespace in the validated id_token subject. This applies to every request carrying a SPIFFE URI, including DNS-bearing csi-driver requests; requests without a SPIFFE URI carry no namespace in their identity and are unaffected.

**`InstanceRegisterInformation.x509CertInstanceId`** — new optional field mirroring the existing `jwtSVIDInstanceId`. Control-plane certs and dynamic workload/waypoint SVIDs carry no `{uid}.instanceid.athenz` SAN, because their CSRs are generated by components whose private keys athenz-issuer never holds. The caller now supplies the instance id explicitly; **ZTS never inspects attestation data to derive one**, and an instance id present in the CSR always takes precedence.

## Behavior change to be aware of

A deployment whose SPIFFE URI namespace intentionally differs from the pod's real namespace will now be rejected by the K8s provider. We believe that combination is not meaningful — the namespace is part of the issued identity — but calling it out explicitly for review.

## Docs

- `docs/design/istio-ambient-cert-flow.md` — end-to-end issuance flows for each certificate, sequence diagrams, per-hop token/identity tables, and the rationale for each patch above.
- `docs/design/istio-ambient-waypoint-setup.md` — the platform-provisioned waypoint deployment that yields the dotted ServiceAccount. Istio's *managed* waypoint path cannot produce one: the gateway controller derives the K8s Service name from the Gateway name and Service names must be DNS-1035 (no dots), so a dotted Gateway name fails Service creation permanently. Documents the resource set, the verification steps, and the caveats.

## Known follow-up

`x509CertInstanceId` is not yet validated against the ServiceAccount UID claim in the token. It can be for this flow, but not for csi-driver-athenz, which derives its instance id from the **Pod** UID against an SA-bound-only token with no pod claim — so a safe check must be scoped to ids that arrived via `x509CertInstanceId`. Rationale and residual risk (attribution only, no identity escalation, and K8s refresh is rejected outright) are documented in the cert-flow doc under Patch 5.

## Testing

- Unit tests for every change; `mvn verify` passes with JaCoCo line coverage at 100% for all touched modules.
- Running end to end in an Istio ambient test cluster on GKE: control-plane certs, workload SVIDs and a waypoint SVID all issued through this path, with mTLS traffic and L7 `AuthorizationPolicy` enforcement verified between workloads.

### DIFF
--- a/libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceUtils.java
+++ b/libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceUtils.java
@@ -45,6 +45,9 @@ public class InstanceUtils {
 
     static final String K8S_SERVICE_ACCOUNT_PREFIX = "system:serviceaccount:";
 
+    static final String SPIFFE_URI_PREFIX    = "spiffe://";
+    static final String SPIFFE_URI_NAMESPACE = "/ns/";
+
     public static String getInstanceProperty(final Map<String, String> attributes,
             final String propertyName) {
 
@@ -207,11 +210,23 @@ public class InstanceUtils {
         }
         String[] hosts = hostnames.split(",");
 
-        // extract the instance id from the request
-
+        // extract the instance id from the request. if absent, fall back
+        // to ZTS_INSTANCE_ID (e.g. set by ZTS from the K8s SA token UID)
         if (!extractCertRequestInstanceId(attributes, hosts, dnsSuffixes, instanceId)) {
-            LOGGER.error("Request does not contain expected instance id entry");
-            return false;
+
+            // extractCertRequestInstanceId may have already appended a partial
+            // value (e.g. conflicting instanceid SAN entries) - do not fall back
+            if (instanceId.length() != 0) {
+                return false;
+            }
+
+            final String ztsInstanceId = getInstanceProperty(attributes, InstanceProvider.ZTS_INSTANCE_ID);
+            if (!StringUtil.isEmpty(ztsInstanceId)) {
+                instanceId.append(ztsInstanceId);
+            } else {
+                LOGGER.error("Request does not contain expected instance id entry");
+                return false;
+            }
         }
 
         // for hostnames that are included in the sanDNS entry in the certificate we have
@@ -390,6 +405,48 @@ public class InstanceUtils {
             return null;
         }
         return components[3];
+    }
+
+    public static String getNamespaceFromIdTokenSubject(final String sub) {
+        if (StringUtil.isEmpty(sub) || !sub.startsWith(K8S_SERVICE_ACCOUNT_PREFIX)) {
+            return null;
+        }
+        String[] components = sub.split(":");
+        if (components.length != 4) {
+            return null;
+        }
+        return components[2];
+    }
+
+    /**
+     * Extract the namespace component from the spiffe uri in the given list of
+     * certificate request uris. The expected format is
+     * spiffe://&lt;trust-domain&gt;/ns/&lt;namespace&gt;/sa/&lt;service-account&gt;.
+     * @param sanUri comma separated list of uris from the certificate request
+     * @return namespace from the spiffe uri, null if the request carries no
+     *      spiffe uri with a namespace component
+     */
+    public static String getNamespaceFromSpiffeUri(final String sanUri) {
+
+        if (StringUtil.isEmpty(sanUri)) {
+            return null;
+        }
+        for (String uri : sanUri.split(",")) {
+            if (!uri.startsWith(SPIFFE_URI_PREFIX)) {
+                continue;
+            }
+            int idx = uri.indexOf(SPIFFE_URI_NAMESPACE);
+            if (idx == -1) {
+                continue;
+            }
+            int startIdx = idx + SPIFFE_URI_NAMESPACE.length();
+            int endIdx = uri.indexOf('/', startIdx);
+            if (endIdx == -1) {
+                continue;
+            }
+            return uri.substring(startIdx, endIdx);
+        }
+        return null;
     }
 
     public static List<IPBlock> parseIPBlocks(final String certAllowedIPAddresses) {
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -5043,7 +5043,8 @@ public class ZTSImpl implements ZTSHandler {
 
         X509ServiceCertRequest certReq;
         try {
-            certReq = new X509ServiceCertRequest(info.getCsr(), spiffeUriManager, certificateDataValidator);
+            certReq = new X509ServiceCertRequest(info.getCsr(), spiffeUriManager, certificateDataValidator,
+                    info.getX509CertInstanceId());
         } catch (CryptoException ex) {
             throw requestError("unable to parse PKCS10 CSR: " + ex.getMessage(),
                     caller, domain, principalDomain);
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/X509CertRequest.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/cert/X509CertRequest.java
@@ -35,6 +35,7 @@ import org.slf4j.LoggerFactory;
 
 import com.yahoo.athenz.auth.util.Crypto;
 import com.yahoo.athenz.auth.util.CryptoException;
+import com.yahoo.athenz.auth.util.StringUtils;
 
 public class X509CertRequest {
 
@@ -57,6 +58,11 @@ public class X509CertRequest {
 
     public X509CertRequest(String csr, SpiffeUriManager spiffeUriManager,
             CertificateDataValidator certificateDataValidator) throws CryptoException {
+        this(csr, spiffeUriManager, certificateDataValidator, null);
+    }
+
+    public X509CertRequest(String csr, SpiffeUriManager spiffeUriManager,
+            CertificateDataValidator certificateDataValidator, final String x509CertInstanceId) throws CryptoException {
 
         certReq = Crypto.getPKCS10CertRequest(csr);
         if (certReq == null) {
@@ -95,24 +101,29 @@ public class X509CertRequest {
 
         uriHostname = X509CertUtils.extractItemFromURI(uris, ZTSConsts.ZTS_CERT_HOSTNAME_URI);
 
-        // extract instanceId
+        // extract instanceId: the CSR itself (athenz URI field, then the SAN
+        // dnsName list) always takes precedence; the caller-supplied value is
+        // only a last resort when the CSR carries none
 
-        // first check to see if we have the instance id is provided
-        // in the athenz uri field
-
-        instanceId = X509CertUtils.extractRequestInstanceIdFromURI(uris);
-
-        // if we have no instance id from the URI, then we're going
-        // to fetch it from the dns list
-
-        if (instanceId == null) {
-            instanceId = X509CertUtils.extractRequestInstanceIdFromDnsNames(dnsNames);
-        }
+        instanceId = extractInstanceId(x509CertInstanceId);
 
         // save the spiffe uri manager and certificate data validator objects
 
         this.spiffeUriManager = spiffeUriManager;
         this.certificateDataValidator = certificateDataValidator;
+    }
+
+    private String extractInstanceId(final String x509CertInstanceId) {
+
+        String csrInstanceId = X509CertUtils.extractRequestInstanceIdFromURI(uris);
+        if (csrInstanceId == null) {
+            csrInstanceId = X509CertUtils.extractRequestInstanceIdFromDnsNames(dnsNames);
+        }
+        if (csrInstanceId != null) {
+            return csrInstanceId;
+        }
+
+        return StringUtils.isEmpty(x509CertInstanceId) ? null : x509CertInstanceId;
     }
 
     boolean knownAthenzUri(final String uri) {
@@ -484,7 +495,7 @@ public class X509CertRequest {
 
         try {
             final String value = Crypto.extractX509CSRSubjectOField(certReq);
-            if (value == null) {
+            if (StringUtils.isEmpty(value)) {
                 return true;
             }
             boolean res = validValues.contains(value);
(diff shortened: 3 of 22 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -5049,6 +5001,18 @@ Response postInstanceX509CertificateRegister(ResourceContext ctx, InstanceRegist
                     caller, domain, principalDomain);
         }
 
+        // When the CSR carries no instanceId (cert-manager / istio-csr flows) but the
+        // request includes a K8s namespace, derive a provisional instanceId from the SA UID
+        // in the attestation JWT payload. The payload is not verified here — confirmInstance
+        // (below) must validate the same token before the certificate is issued. A forged
+        // UID cannot succeed without a valid signature.
+        if (StringUtil.isEmpty(certReq.getInstanceId()) && !StringUtil.isEmpty(info.getNamespace())) {
+            final String saUid = extractK8sSaUidFromAttestation(info.getAttestationData());
```
havetisyan: this is not right. ZTS server itself NEVER looks at the attestation data and it doesn't know what the format is. the expectation here is the caller must specify the request id in the request and the provider is responsible for validating that the data is correct acccording to the attestation data.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -5039,6 +4974,23 @@ Response postInstanceX509CertificateRegister(ResourceContext ctx, InstanceRegist
                     caller, domain, principalDomain);
         }
 
+        LOGGER.info("postInstanceRegisterInformation: domain={} service={} provider={} namespace={} " +
```
havetisyan: any reason why we're adding this line as INFO. This should already be in the access log.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -5039,6 +4974,23 @@ Response postInstanceX509CertificateRegister(ResourceContext ctx, InstanceRegist
...
+            // token/athenzJWK are request-scoped Boolean flags (does the caller want a token
+            // / the Athenz JWK in the response) -- not credential values -- but they carry no
+            // debugging value here either, so they're left out rather than risk confusion.
+            LOGGER.debug("postInstanceRegisterInformation details: hostCnames={} expiryTime={} " +
+                            "x509SignerKeyId={} sshSignerKeyId={} " +
+                            "jwtSVIDInstanceId={} jwtSVIDAudience={} jwtSVIDNonce={} attestationData={}",
+                    info.getHostCnames(), info.getExpiryTime(),
+                    info.getX509CertSignerKeyId(), info.getSshCertSignerKeyId(),
+                    info.getJwtSVIDInstanceId(), info.getJwtSVIDAudience(), info.getJwtSVIDNonce(),
+                    redactJwtSignatures(info.getAttestationData()));
```
havetisyan: again same comment as down below - ZTS does not know what's the attestation data and should not log it anywhere

havetisyan (changes requested):

libs/java/instance_provider/src/main/java/com/yahoo/athenz/instance/provider/impl/InstanceUtils.java
```diff
@@ -207,11 +207,17 @@ public static boolean validateCertRequestSanDnsNames(final Map<String, String> a
...
+        // Prefer instance id from CSR SAN DNS or SAN URI (CSI and legacy flows). Only when
+        // absent, fall back to ZTS_INSTANCE_ID — set server-side by ZTS from the validated
+        // K8s SA token UID (cert-manager control-plane certs without {uid}.instanceid.athenz).
         if (!extractCertRequestInstanceId(attributes, hosts, dnsSuffixes, instanceId)) {
-            LOGGER.error("Request does not contain expected instance id entry");
-            return false;
+            final String ztsInstanceId = getInstanceProperty(attributes, InstanceProvider.ZTS_INSTANCE_ID);
+            if (!StringUtil.isEmpty(ztsInstanceId)) {
+                instanceId.append(ztsInstanceId);
+            } else {
```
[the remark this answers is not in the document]
psasidhar: fixed the issue raised

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -4973,6 +4974,21 @@ Response postInstanceX509CertificateRegister(ResourceContext ctx, InstanceRegist
...
+                info.getHostname(), info.getCloud());
+        if (LOGGER.isDebugEnabled()) {
+            LOGGER.debug("postInstanceRegisterInformation details: hostCnames={} expiryTime={} token={} " +
+                            "x509SignerKeyId={} sshSignerKeyId={} athenzJWK={} athenzJWKModified={} " +
+                            "jwtSVIDInstanceId={} jwtSVIDAudience={} jwtSVIDNonce={} attestationData={}",
+                    info.getHostCnames(), info.getExpiryTime(),
+                    info.getToken(), info.getX509CertSignerKeyId(), info.getSshCertSignerKeyId(),
+                    info.getAthenzJWK(), info.getAthenzJWKModified(),
+                    info.getJwtSVIDInstanceId(), info.getJwtSVIDAudience(), info.getJwtSVIDNonce(),
+                    redactJwtSignatures(info.getAttestationData()));
```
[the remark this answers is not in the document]
psasidhar: removed this logging here, and logged issuer/subject in InstanceK8sProvider at debug level.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
psasidhar: added x509CertInstanceId field and started relying on that.

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
psasidhar: removed this logging

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
psasidhar: removed it

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/X509CertRequest.java
```diff
@@ -95,26 +101,30 @@ public X509CertRequest(String csr, SpiffeUriManager spiffeUriManager,
...
 
         // save the spiffe uri manager and certificate data validator objects
 
         this.spiffeUriManager = spiffeUriManager;
         this.certificateDataValidator = certificateDataValidator;
     }
 
+    private String extractInstanceId(final String x509CertInstanceId) {
+
+        if (!StringUtils.isEmpty(x509CertInstanceId)) {
```
havetisyan: why is this taking precedence over uri and dns name values? it should be the last resort in this file - only look at the data if it's not in the csr.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/InstanceCertManager.java
```diff
@@ -1274,6 +1280,14 @@ public void logX509Cert(final Principal principal, final String ip, final String
             certStore.log(principal, ip, provider, instanceId, x509Cert);
         } catch (Exception ignored) {
         }
+
+        // also log via X509CertUtils when a servicePrincipal is available
+        // (e.g. CN on the register path with no authenticated caller)
+
+        if (servicePrincipal != null && !servicePrincipal.isEmpty()) {
+            X509CertUtils.logCert(X509_CERT_LOGGER, principal, servicePrincipal, ip, provider,
```
havetisyan: not sure if I follow this one - don't we already log all certificates? wouldn't this cause an additional line for the same entry and cause confusion?

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/spiffe/impl/SpiffeUriTrustDomain.java
```diff
@@ -37,15 +37,37 @@ public class SpiffeUriTrustDomain implements SpiffeUriValidator {
     private static final String SPIFFE_TRUST_DOMAIN = System.getProperty(SPIFFE_PROP_TRUST_DOMAIN, "athenz.io");
 
     /**
-     * Service Cert URI: spiffe://<trustDomain>/ns/<namespace>/sa/<domainName>.<serviceName>
-     *     Example: spiffe://athenz.io/ns/prod/sa/athenz.api
+     * Service Cert URI: accepts Form 1 (dotted SA) and Form 2 (dot-free SA):
```
havetisyan: we should not overload this implementation since it's major change where the domain is no longer present.
instead the way it works in athenz, is you define multiple spiffe validators and once configured, the server goes through all of them. this way we don't get the unexpected spiffe validation without the domain if one didn't want it.

so define a new class for the AmbientMode

havetisyan (changes requested):
Sasi, let's talk about this again. if there is no cn and the spiffe uri has no domain and just a service name - how do we distinguish between different identities?

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/X509CertRequest.java
psasidhar: changed the order.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/spiffe/impl/SpiffeUriTrustDomain.java
psasidhar: since we don't need dot-free support anymore, this class is untouched.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/InstanceCertManager.java
psasidhar: yes, it's a bug. fixed it.

havetisyan (approved):

PR #3457, opened 2026-08-20
add support for multiple aws association with domain

abvaidya:
# Description

# Contribution Checklist:


libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -78,7 +78,7 @@ public class JDBCConnection implements ObjectStoreConnection {
     private static final String SQL_GET_ACTIVE_DOMAIN_ID = "SELECT domain_id FROM domain WHERE name=? AND enabled=true;";
     @SuppressWarnings("SqlResolve")
     private static final String SQL_GET_DOMAINS_WITH_NAME = "SELECT name FROM domain WHERE name LIKE ?;";
-    private static final String SQL_GET_DOMAIN_WITH_AWS_ACCOUNT = "SELECT name FROM domain WHERE account=?;";
+    private static final String SQL_GET_DOMAIN_WITH_AWS_ACCOUNT = "SELECT name FROM domain WHERE FIND_IN_SET(?, account);";
```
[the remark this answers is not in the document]
abvaidya: values are trimmed and validateAwsAccountValue is fixed to handle `,,,`

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/metastore/DomainMetaStore.java
```diff
@@ -86,6 +87,30 @@ public interface DomainMetaStore {
      */
     void setAWSAccountDomain(final String domainName, final String awsAccountId) throws ServerResourceException;
 
+    /**
```
havetisyan: I don't think we need this method. Only the comments for the setAwsAccountDomain needs to be updated such that the implementation knows that it could be comma separated since it's something they have to enable anyway in the sysytem.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -1129,11 +1129,21 @@ void verifyDomainAwsAccountUniqueness(final String domainName, final String acco
...
+
+        if (domainOptions != null && domainOptions.getAllowMultipleAWSAccounts()) {
+            return;
+        }
         if (domainOptions != null && !domainOptions.getEnforceUniqueAWSAccounts()) {
             return;
         }
-        uniquenessCheck(lookupDomainByCloudProvider(ObjectStoreConnection.PROVIDER_AWS, account), domainName,
-                "Account Id: " + account, caller);
+        for (String awsAccount : Utils.parseAwsAccounts(account)) {
```
havetisyan: at this point we only have a single aws account id since if we already allow multiple aws accounts, then we have already returned from line 1138 above. so just need to include that detail in the comment and skip creating a set, etc

servers/zms/schema/zms_server.sql
```diff
@@ -29,7 +29,7 @@ CREATE TABLE IF NOT EXISTS `zms_server`.`domain` (
   `enabled` TINYINT(1) NOT NULL DEFAULT 1,
   `audit_enabled` TINYINT(1) NOT NULL DEFAULT 0,
   `ypm_id` INT UNSIGNED NOT NULL DEFAULT 0,
-  `account` VARCHAR(128) NOT NULL DEFAULT '',
+  `account` VARCHAR(2048) NOT NULL DEFAULT '',
```
havetisyan: the PR is missing the update workbench file

havetisyan (changes requested):

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
```diff
@@ -2614,7 +2642,7 @@ void validateDomainValues(Domain domain) {
         validateIntegerValue(domain.getFeatureFlags(), "featureFlags");
 
         validateString(domain.getApplicationId(), TYPE_COMPOUND_NAME, caller);
-        validateString(domain.getAccount(), TYPE_COMPOUND_NAME, caller);
+        validateAwsAccountValue(domain.getAccount(), caller);
```
havetisyan: before we were guaranteed that account is a valid value with our schema. now that we accept a comma separated list, it's possible that we'll end up with " acct, acct2" in the list since our Utils function trims all spaces and then we validate each one separately. that may cause other issues later. so we should probably regenerate the list without any spaces to be sure

servers/zms/schema/zms_server.sql
abvaidya: added updated mwb file

servers/zms/src/main/java/com/yahoo/athenz/zms/ZMSImpl.java
abvaidya: fixed

havetisyan (approved):

PR #3465, opened 2026-08-25
MSD - policy snapshot usage recording and retrieval; snapshot force delete flag

ArtjomsPorss:
Add transport policy snapshot usage post and get endpoints.
- `POST  /domain/{domainName}/service/{serviceName}/snapshot/{snapshotName}/usage` — record usage.
  Snapshot name goes in a `TransportPolicySnapshotUsageRequest` body, following
  `createTransportPolicySnapshot`, so the body can gain fields later without an API change.
- `GET  /domain/{domainName}/service/{serviceName}/snapshot/{snapshotName}/usage` — read usage.

GET response contains TransportPolicySnapshotUsage, with a list of principal + time containing objects that were submitted via POST endpoint. Response also contains boolean `partial` - which can be set to true when server cannot return full list of principals. Along with warning - can be used for reasoning code so that consumer knows the reason for partial response.

POST endpoint authorizes on a new `msd.RecordSnapshotHeartbeat` action.

`DeleteTransportPolicySnapshot` now has `force` parameter, which is the only signature change in the go client.
On the Java side, `MSDClient` now exposes the snapshot APIs, including a
`deleteTransportPolicySnapshot` overload without `force`.


core/msd/src/main/rdl/Snapshot.rdli
```diff
@@ -56,11 +56,13 @@ resource TransportPolicySnapshots GET "/domain/{domainName}/service/{serviceName
...
 }
 
-// Api to get a specific transport policy snapshot by name
+// Api to get a specific transport policy snapshot by name. When reportUsage is true the caller is
+// recorded as having used this snapshot; otherwise previously reported usage is returned in lastUsage.
 resource TransportPolicySnapshot GET "/domain/{domainName}/service/{serviceName}/snapshot/{snapshotName}" (name=getTransportPolicySnapshot) {
     DomainName domainName; // name of the domain
     EntityName serviceName; // name of the service
     EntityName snapshotName; // name of the snapshot
+    Bool reportUsage (optional, default=false, header="MSD-Report-Snapshot-Usage"); // report the caller's use of this snapshot instead of returning reported usage
```
psasidhar: I suggest renaming this to "recordUsage", which will be less ambiguous in the api.

reportUsage=true --> can mean give me a report of the usage tracked so far, or it can mean, "i am reporting the usage now" . Ambiguos. 

where as "recordUsage=true" would clearly mean - store my usage via this GET call I am making now.

ArtjomsPorss: renamed to `recordUsage` as recommended.
Also renamed wherever it was referenced.

Also renamed the GET header param to `MSD-Record-Snapshot-Usage` - instead of `MSD-Report-Snapshot-Usage` - to conform with the `recordUsage`.

clients/go/msd/client.go
```diff
@@ -1212,10 +1212,13 @@ func (client MSDClient) GetTransportPolicySnapshots(domainName DomainName, servi
 	}
 }
 
-func (client MSDClient) GetTransportPolicySnapshot(domainName DomainName, serviceName EntityName, snapshotName EntityName) (*TransportPolicySnapshot, error) {
+func (client MSDClient) GetTransportPolicySnapshot(domainName DomainName, serviceName EntityName, snapshotName EntityName, recordUsage *bool) (*TransportPolicySnapshot, error) {
```
yosrixp: Can we have another go func and  keep it backward compatible ? Only TF provider needs this method with the extra param while MSD controller doesn't really need to pass the recordUsage param.
I'm expecting to see the MSD go client to have

```
func (client MSDClient) GetTransportPolicySnapshot(domainName DomainName, serviceName EntityName, snapshotName EntityName, recordUsage ...*bool) (*TransportPolicySnapshot, error) {}

// Existing caller 
snapshot, err := client.GetTransportPolicySnapshot(
    domain,
    service,
    snapshot,
)

// New caller
recordUsage := true

snapshot, err := client.GetTransportPolicySnapshot(
    domain,
    service,
    snapshot,
    &recordUsage,
)
```

clients/go/msd/client.go
```diff
@@ -1244,11 +1247,11 @@ func (client MSDClient) GetTransportPolicySnapshot(domainName DomainName, servic
 	}
 }
 
-func (client MSDClient) DeleteTransportPolicySnapshot(domainName DomainName, serviceName EntityName, snapshotName EntityName, resourceOwner string) error {
+func (client MSDClient) DeleteTransportPolicySnapshot(domainName DomainName, serviceName EntityName, snapshotName EntityName, force *bool, resourceOwner string) error {
```
yosrixp, line 1247: same applies here and you can make the *bool as the last param of this function.

clients/go/msd/client.go
ArtjomsPorss: The clients are generated by rdl - is there a way to specify the format you recommend to the generator?

If not, the recordUsage can be a separate endpoint. This way the existing GetTransportPolicySnapshot contract won't be affected by RDL.

One thing it affects though - we should probably separate get snapshot (with snapshot usage) as separate endpoint. To stop returning all snapshot usage data on every get snapshot request.

core/msd/src/main/rdl/Snapshot.rdli
ArtjomsPorss: Added a dedicated POST and GET endpoints for snapshot usage, removed the GET header param on GET snapshot - not needed anymore.

clients/go/msd/client.go
ArtjomsPorss: as per discussion, there are 2 dedicated endpoints for snapshot usage. so these endpoints are not affected anymore

clients/go/msd/client.go
ArtjomsPorss: this is the only change that stays in go client as is. As per discussion, it is ok for go client to have breaking changes. 
Java client have a proper overloaded method as you requested @yosrixp

core/msd/src/main/java/com/yahoo/athenz/msd/MSDSchema.java
```diff
@@ -1185,6 +1206,45 @@ private static Schema build() {
...
 ;
 
+        sb.resource("TransportPolicySnapshotUsageRequest", "POST", "/domain/{domainName}/service/{serviceName}/snapshot/{snapshotName}/usage")
+            .comment("Api for a caller to record that it is using a snapshot. Callers are expected to repeat this while the snapshot remains in use rather than on every fetch. The principal is taken from the caller's identity and the time is stamped by the server.")
+            .name("recordTransportPolicySnapshotUsage")
+            .pathParam("domainName", "DomainName", "name of the domain")
+            .pathParam("serviceName", "EntityName", "name of the service")
+            .pathParam("snapshotName", "EntityName", "name of the snapshot")
+            .input("usage", "TransportPolicySnapshotUsageRequest", "snapshot usage request object")
+            .auth("msd.RecordSnapshotHeartbeat", "{domainName}:service.{serviceName}")
```
ArtjomsPorss, line 1216: @yosrixp, I added separate authorization action `msd.RecordSnapshotHeartbeat`

yosrixp (approved):
Looks good

psasidhar (approved):

PR #3468, opened 2026-08-28
Add Java Crypki library so ZTS can mint X.509 certificates in-process.

MartinTrojans:
# Description
Moves the HTTP Crypki client out of ZTS and adds pluggable KMS/HSM backends without changing existing HttpCertSignerFactory configuration.

# Contribution Checklist:

### DIFF
--- a/libs/java/crypki/pom.xml
+++ b/libs/java/crypki/pom.xml
@@ -27,7 +27,7 @@
   <packaging>jar</packaging>
 
   <properties>
-    <code.coverage.min>0.9750</code.coverage.min>
+    <code.coverage.min>1.00</code.coverage.min>
   </properties>
 
   <dependencies>
--- a/libs/java/crypki/pom.xml
+++ b/libs/java/crypki/pom.xml
@@ -1,0 +1,86 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    Copyright The Athenz Authors
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+        http://www.apache.org/licenses/LICENSE-2.0
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
+    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+    See the License for the specific language governing permissions and
+    limitations under the License.
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>com.yahoo.athenz</groupId>
+    <artifactId>athenz</artifactId>
+    <version>1.12.48-SNAPSHOT</version>
+    <relativePath>../../../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>athenz-crypki</artifactId>
+  <name>athenz-crypki</name>
+  <description>Generic Athenz X.509 Crypki signer SPI with pluggable backends</description>
+  <packaging>jar</packaging>
+
+  <properties>
+    <code.coverage.min>0.9750</code.coverage.min>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>com.yahoo.athenz</groupId>
+      <artifactId>athenz-auth-core</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>com.yahoo.athenz</groupId>
+      <artifactId>athenz-server-common</artifactId>
+      <version>${project.parent.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcprov-jdk18on</artifactId>
+      <version>${bouncycastle.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.bouncycastle</groupId>
+      <artifactId>bcpkix-jdk18on</artifactId>
+      <version>${bouncycastle.version}</version>
+    </dependency>
+    <dependency>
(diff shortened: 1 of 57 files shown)


libs/java/server_gcp_common/src/main/java/io/athenz/server/gcp/common/cert/impl/GcpKmsClient.java
```diff
@@ -0,0 +1,79 @@
...
+        this.caCertPath = caCertPath;
+    }
+
+    @Override
+    public byte[] sign(String keyId, byte[] data, String signingAlgorithm) {
+        try {
+            return client.asymmetricSign(AsymmetricSignRequest.newBuilder()
+                    .setName(keyId)
+                    .setData(ByteString.copyFrom(data))
+                    .build()).getSignature().toByteArray();
```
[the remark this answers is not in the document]
MartinTrojans: Addressed. `GcpKmsClient.sign` now hashes the TBS bytes from `signingAlgorithm` and sets `AsymmetricSignRequest.digest` (`sha256` / `sha384` / `sha512`). The default `SHA256withRSA` path no longer uses `setData`. Missing or unsupported algorithms are rejected.

We did not add a raw/`data` path. ZTS only uses digest algorithms, and Cloud KMS accepts `data` only for raw/external-key versions.

libs/java/crypki/src/main/java/com/yahoo/athenz/crypki/x509/X509CertificateMinter.java
```diff
@@ -0,0 +1,203 @@
...
+            throw new CrypkiException("Unable to parse CSR");
+        }
+        int validitySeconds = request.getValiditySeconds() > 0
+                ? Math.min(request.getValiditySeconds(), maxValiditySeconds) : maxValiditySeconds;
+        Instant notAfter = Instant.now();
+        Instant notBefore = notAfter.minusSeconds(BACKDATE_SECONDS);
+        notAfter = notBefore.plusSeconds(validitySeconds + BACKDATE_SECONDS);
+        try {
+            JcaPKCS10CertificationRequest jcaReq = new JcaPKCS10CertificationRequest(certReq);
+            PublicKey publicKey = jcaReq.getPublicKey();
```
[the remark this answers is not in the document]
MartinTrojans: Addressed. `X509CertificateMinter` now verifies the PKCS#10 signature with the CSR's embedded public key (`isSignatureValid`) before building the certificate, matching Go Crypki's `CheckSignature()`.

libs/java/server_aws_common/src/main/java/io/athenz/server/aws/common/cert/impl/AwsCloudHsmClient.java
```diff
@@ -0,0 +1,273 @@
...
+
+    static SigningKey loadCloudHsmJce(String label, char[] pin, X509Certificate caCertificate)
+            throws Exception {
+        Class<?> providerClass = Class.forName("com.amazonaws.cloudhsm.jce.provider.CloudHsmProvider");
+        Provider provider = (Provider) providerClass.getDeclaredConstructor().newInstance();
+        if (Security.getProvider(provider.getName()) == null) {
+            Security.addProvider(provider);
+        }
+        if (provider instanceof AuthProvider) {
+            ((AuthProvider) provider).login(null, callbacks -> {
```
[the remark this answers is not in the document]
MartinTrojans: Addressed. If a CloudHSM provider is already registered in `Security`, we reuse that instance for `login` instead of authenticating a newly constructed, unregistered provider. Later `KeyStoreWithAttributes.getInstance("CloudHSM")` then uses the same authenticated provider.

libs/java/crypki/pom.xml
```diff
@@ -0,0 +1,86 @@
...
+    <relativePath>../../../pom.xml</relativePath>
+  </parent>
+
+  <artifactId>athenz-crypki</artifactId>
+  <name>athenz-crypki</name>
+  <description>Generic Athenz X.509 Crypki signer SPI with pluggable backends</description>
+  <packaging>jar</packaging>
+
+  <properties>
+    <code.coverage.min>0.9750</code.coverage.min>
```
havetisyan: any new code should be full 100% coverage. so let's review and add any missing tests.

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/crypki/JavaCrypkiCertSignerFactory.java
```diff
@@ -0,0 +1,54 @@
...
+ * Opt-in factory for the in-process Java Crypki backends. Existing
+ * {@link HttpCertSignerFactory} deployments are unchanged unless
+ * {@code athenz.zts.java_crypki_enabled=true} and this factory is
+ * selected via {@code athenz.zts.cert_signer_factory_class}.
+ */
+public class JavaCrypkiCertSignerFactory implements CertSignerFactory {
+
+    @Override
+    public CertSigner create() {
+        if (!Boolean.parseBoolean(System.getProperty(ZTSConsts.ZTS_PROP_JAVA_CRYPKI_ENABLED, "false"))) {
```
havetisyan: since you're handling the setup nicely with the factory class, there is no need for this attribute anymore. if set the property setting to the JavaCrypkiCertSignerFactory then I'm implying that I want to enable the functionality

libs/java/crypki/pom.xml
MartinTrojans: Addressed. Raised code.coverage.min to 1.00 and added tests for the remaining X509CertificateMinter / HttpCrypkiSigner paths (CSR parse/sign errors, CSR signature verify failures, PKCS#11 signer fallback, missing SAN attributes, and JSON serialize failure).

servers/zts/src/main/java/com/yahoo/athenz/zts/cert/impl/crypki/JavaCrypkiCertSignerFactory.java
MartinTrojans: Addressed. Removed athenz.zts.java_crypki_enabled. Selecting JavaCrypkiCertSignerFactory (plus athenz.zts.java_crypki_factory_class for the backend) is now the opt-in.

havetisyan (approved):

PR #3477, opened 2026-09-02
Feat: ID-JAG with Multiple Domain Supported for Full ID-JAG => AccessToken Exchange

mlajkim:
# Description

PR [#3407](https://github.com/AthenZ/athenz/pull/3407) introduced Access Tokens with scopes from multiple domains. This allows an Access Token to carry both the permission for its current audience and permissions that may be exchanged for downstream services.

However, `ID-JAG` fetched by Athenz Server (as an external IdP AS) still supports only a single scope domain. This means an `ID-JAG` cannot be used to issue a multi-domain Access Token.

## What's done?

This PR allows an ID-JAG to be issued with role scopes from multiple domains:

```sh
curl -sS -X POST "$_zts_url" \
  --cert ./keys/human-idjag-learner-claude.crt \
  --key ./keys/human-idjag-learner-claude.key \
  -H 'Content-Type: application/x-www-form-urlencoded' \
  --data-urlencode 'grant_type=urn:ietf:params:oauth:grant-type:token-exchange' \
  --data-urlencode 'requested_token_type=urn:ietf:params:oauth:token-type:id-jag' \
  --data-urlencode 'subject_token_type=urn:ietf:params:oauth:token-type:id_token' \
  --data-urlencode "subject_token=$_id_token" \
  --data-urlencode "scope=mcp-hub:role.access api:role.getter" \
  --data-urlencode "audience=$_zts_audience"
```

```sh
# {
#   "kid": "athenz-zts-server-56f8855d5b-bh2f4",
#   "typ": "oauth-id-jag+jwt",
#   "alg": "RS256"
# }
# {
#   "sub": "human.idjag-learner",
#   "aud": "https://athenz-zts-server.athenz:4443/zts/v1",
#   "scp": [
#     "mcp-hub:role.access",
#     "api:role.getter"
#   ],
#   "ver": 1,
#   "auth_time": 1788339728,
#   "scope": "mcp-hub:role.access api:role.getter",
#   "iss": "https://athenz-zts-server.athenz:4443/zts/v1",
#   "exp": 1788346928,
#   "iat": 1788339728,
#   "jti": "e91c3db6-8ef3-4898-b8f4-249f7866a9d9",
#   "client_id": "human.idjag-learner.claude"
# }
```

# Contribution Checklist:

### DIFF
--- a/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
+++ b/servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
@@ -3465,26 +3465,9 @@ public class ZTSImpl implements ZTSHandler {
         // before using any of our values let's validate that they
         // match our schema
 
-        final String domainName = tokenScope.getDomainName();
-        setRequestDomain(ctx, domainName);
-        validate(domainName, TYPE_DOMAIN_NAME, principalDomain, caller);
-
-        String[] requestedRoles = tokenScope.getRoleNames(domainName);
-        if (requestedRoles == null) {
-            throw requestError("Scope value does not contain any roles", caller, domainName, principalDomain);
-        }
-
-        for (String requestedRole : requestedRoles) {
-            validate(requestedRole, TYPE_ENTITY_NAME, principalDomain, caller);
-        }
-
-        // first retrieve our domain data object from the cache
-
-        DataCache data = dataStore.getDataCache(domainName);
-        if (data == null) {
-            setRequestDomain(ctx, ZTSConsts.ZTS_UNKNOWN_DOMAIN);
-            throw notFoundError("No such domain: " + domainName, caller, ZTSConsts.ZTS_UNKNOWN_DOMAIN, principalDomain);
-        }
+        final Set<String> scopeDomainNames = tokenScope.getDomainNames();
+        final String requestDomainName = tokenScope.getDomainName();
+        setRequestDomain(ctx, requestDomainName);
 
         // if the token was issued by ZTS then the identity is the subject
         // otherwise, we need to extract the identity from the token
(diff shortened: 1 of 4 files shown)


servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3469,26 +3465,9 @@ AccessTokenResponse processJAGTokenIssueRequest(ResourceContext ctx, Principal p
...
-
-        // first retrieve our domain data object from the cache
-
-        DataCache data = dataStore.getDataCache(domainName);
-        if (data == null) {
-            setRequestDomain(ctx, ZTSConsts.ZTS_UNKNOWN_DOMAIN);
-            throw notFoundError("No such domain: " + domainName, caller, ZTSConsts.ZTS_UNKNOWN_DOMAIN, principalDomain);
-        }
+        final Set<String> scopeDomainNames = tokenScope.getDomainNames();
+        final String requestDomainName = tokenScope.getDomainName();
```
havetisyan: if there are multiple domains, then the requestDomainName is going to be returned as null. So that means the next call setRequestDomain is going to be null which is probably not what we want. maybe in this scenario we default to the first domain as the request domain

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3498,39 +3477,64 @@ AccessTokenResponse processJAGTokenIssueRequest(ResourceContext ctx, Principal p
                 identityProvider.getTokenIdentity(subjectToken);
         if (StringUtil.isEmpty(subjectIdentity)) {
             LOGGER.error("processJAGTokenIssueRequest: unable to extract subject identity from token");
-            throw requestError("Invalid subject token - missing subject", caller, domainName, principalDomain);
+            throw requestError("Invalid subject token - missing subject", caller, requestDomainName, principalDomain);
```
havetisyan, line 3480: same as above requestDomainName is null if we have multiple domains.

havetisyan (changes requested):

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
```diff
@@ -3494,39 +3477,64 @@ AccessTokenResponse processJAGTokenIssueRequest(ResourceContext ctx, Principal p
...
-            roleList.add(ResourceUtils.roleResourceName(domainName, subjectRole));
+            for (String requestedRole : requestedRoles) {
+                if (!authorizer.access(ZTSConsts.ZTS_ACTION_JAG_EXCHANGE,
+                        ResourceUtils.roleResourceName(domainName, requestedRole), principal, null)) {
+                    LOGGER.error("processJAGTokenIssueRequest: access check failure for {} - {}:role.{}",
+                            principalName, domainName, requestedRole);
+                    throw forbiddenError("Principal not authorized for token exchange for the requested role",
+                            caller, domainName, principalDomain);
+                }
+            }
```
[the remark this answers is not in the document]
havetisyan: it's a good observation even though you didn't change the logic here. but it kind of makes sense, if we are not going to give you back the full set why do we check everything. I think at some point we had that everything had to be present and then we changed it to be allowed to return a subset but didn't make the change here. so we should change it and only look for subject roles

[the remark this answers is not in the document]
mlajkim: RequestedRole => subject roles here: https://github.com/AthenZ/athenz/pull/3477/changes/883bc7859b51894cfb9f87a4e7cfc96d0bf47109

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: Fixed: https://github.com/AthenZ/athenz/pull/3477/changes/ca60e56987e6850157b5d83ab23671e9ff1e4912

servers/zts/src/main/java/com/yahoo/athenz/zts/ZTSImpl.java
mlajkim: Fixed: https://github.com/AthenZ/athenz/pull/3477/changes/ca60e56987e6850157b5d83ab23671e9ff1e4912

havetisyan (approved):

PR #3487, opened 2026-09-04
add self-service API endpoints & update clients

tsultanov00:
# Description
Adds two new ZMS endpoints that let any authenticated principal discover self-service
roles and groups across all domains, and see their own membership status inline. This
backs the new self-service UI (search, request, and "My Roles & Groups").

## API contract (RDL)
New `core/zms/src/main/rdl/SelfServe.rdli` (included from `ZMS.rdl`):
- `GET /zms/v1/selfserve/role?substring={substring}&memberOnly={memberOnly}`
- `GET /zms/v1/selfserve/group?substring={substring}&memberOnly={memberOnly}`

Behaviour:
- `substring` — case-insensitive match. Roles match on **name or description**; groups
  match on **name only** (groups carry no description). Empty substring returns all.
- `memberOnly` (default `false`) — when `true`, returns only objects the caller is already
  a member of or has a pending request for (powers the "My Roles & Groups" view).
- `authenticate;` (not `authorize;`) — self-service objects are discoverable/requestable by
  any authenticated user by design, so no per-object authorization check is applied. The
  membership overlay is always scoped to the calling principal and derived server-side, so
  a caller can never see another user's membership.

New response type `SelfServeObject` returns everything the UI needs in one shot:
`domainName`, `name`, `description` (roles only — groups carry no description),
`selfRenew`/`selfRenewMins`, `reviewEnabled`, `auditEnabled`, `deleteProtection`,
`memberExpiryDays` (role/group-level max, `0` = unlimited) and `domainMemberExpiryDays`
(domain-level max, `0` = unlimited), plus the caller-scoped overlay: `memberStatus`
(`member` | `pending` | `none`), `expiration`, and `inheritedFrom` (the group a role
membership is inherited through).

## Server layer
- `ZMSImpl` / `ZMSHandler` / `ZMSResources`: implement `getSelfServeRoles` /
  `getSelfServeGroups`, extract the calling principal from the request context, and pass
  `principal` + `memberOnly` down. No extra authorization step (see rationale above).
- `DBService`: threads `principal` and `memberOnly` through to the store.

## Data layer
- `ObjectStoreConnection`: updated `getSelfServeRoles` / `getSelfServeGroups` signatures to
  take `principal` and `memberOnly`.
- `JDBCConnection`: single-query implementation for efficiency at scale:
  - Fetches all matching self-service objects and, via `LEFT JOIN`s (role_member /
    pending_role_member / a derived table for group-inherited roles; principal_group_member
    / pending for groups), overlays the caller's membership, expiration, and inherited
    source — **no N+1 per-object lookups**.
  - Returns both `memberExpiryDays` (the role/group `member_expiry_days`) and
    `domainMemberExpiryDays` (the domain `member_expiry_days`) as raw values (`0` =
    unlimited). The **effective cap is derived by the caller/UI** (the minimum of the two,
    treating `0` as unlimited), keeping business logic out of the storage layer.
  - `memberOnly` is applied as an optional SQL filter; results are ordered by domain then name.

## Generated clients
Regenerated stubs from the RDL:
- Go client — `clients/go/zms` (`client.go`, `model.go`, `zms_schema.go`)
- Java client + models — `ZMSRDLGeneratedClient`, `SelfServeObject(s)`, `ZMSSchema`

# Contribution Checklist:


clients/java/zms/src/main/java/com/yahoo/athenz/zms/ZMSRDLGeneratedClient.java
```diff
@@ -4592,6 +4592,72 @@ public ReviewObjects getGroupsForReview(String principal) throws URISyntaxExcept
         }
     }
 
+    public SelfServeObjects getSelfServeRoles(String substring, Boolean memberOnly) throws URISyntaxException, IOException {
```
havetisyan: for these two new methods, please update ZMSClient.java and provide the wrapper methods with test cases.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -720,6 +720,40 @@ public class JDBCConnection implements ObjectStoreConnection {
...
+            + " JOIN domain ON role.domain_id=domain.domain_id"
+            + " LEFT JOIN role_member rm ON rm.role_id=role.role_id AND rm.principal_id=?"
+            + " LEFT JOIN pending_role_member prm ON prm.role_id=role.role_id AND prm.principal_id=?"
+            + " LEFT JOIN (SELECT rmg.role_id AS role_id, MIN(gp.name) AS inherited_from,"
+            + " MIN(rmg.expiration) AS inherited_expiration FROM principal_group_member pgm"
+            + " JOIN principal_group pg ON pg.group_id=pgm.group_id JOIN domain gd ON gd.domain_id=pg.domain_id"
+            + " JOIN principal gp ON gp.name=CONCAT(gd.name, ':group.', pg.name)"
+            + " JOIN role_member rmg ON rmg.principal_id=gp.principal_id WHERE pgm.principal_id=? GROUP BY rmg.role_id) inh"
+            + " ON inh.role_id=role.role_id WHERE role.self_serve=true AND role.trust=''"
+            + " AND (LOWER(role.name) LIKE ? OR LOWER(role.description) LIKE ?)";
```
havetisyan: no need for LOWER(role.name). the role name is always guaranteed to be lower case in athenz. the description one is needed.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -720,6 +720,40 @@ public class JDBCConnection implements ObjectStoreConnection {
...
+            + " principal_group.name AS group_name, principal_group.self_renew, principal_group.self_renew_mins,"
+            + " principal_group.review_enabled, principal_group.audit_enabled, principal_group.delete_protection,"
+            + " principal_group.member_expiry_days,"
+            + " domain.member_expiry_days AS domain_member_expiry_days,"
+            + " gm.principal_id AS member_principal, gm.expiration AS member_expiration,"
+            + " pgm.principal_id AS pending_principal, pgm.expiration AS pending_expiration"
+            + " FROM principal_group JOIN domain ON principal_group.domain_id=domain.domain_id"
+            + " LEFT JOIN principal_group_member gm ON gm.group_id=principal_group.group_id AND gm.principal_id=?"
+            + " LEFT JOIN pending_principal_group_member pgm ON pgm.group_id=principal_group.group_id AND pgm.principal_id=?"
+            + " WHERE principal_group.self_serve=true AND LOWER(principal_group.name) LIKE ?";
```
havetisyan: no need for LOWER(principal_group.name). the group name is always guaranteed to be lower case in athenz.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -8302,6 +8336,120 @@ public ReviewObjects getGroupsForReview(String principal) throws ServerResourceE
         return new ReviewObjects().setList(reviewRoles);
     }
 
+    static String selfServeSearchPattern(final String substring) {
+        return "%" + (substring == null ? "" : substring.toLowerCase()) + "%";
```
havetisyan: substring was already normalized in ZMS and was lowercase so no need to do the same here.

havetisyan: so is substring is null, we generate a search string "%%" which is ok I guess, but since we already have a check for substring == null, we can just return either % or %substring%

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -720,6 +720,40 @@ public class JDBCConnection implements ObjectStoreConnection {
             + " role.domain_id=domain.domain_id JOIN role_member ON role.role_id=role_member.role_id"
             + " WHERE role_member.principal_id=? AND role_member.active=true AND role.name='admin')"
             + " ORDER BY domain.name, principal_group.name;";
+    private static final String SQL_GET_SELF_SERVE_ROLES = "SELECT domain.name AS domain_name, role.name AS role_name,"
```
havetisyan: I don't follow why we're overloading the same sql command to handle both requests: get the list of self-serve roles and get the list of self-serve roles where I'm a member. For the first request, where the princpal is null - this is a very simple SQL query with a join on the domain table to get the domain name and nothing else. So let's not overload and keep 2 separate sql commands

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -720,6 +720,40 @@ public class JDBCConnection implements ObjectStoreConnection {
...
+            + " MIN(rmg.expiration) AS inherited_expiration FROM principal_group_member pgm"
+            + " JOIN principal_group pg ON pg.group_id=pgm.group_id JOIN domain gd ON gd.domain_id=pg.domain_id"
+            + " JOIN principal gp ON gp.name=CONCAT(gd.name, ':group.', pg.name)"
+            + " JOIN role_member rmg ON rmg.principal_id=gp.principal_id WHERE pgm.principal_id=? GROUP BY rmg.role_id) inh"
+            + " ON inh.role_id=role.role_id WHERE role.self_serve=true AND role.trust=''"
+            + " AND (LOWER(role.name) LIKE ? OR LOWER(role.description) LIKE ?)";
+    private static final String SQL_GET_SELF_SERVE_ROLES_MEMBER_FILTER =
+            " AND (rm.principal_id IS NOT NULL OR prm.principal_id IS NOT NULL OR inh.role_id IS NOT NULL)";
+    private static final String SQL_GET_SELF_SERVE_ROLES_SUFFIX = " ORDER BY domain.name, role.name;";
+    private static final String SQL_GET_SELF_SERVE_GROUPS = "SELECT domain.name AS domain_name,"
```
havetisyan: same comment as above with roles - 2 separate commands

core/zms/src/main/rdl/SelfServe.rdli
```diff
@@ -0,0 +1,60 @@
...
+type SelfServeObjects Struct {
+    Array<SelfServeObject> list; //list of self-service objects
+}
+
+// Fetch all self-service roles across all domains whose name or description contains the
+// given substring. The match is case-insensitive. If the substring is not specified, then
+// all self-service roles are returned. Any authenticated principal may make this call since
+// self-service roles are, by definition, discoverable and requestable by any user.
+resource SelfServeObjects GET "/selfserve/role?substring={substring}&memberOnly={memberOnly}" (name=GetSelfServeRoles) {
+    String substring (optional); //substring to match against the role name or description; if empty all self-service roles are returned
```
havetisyan: let's rename substring to matchString to be more clear in the api calls what it represents.

core/zms/src/main/rdl/SelfServe.rdli
```diff
@@ -0,0 +1,60 @@
...
+        ResourceError TOO_MANY_REQUESTS;
+    }
+}
+
+// Fetch all self-service groups across all domains whose name contains the given substring.
+// The match is case-insensitive. If the substring is not specified, then all self-service
+// groups are returned. Any authenticated principal may make this call since self-service
+// groups are, by definition, discoverable and requestable by any user.
+resource SelfServeObjects GET "/selfserve/group?substring={substring}&memberOnly={memberOnly}" (name=GetSelfServeGroups) {
+    String substring (optional); //substring to match against the group name; if empty all self-service groups are returned
```
havetisyan: same as above - let's call it matchString

servers/zms/schema/zms_server.sql
```diff
@@ -111,6 +111,7 @@ CREATE TABLE IF NOT EXISTS `zms_server`.`role` (
   `notify_details` VARCHAR(512) NOT NULL DEFAULT '',
   PRIMARY KEY (`role_id`),
   UNIQUE INDEX `uq_domain_role` (`domain_id` ASC, `name` ASC),
+  INDEX `idx_self_serve` (`self_serve` ASC),
```
havetisyan, line 114: why only roles and not in principal_groups?

servers/zms/schema/updates/update-20260909.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX `idx_self_serve` ON `zms_server`.`role` (`self_serve` ASC);
```
havetisyan, line 1: please review the steps in the schema_update.md in the schema directory. you're missing the updated mwb file

havetisyan (changes requested):

servers/zms/schema/zms_server.sql
tsultanov00: I tested the index on groups, it actually slowed the query down slightly. The query is fast enough to not warrant further optimisation. However, out of ~370k roles, ~1.4% were self service so there was a significant performance improvement with this index (~5x).

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
tsultanov00: Split it into two separate statements now and the method just picks one.

With regards to the simple SQL query, that won't quite work, since we need the caller's membership state in discovery mode too, not just for memberOnly. The Find tab uses memberStatus/inheritedFrom on each row to know if it's requestable, already yours (Leave), or pending (Cancel).

servers/zms/schema/zms_server.sql
havetisyan: it was only fast because we have large number of roles and not groups. But some other company might have large number of groups and it will be impacted by not having the index. So it should be created on both tables.

tsultanov00: understood, I will create an index on the groups table too

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -720,6 +720,44 @@ public class JDBCConnection implements ObjectStoreConnection {
             + " role.domain_id=domain.domain_id JOIN role_member ON role.role_id=role_member.role_id"
             + " WHERE role_member.principal_id=? AND role_member.active=true AND role.name='admin')"
             + " ORDER BY domain.name, principal_group.name;";
+    private static final String SQL_SELF_SERVE_ROLES_BASE = "SELECT domain.name AS domain_name, role.name AS role_name,"
```
havetisyan: this is still not simplified enough just to return the roles without any princpal details. Here is what I expect to see:

SELECT domain.name AS domain_name, role.name AS role_name,
         role.description, role.self_renew, role.self_renew_mins, role.review_enabled, role.audit_enabled,
         role.delete_protection, role.member_expiry_days,
         domain.member_expiry_days AS domain_member_expiry_days
FROM role
JOIN domain ON domain.domain_id = role.domain_id
WHERE role.self_serve = 1
  AND (role.name LIKE ? OR LOWER(role.description) LIKE ?)
ORDER BY domain.name, role.name;

that's all the query we need to return all the self serve roles that match the given matchString. we should have 2 completely separate SQL commands handling the cases.

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -720,6 +720,44 @@ public class JDBCConnection implements ObjectStoreConnection {
...
+            + " JOIN principal gp ON gp.name=CONCAT(gd.name, ':group.', pg.name)"
+            + " JOIN role_member rmg ON rmg.principal_id=gp.principal_id WHERE pgm.principal_id=? GROUP BY rmg.role_id) inh"
+            + " ON inh.role_id=role.role_id WHERE role.self_serve=true AND role.trust=''"
+            + " AND (role.name LIKE ? OR LOWER(role.description) LIKE ?)";
+    private static final String SQL_GET_SELF_SERVE_ROLES = SQL_SELF_SERVE_ROLES_BASE
+            + " ORDER BY domain.name, role.name;";
+    private static final String SQL_GET_SELF_SERVE_ROLES_MEMBER = SQL_SELF_SERVE_ROLES_BASE
+            + " AND (rm.principal_id IS NOT NULL OR prm.principal_id IS NOT NULL OR inh.role_id IS NOT NULL)"
+            + " ORDER BY domain.name, role.name;";
+    private static final String SQL_SELF_SERVE_GROUPS_BASE = "SELECT domain.name AS domain_name,"
```
havetisyan: same as above

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -8302,6 +8340,116 @@ public ReviewObjects getGroupsForReview(String principal) throws ServerResourceE
...
+        return matchString == null ? "%" : "%" + matchString + "%";
+    }
+
+    @Override
+    public SelfServeObjects getSelfServeRoles(String matchString, String principal, boolean memberOnly) throws ServerResourceException {
+
+        final String caller = "getSelfServeRoles";
+
+        final String searchPattern = selfServeSearchPattern(matchString);
+        final int principalId = StringUtil.isEmpty(principal) ? 0 : getPrincipalId(principal);
```
havetisyan, line 8400: the logic here is not quote right. if the principal was never referenced in the db, the getPrincpalid will return 0 which the rest of the files treats as a failure and the check is missing here. if the getPrincpalid returns 0, then we should return an empty set indicating the user is not part of any roles

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
```diff
@@ -8302,6 +8340,116 @@ public ReviewObjects getGroupsForReview(String principal) throws ServerResourceE
...
+        return new SelfServeObjects().setList(selfServeRoles);
+    }
+
+    @Override
+    public SelfServeObjects getSelfServeGroups(String matchString, String principal, boolean memberOnly) throws ServerResourceException {
+
+        final String caller = "getSelfServeGroups";
+
+        final String searchPattern = selfServeSearchPattern(matchString);
+        final int principalId = StringUtil.isEmpty(principal) ? 0 : getPrincipalId(principal);
```
havetisyan, line 8449: same as above - handle getPrincipalId(principal) returning as 0 - as failure and return an empty set - not part of any groups

havetisyan (changes requested):

libs/java/server_common/src/main/java/com/yahoo/athenz/common/server/store/impl/JDBCConnection.java
tsultanov00: added this sql query for the scenario when member only is false and principal ID is 0. Same for groups query.

PR #3489, opened 2026-09-07
add self-service UI for role/group management

tsultanov00:
# Self-Service Roles & Groups – UI

Adds a new **Self-Service** experience that lets any authenticated user discover, request, and manage self-service roles and groups from a single screen — without needing a domain admin to add them manually.

## New Self-Service page

A dedicated `/self-service` page with two tabs:
- **Find** – discover and request self-service roles/groups across all domains.
- **My Roles & Groups** – view and manage the roles/groups the logged-in user already belongs to.

image

> _Screenshot: Self-Service landing page (Find tab, empty search prompt)_

image

> _Screenshot: My Roles & Groups tab_

## Search & filtering

- Free-text search that matches **roles by name and description** and **groups by name**.
- Results are split into distinct **Roles** and **Groups** sections so users can tell the two apart.
- Each result row shows the **domain name directly under the role/group name** for immediate context.
- Each role/group name is a **hyperlink that opens the members page in a new tab** for quick inspection.
- A **domain filter dropdown** to narrow results across domains. It defaults to **"All domains"** and only offers that option until a search is run.
- Friendly **empty state** when nothing matches.

image

> _Screenshot: Search results split into Roles and Groups sections_

## Requesting access

- **Request Access modal** for submitting a membership request, requiring a justification.
- Roles support an **expiry date picker**; groups intentionally show **no expiry pickers** (self-renew model).
- **Multi-select**: users can select several roles and groups at once and request them together, with a breakdown of the selection.

image

> _Screenshot: Request Access modal_

## Managing existing memberships (My Roles & Groups)

- **Leave** an active membership directly from the list.
- **Cancel** a pending request.
- **Membership expiry** is shown for any membership with an expiration date, highlighted in red when it is expiring soon (within 14 days) and grey otherwise.
- **Inherited roles** show the **source group**, disable the **Leave** action, and provide a tooltip that names the source domain/group with a **link to the source group's members page**.

image

> _Screenshot: My Roles & Groups list (active + pending states)_

image

> _Screenshot: Inherited role showing source group with disabled Leave + tooltip link_

## Extending membership

- **Extend Membership modal** with a calendar date picker, available for **any active membership (roles and groups)** the user holds directly.
- Extend is only offered when the membership can actually expire — i.e. the role/domain sets a `member_expiry_days` cap **or** the role/group self-renews. Permanent memberships (no cap and no self-renew) and **inherited** memberships don't show Extend.
- The picker is **capped at the maximum allowed expiry** — the earliest of the self-renew window (now + self-renew minutes) and the role/domain expiry policy (the effective minimum of the domain and role/group settings). The exact cutoff is shown as a **date/time** and out-of-range dates are greyed out.
- On submit, the outcome reflects what the server actually did: changes that apply immediately show **"Membership extended"**, while changes that require approval are submitted as a **request** ("Request submitted").

image

> _Screenshot: Extend Membership modal (calendar capped at max expiry)_

# Contribution Checklist:


havetisyan:
**Blockers**

- Backend failures render as "you have nothing." SelfServiceView.js writes errorMessage in loadMemberships (:151) and onSearch (:231), but render() only ever passes it to the three modals (:650, :667, :683); Alert at :695 is success-only. A failed membership load therefore displays "You do not currently hold any self-service roles or groups" (:543), and a failed search displays "No self-service roles or groups match '…'" (:483). Needs a page-level error banner.

- Partial failure is reported as success, and the failed items are lost. handleRequestSubmit (:316-327): when some requests fail, it falls through to showSuccess — a green alert titled "N requests submitted" whose body is only ${failed.length} request(s) could not be sent. The per-item result.error strings are discarded, the user is never told which items failed, and showSuccess (:286) clears requestItems and selected, so there's no retry path short of searching and re-selecting by hand. Same in handleLeaveSubmit (:356).

- The all-failed error is erased before it's shown. Both handlers call refreshAfterAction() before setState({errorMessage}) (:308-315). refreshAfterAction → onSearch → .then sets errorMessage: null (:225), resolving after. With an active query, a total failure shows no error at all.

- No in-flight guard — double-submit duplicates everything. AddModal already supports saving (AddModal.js:100), but none of the three modals is passed it and the view has no in-flight flag. Two clicks fire two complete Promise.all fan-outs.

- Unbounded parallel fan-out. Promise.all(items.map(...)) (:305, :334) issues one PUT per selected item with no concurrency cap and no cap on selection size. Since the ZMS endpoint accepts no limit, a broad search plus select-all is an arbitrary number of concurrent writes through the UI server.

- Extend can submit a past expiration. ExtendMembershipModal.onSubmit (:144) checks only that the field is non-empty — no > now, no <= maxDate — and selfServeZms.js:200 passes params.expiration through untouched. DateUtils.isBeforeCurrenTime already exists for this.

**Should fix**

- Dead method-name probing. selfServeZms.js:26-42 probes nine candidate names; seven can never resolve (getSelfServeResources, searchSelfServe, getSelfServe, …RoleList, getSelfserveRoles, …GroupList, getSelfserveGroups). rdl-rest generates exactly getSelfServeRoles / getSelfServeGroups from the name= annotations. The entire COMBINED_SEARCH_METHODS branch (:117-121) is unreachable. Please collapse to the two real names — the speculative fallbacks read as uncertainty about an API that is pinned.

- The pagination loop is dead and unsafe if revived. searchPages (:71-100) loops 20 times on mapped.next, but SelfServeObjects has no next field and neither endpoint declares skip/next/limit, so rdl-rest drops those params. It always exits after one iteration today; if ZMS ever adds next, the loop re-requests identical params and fetches the same page 20 times. Same for domain in toZmsSearchParams (selfServeContract.js:88) — not a declared param, hence the client-side applyDomainFilter.

- selfRenewMins is taken from the client to compute an expiration. SelfServiceView.runAction (:267) sends item.selfRenewMins, and selfServeZms.extendExpiration (:203-206) does Number(params.selfRenewMins) || 20160 — a client-supplied renewal window plus an unexplained 14-day magic number. ZMS should clamp, but the server util shouldn't trust it; read it back or drop the fallback.

- "Expiring soon" has two different definitions. 14 days, midnight-normalised, in SelfServiceResultRow.js:258; 30 days, measured off a live Date.now() with Math.round, in SelfServiceView.js:531, rendered as "expiring within 30 days" (:536). The PR description says 14. Both also count already-expired memberships, so a lapsed row reads -37 days left (:263) — and days === 1 reads "1 days left".

- Groups do get the expiration you tell users they don't. RequestAccessModal.js:267 says "Expiration and reminder apply to roles only", but runAction sends the same expiration for every item and membershipBody (selfServeZms.js:190) includes it for groups, gating only reviewReminder on isGroup. Conversely, includesRoles (:216, :238) hides both pickers, so a groups-only selection can't set an expiry at all. Server comment and UI copy directly contradict each other — worth settling which behavior is intended.

- The domain filter follows the user across tabs invisibly. renderMyRoles filters on this.state.domain (:506), onTabClick (:156) clears selected but not domain, and SelfServiceSearchBar — the only place the filter is visible or clearable — renders only on the Find tab (:611). The tab label keeps the unfiltered server count while the list below is silently truncated.

- Search race. onSearch (:213) has no request token or abort; two in-flight searches resolve in arbitrary order and the later-resolving one wins, while submittedQuery already reads as the newest. Also, onDomainChange (:174) guards on submittedQuery but onSearch reads this.state.query, so changing the domain submits a term the user typed but never entered — or wipes results if they cleared the box.

- FlatPicker snapshots maxDate at mount. ExtendMembershipModal:162 recomputes effectiveMaxDate() every render, but FlatPicker initialises flatpickr only in componentDidMount. The "You can extend until…" text (:204) drifts past what the picker actually allows — visible with short selfRenewMins windows.

- hasExpiryPolicy is stricter than the stated rule. SelfServiceResultRow.js:166 requires selfRenew && selfRenewMins > 0, but the contract defaults selfRenewMins to 0 (selfServeContract.js:46), so a selfRenew: true object without minutes hides Extend entirely — while the server has a fallback for exactly that case. The PR's own fixture (SelfServiceView.extend.test.js:47) is that shape.

- Inherited tooltip always says "role" (SelfServiceResultRow.js:240), though renderLeaveAction handles groups too.

- URL segments aren't encoded. membersHref (:156) and groupHref (:230) interpolate domainName/name from an API response with no encodeURIComponent. Rooted at /domain/… so javascript: isn't reachable, but a name containing /, ?, # or % re-targets the link.

havetisyan (approved):

PR #3495, opened 2026-09-11
iterate multi-account issuer resources during launch auth + tests

harounemahdi:
# Description
Due to the new changes on accepting multiple AWS and GCP accounts under a domain introduced in #3457 we also must make changes to the AWS and GCP service/engine validators to have the ability to iterate through multiple accounts when deciding whether a service has launch authorization on k8s.

i,e. domains registered against multiple accounts (or projects re; GCP), the resource became `domain:service:<acct1>,<acct2>` and no policy assertion could match _unless_ we specifically created an assertion that matches the accounts exactly. This causes confusion when adding new accounts to a domain and we lose granularity on which accounts should be asserted against..

Added unit tests that cover the util fn and common multi-account cases for both AWS and GCP.

NOTE: I have been informed GCP does not yet support multiple projects per domain but these changes are backwards compatible.

# Contribution Checklist:

## Attach Screenshots (Optional) 

@abvaidya


harounemahdi:
yep, you were right on this. We modified our custom attr validator to accommodate multiple accounts. Closing this as it's unneeded.

